PR #24, opened 2020-03-20
Use older time lib

mpscholten:


dpwiz:
Why?

mpscholten:
This branch is used internally for a production app where we couldn't yet update the time library. We have this draft PR to avoid removing this branch by accident ;) We plan to update that app in the future.

mpscholten:
This has been refactored now and now we don't need this branch anymore 🎉

PR #63, opened 2020-05-28
Moved Contributing section into it's own file

mpscholten:
It's useful because otherwise users might try to run the commands in the
README to start their first project.


CONTRIBUTING.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Running an application against a local copy of the framework
+
+To work on the framework in your application, you need to clone this repository inside your application directory:
+
+```
+git clone git@github.com:digitallyinduced/haskellframework.git IHP
+```
+
+The best workflow is to use `ghci` to load your application together with the framework version in `IHP`:
```
Laobiz, line 13:
```suggestion
The best workflow is to use `ghci` to load your application together with the framework version in `IHP`. To make sure that the symlink is loaded run `make all` before that.:
```

PR #71, opened 2020-05-30
Added form & session documentation

mpscholten:


Guide/form.markdown
```diff
@@ -5,8 +5,346 @@
 
 ## Introduction
 
+In IHP Forms are an essential way to interact with your application. Dealing with a lot of form markup can quickly become complex because of the need to deal with consistent styling and especially when dealing with lots of validation. IHP provides helpers to generate form markup to help you deal with the complexity.
+
+By default forms in IHP follows the class names used by Bootstrap 4. Therefore the forms work with Bootstrap 4 out of the box. Of course the default form generation can be customized to support other CSS frameworks.
+
+Unless javascript helpers have been deactivated, your form will be submitted using AJAX and TurboLinks instead of a real browser based form submission.
```
hendi, line 12: Please clarify the following:
- How can they be deactivated (link)?
- What happens when JS helpers are used, but JS is disabled? Does it fall back to HTML form submission gracefully, or does it break?

Guide/form.markdown
```diff
@@ -5,8 +5,346 @@
 
 ## Introduction
 
+In IHP Forms are an essential way to interact with your application. Dealing with a lot of form markup can quickly become complex because of the need to deal with consistent styling and especially when dealing with lots of validation. IHP provides helpers to generate form markup to help you deal with the complexity.
+
+By default forms in IHP follows the class names used by Bootstrap 4. Therefore the forms work with Bootstrap 4 out of the box. Of course the default form generation can be customized to support other CSS frameworks.
```
hendi, line 10: How/where can it be customized for different CSS frameworks?

Guide/form.markdown
```diff
@@ -5,8 +5,346 @@
...
+<div class="form-group" id="form-group-post_title">
+    <label for="post_title">Title</label>
+    <input type="text" name="title" placeholder="" id="post_title" class="form-control is-invalid ">
+    <div class="invalid-feedback">This field cannot be empty</div>
+</div>
+```
+
+## Forms Are Also HSX
+
+It's important to understand that the while the form helpers like `{textField #title}` are called by `formFor`, you can still use HSX there. So you can just add any kind of HSX code inside your form:
```
hendi, line 103: "It's important to understand that **_the_** while the form helpers ..."

Guide/form.markdown
```diff
@@ -5,8 +5,346 @@
...
+```
+
+
+### Advanced Customization Options
+
+The following options are not commonly used, but are useful sometimes.
+
+#### Adding custom attributes to the input element
+
+Form rendering is built on top of blaze html. So you need to import the blaze html functions for this. Add this at the top of your module:
```
hendi, line 244: Add link to https://hackage.haskell.org/package/blaze-html

mpscholten: Done

Guide/form.markdown
mpscholten: Fixed, good catch :)

Guide/form.markdown
mpscholten: Added a section for this. This feature basically is still in todo. While technically it's already possible, we still have to allow the user to change the rendering in a global way.

Guide/form.markdown
mpscholten: Good points :) Also added a section for this.

PR #75, opened 2020-06-02
More validators

hendi:
This adds more validators, including tests.

### DIFF
--- a/IHP/ValidationSupport/ValidateField.hs
+++ b/IHP/ValidationSupport/ValidateField.hs
@@ -16,6 +16,8 @@ import GHC.TypeLits (KnownSymbol, Symbol)
 import GHC.Records
 import IHP.ModelSupport
 import IHP.HaskellSupport
+import Text.Regex.TDFA
+import Data.List ((!!))
 
 -- | A function taking some value and returning a 'ValidatorResult'
 --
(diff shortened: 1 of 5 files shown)


IHP/ValidationSupport/ValidateField.hs
```diff
@@ -128,7 +166,27 @@ nonEmpty value | null value = Failure "This field cannot be empty"
...
+-- Success
+--
+-- >>> empty (Just "hello")
+-- Failure "This field must be empty"
+--
+-- >>> empty Nothing
+-- Success
+empty :: MonoFoldable value => value -> ValidatorResult
+empty value | null value = Success
+empty _ = Failure "This field must be empty"
```
mpscholten, line 185: I would suggest naming this `isEmpty` to avoid name conflicts inside the controller actions (e.g. when you have a local variable `empty` in scope. It's also usually used for data structures, [e.g. Data.Map exports a `empty` function](https://hackage.haskell.org/package/containers-0.4.0.0/docs/Data-Map.html)).

Test/ValidationSupport/ValidateFieldSpec.hs
```diff
@@ -0,0 +1,197 @@
...
+
+When in the IHP directory, you can run this file like:
+
+ > nix-shell NixSupport/shell.nix
+ > ghci
+ > :l Test/ValidationSupport/ValidateFieldSpec.hs
+ > main
+
+-}
+module Test.ValidationSupport.ValidateFieldSpec where
```
mpscholten, line 13: Very nice :)

IHP/ValidationSupport/ValidateField.hs
hendi: I agree, and I'd prefer `isNonEmpty` as well. I had that change originally but ran into the issue that `IHP.Prelude` exports `isEmpty` already. Any suggestions for how to handle this?

mpscholten: Maybe we can call it `isEmptyValue`? :)

Regarding `nonEmpty`: After think about it, I would keep the `nonEmpty` because `isNonEmptyValue` would be very long, and the `nonEmpty` validator is used really often.

hendi: :+1: for pragmatism! I've changed the code accordingly.

PR #82, opened 2020-06-03
add `attributes` view helper

hendi:
This works similar to the `classes` helper already present and allows hsx code like this:

```haskell
button :: Text -> Text -> Bool -> Bool -> Html
button label target isPrimary isEnabled = [hsx|
        <a href="{target}"
           class={classes ["btn", ("btn-primary", isPrimary), ("disabled", not isEnabled)]}
           {attributes [("role", "button")] [("disabled", not isEnabled)]}
        >
            {label}
        </a>
    |]
```

### DIFF
--- a/IHP/ViewSupport.hs
+++ b/IHP/ViewSupport.hs
@@ -47,6 +47,7 @@ import Text.Blaze.Html5.Attributes as A
 import qualified IHP.ControllerSupport as ControllerSupport
 import qualified IHP.Controller.Session as Session
 import IHP.HtmlSupport.QQ (hsx)
+import Data.Text (concat)
 
 type HtmlWithContext context = (?viewContext :: context) => Html5.Html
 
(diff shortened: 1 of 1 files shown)


IHP/ViewSupport.hs
```diff
@@ -111,6 +112,47 @@ classes !classNameBoolPairs =
...
+    attrNameValue =
+        attrNameValuePairs
+        |> map (\nv -> Data.Text.concat [fst nv, "=", show $ snd nv])
+
+    attrName =
+        attrNameBoolPairs
+        |> filter snd
+        |> map fst
+    in
+        unwords $ attrNameValue <> attrName
```
mpscholten, line 152: I think this approach will introduce XSS problems because the attribute value might be supplied by user input. We would need to use [the blaze html `attribute` function](https://hackage.haskell.org/package/blaze-html-0.9.1.2/docs/src/Text.Blaze.Html5.Attributes.html#alt) to make it safe.

But then this function will need to return `[Attribute]` which then cannot be used like `<a {attributes [] []}>`. 

**Here are two possible solutions:**

For use cases like `disabled` or `hidden` we should IMO support boolean attribute values.  This is what you [would expect to work when coming from react's JSX](https://stackoverflow.com/a/29103418).

Then we can just write:

```html
<a role="button" hidden={True}>
```

which is a lot more clean than:

```
<a {attributes [("role", "button")] [("hidden", True)]}>
```

For this change, we would need to update the HSX compiler here: https://github.com/digitallyinduced/ihp/blob/master/IHP/HtmlSupport/QQ.hs#L118

---

Another solution for more dynamic use cases:

[In HSX we already have the following syntax](https://github.com/digitallyinduced/ihp/blob/master/IHP/HtmlSupport/Parser.hs#L76): 
```haskell
let x :: [Attributes] = ... in [hsx|<a {...x}></a>|]
```
(the syntax is implemented / reservered for future use, but the compiler part is missing, so this might fail to compile).

Maybe we can reuse this syntax to allow for:
```haskell
<a
    role="button"
    {...[ ("data-" <> getSomeString, "some-value") ]}
>
```

This is also inspired by JSX syntax.

hendi: Good ideas, `<a href="#" hidden={False} {...[("role", role)]}>` looks much cleaner indeed!

PR #84, opened 2020-06-03
Laobiz/ch1541/hspec tests automatisch ausführen

Laobiz:


.github/workflows/tests.yml
```diff
@@ -0,0 +1,18 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - uses: cachix/install-nix-action@v9
+    - uses: cachix/cachix-action@v6
+      with:
+        name: digitallyinduced
+        signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
+        authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+    - run: |
+          nix-shell ./NixSupport/shell.nix --run "runghc $(make -f lib/IHP/Makefile.dist print-ghc-extensions) ./Test/IDE/SchemaDesigner/ParserSpec.hs"
```
hendi, line 18: This is missing `Test/ValidationSupport/ValidateFieldSpec.hs`. Maybe use something like `find Test -name '*Spec.hs'` for auto-discovery?

Laobiz: Good point. I was thinking we add a `Test.sh` later on where we can add all tests but even though this is a great solution now :) Added it in https://github.com/digitallyinduced/ihp/pull/84/commits/256a8da7c125d2d499c3d32a7ee6d9d94a2b3156 :)

hendi:
This does not work for me (might just be my local setup though!)

```
$ nix-shell ./NixSupport/shell.nix --run "runghc $(make -f lib/IHP/Makefile.dist print-ghc-extensions) "Test/IDE/SchemaDesigner/ParserSpec.hs""

IHP/IDE/SchemaDesigner/Compiler.hs:13:1: error:
    The type signature for `writeSchema' lacks an accompanying binding
   |
13 | writeSchema :: [Statement] -> IO ()
   | ^^^^^^^^^^^
``` 

and

```
$ nix-shell ./NixSupport/shell.nix --run "runghc $(make -f lib/IHP/Makefile.dist print-ghc-extensions) "Test/ValidationSupport/ValidateFieldSpec.hs""

IHP/ValidationSupport/Types.hs:23:21: warning: [-Wmissing-space-after-bang]
    Did you forget to enable BangPatterns?
    If you mean to bind (!) then perhaps you want
    to add a space after the bang for clarity.
   |
23 | attachFailure field !message = modify #meta prependAnnotation
   |                     ^^^^^^^^

IHP/ValidationSupport/Types.hs:23:1: error:
    Parse error in pattern: attachFailure
   |
23 | attachFailure field !message = modify #meta prependAnnotation
   | ^^^^^^^^^^^^^^^^^^^
```

Laobiz:
You need the most recent `Makefile.dist` from the master.  The problem is that not all ghc-options were inside of it so I added them directly to the master in https://github.com/digitallyinduced/ihp/commit/d350a04a5751a755cf43395698a5c4d03010b601.

hendi:
Sorry, you're right, I was on the wrong branch.

Works, :+1:

PR #90, opened 2020-06-04
Added controller documentation

mpscholten:


Guide/controller.markdown
```diff
@@ -0,0 +1,355 @@
...
+
+Let's say we want to create a new controller with a single action `ShowPostAction`. First we need to define our types in `Web/Types.hs`:
+
+```haskell
+data PostsController
+    = ShowPostAction { postId :: !(Id Post) }
+    deriving (Eq, Show, Data)
+```
+
+This define a type `PostsController` with a data constructor `ShowPostAction { postId :: !(Id Post) }`. The arguent `postId` will later be filled with the `postId` parameter of the request url. This is done automatically by the IHP router. IHP also requires the controller to have `Eq`, `Show` and `Data` instances. Therefore we derive them here.
```
Laobiz, line 28:
```suggestion
This defines a type `PostsController` with a data constructor `ShowPostAction { postId :: !(Id Post) }`. The arguent `postId` will later be filled with the `postId` parameter of the request url. This is done automatically by the IHP router. IHP also requires the controller to have `Eq`, `Show` and `Data` instances. Therefore we derive them here.
```

Guide/controller.markdown
```diff
@@ -0,0 +1,355 @@
...
+
+instance Controller PostsController where
+    action ShowPostAction { postId } = renderPlain "Hello World"
+```
+
+This implementation for `ShowPostAction` responds with a simple plain text message. The `action` implementation is usally a big pattern match over all possible actions of a controller.
+
+## Reading Query and Body Parameters
+
+Inside the action you can access request parameters using the `param` function. A parameter can either be a url parameter like `?paramName=paramValue` (*this is also called a query parameter*), or given as a form field like `<form><input name="paramName" value="paramValue"/></form>` (*in that case we're talking about a body parameter*). The `param` function will work with query and body parameters, so you don't have to worry about that.
```
Laobiz, line 45: What does this do when there are a body and a query parameter with the same name? Maybe not common but good to know.

mpscholten: Added
```
in case a query and body parameter is set with the same name, the body parameter will take priority
```
to the markdown :)

PR #91, opened 2020-06-05
Added documentation on naming things

mpscholten:


hendi:
:+1: (both documentation as well as intention)

PR #94, opened 2020-06-07
Added documentation on routing

mpscholten:


Guide/routing.markdown
```diff
@@ -0,0 +1,190 @@
...
+To generate a full url, use `urlTo`:
+
+```haskell
+urlTo NewUserAction
+-- http://localhost:8000/NewUser
+```
+
+## AutoRoute
+
+Let's say our `PostsController` is define in `Web/Types.hs` like this:
```
hendi, line 47: "defined" is missing a d

Guide/routing.markdown
```diff
@@ -0,0 +1,190 @@
...
+
+This will generate a routing like:
+
+```haskell
+GET /EditPost?postId={postId}&userId={userId} => EditPostAction { postId, userId }
+```
+
+### Parameter Types
+
+AutoRoute by default only works with UUID arguments (and Id types like `Id Post`). If you have a controller like `data HelloWorldController = HelloAction { name :: Text }` where you have text parameter, you have to configure AutoRoute like this:
```
hendi, line 95: "where you have **a** text parameter

Guide/routing.markdown
```diff
@@ -0,0 +1,190 @@
...
+
+```haskell
+instance AutoRoute HelloWorldController where
+    parseArgument = parseTextArgument
```
hendi, line 99: - Is there a `parseIntArgument`?
- How would one use own arguments there, in a type safe way? E.g. a type for a Youtube-style shortid.

Guide/routing.markdown
```diff
@@ -0,0 +1,190 @@
...
+If you need more strong rules, consider using the other routing APIs available or overriding the `allowedMethodsForAction` like this:
+
+```haskell
+instance AutoRoute HelloWorldController where
+    allowedMethodsForAction "HelloAction" = [ GET ]
+```
+
+### Application Prefix
+
+When using multiple application in your IHP project, e.g. having an admin backend, AutoRoute will prefix with action urls with the application name. E.g. a controller `HelloWorldController` defined in `Admin/Types.hs` will be automatically be prefixed with `/admin` and generate urls such as `/admin/HelloAction`.
```
hendi, line 132: will prefix <strike>with</strike> action urls

hendi: "be automatically be"

Guide/routing.markdown
```diff
@@ -0,0 +1,190 @@
...
+```
+
+We want requests to `/posts` to map to `ShowAllMyPostsAction`. For that we need to add a `CanRoute` instance:
+
+```haskell
+instance CanRoute PostsController where
+    parseRoute' = string "/posts" <* endOfInput >> pure ShowAllMyPostsAction
+```
+
+The `parseRoute'` function is a parser which reads an url and returns an action of type `PostsController`. The router uses [attoparsec](https://hackage.haskell.org/package/attoparsec). Take a look at the attoparsec documentation to get a good understanding on how to parse complex urls.
```
hendi, line 153: This should include a table of the most commonly used patterns in URLs, instead of referring users to a parser combinator library!

Guide/routing.markdown
```diff
@@ -0,0 +1,190 @@
...
+POST /UpdatePost?postId={postId}    => UpdatePostAction { postId }
+PATCH /UpdatePost?postId={postId}   => UpdatePostAction { postId }
+DELETE /DeletePost?postId={postId}  => DeletePostAction { postId }
+```
+
+The urls are very close to the actual action which is called. Action parameters are taken automatically from the request query. This design helps you to always know which action is actually called, when requesting an url. 
+
+### AutoRoute & Beautiful Urls
+
+Lots of modern browser don't even show the full url bar anymore (e.g. Safari and most mobile browsers). Therefore AutoRoute doesn't aim to generate the "most" beautiful urls out of the box. It's rather optimized for the needs of developers. If you need beautiful urls for SEO reasons, instead of using AutoRoute you can use the more manual APIs of IHP Routing.
```
hendi, line 77: Needs a link to the relevant section in the docs, currently the section `Custom Routing` on this file

Guide/routing.markdown
```diff
@@ -0,0 +1,190 @@
...
+Next to the routing itself, we also need implement the url generation:
+
+```haskell
+instance HasPath PostsController where
+    pathTo ShowAllMyPostsAction = "/posts"
+```
+
+### Real-World Example
+
+Here is an real world example of a custom routing implementation for a custom Apple Web Service interface implemented at digitally induced:
```
hendi, line 164: "**a** real world example"

Guide/routing.markdown
```diff
@@ -159,9 +159,68 @@ instance HasPath PostsController where
     pathTo ShowAllMyPostsAction = "/posts"
 ```
 
+### Beautiful URLs
+
+Let's say we want to give our blog post application a beautiful url structures for SEO reasons. Our controller is defined as:
```
hendi, line 164: structure<strike>s</strike>

Guide/routing.markdown
```diff
@@ -159,9 +159,68 @@ instance HasPath PostsController where
...
+    = ShowPostAction { postId :: !(Maybe (Id Post)), slug :: !(Maybe Text) }
+```
+
+This will also require us to make changes to our action implementation:
+
+```haskell
+action ShowPostAction { postId, slug } = do 
+    post <- case slug of
+            Just slug -> query @Post |> filterWhere (#slug, slug) |> fetchOne
+            Nothing   -> fetchOne postId
```
hendi, line 196: Does `fetchOne` accept a `Maybe (Id Post)`?

mpscholten: Yes :) https://github.com/digitallyinduced/ihp/blob/master/IHP/QueryBuilder.hs#L430 It's a super useful helper instance because this pattern comes up a lot in practise.

Guide/routing.markdown
mpscholten: `parseIntArgument` has been added now :)

Also added a section on creating a custom `parseArgument`

Guide/routing.markdown
mpscholten: I've reworded this section a  bit. We're mostly using AutoRoute at digitallyinduced, therefore we don't really have that much example codes or patterns we can document. I hope we can add some more documentation as well as better apis for custom routes in the future.

Guide/routing.markdown
mpscholten: Fixed, good catch :)

Guide/routing.markdown
mpscholten: Good catch, fixed :)

PR #112, opened 2020-06-13
update new column button to say "New Column" instead of just "New"

jeyj0:


mpscholten:
Also thanks for this one :) I guess the CI is broken because some other stuff. I'll fix that today.

PR #118, opened 2020-06-17
added showgencode for enums; removed empty enum error

nezbite:


IHP/IDE/SchemaDesigner/Controller/Schema.hs
```diff
@@ -50,6 +50,12 @@ instance Controller SchemaController where
         let generatedHaskellCode = SchemaCompiler.compileStatementPreview statements table
         render GeneratedCodeView { .. }
 
+    action ShowGeneratedEnumCodeAction { enumName } = do
+        statements <- readSchema
+        let (Just enum) = findEnumByName enumName statements
```
mpscholten, line 55: Right now this looks pretty much the same as the. `ShowGeneratedCodeAction`. Can't we just introduce a `findStatementByName` (which basically find enum or table) and then use `findStatementByName` instead of `findTableByName` in the `ShowGeneratedCodeAction`?

nezbite: In case of an EnumType and a Table having the same name, this function would have a hard time finding the right one; When referring to a Table and the function returns an Enum or vice versa there would be an issue..

mpscholten: You cannot have a enum named like a table at the same time in Postgres

PR #124, opened 2020-06-23
Run nix-shell with --pure when creating the .envrc

madjar:
This fixes a bug where the `.envrc` keeps growing every time you regenerate it.

By default, nix-shell gets the current $PATH and prepends the nix store $PATH. That means that if you already have the current `.envrc` loaded, both the current content of `.envrc` and the new computed path will be save to `.envrc`, making it grow over time. This is fixed by calling nix-shell with `--pure`.


mpscholten:
Good catch! Thanks :)

Btw, we're already looking into the failing CI.

madjar:
Super, thanks!

PR #129, opened 2020-06-23
Corrected a few typos in architecture.markdown

Dansvidania:
corrected a few typos from `to much` to `too much` and from `few javascript` to `little javascript`


mpscholten:
Good catch :) Thanks you!

PR #134, opened 2020-06-24
improved and streamlined the installation process for windows and linux

nezbite:


Guide/installation.markdown
```diff
@@ -28,7 +28,7 @@ Before installing nix and IHP, you need `curl` and `git` if you don't have them
 ```bash
 sudo apt update
 sudo apt upgrade
-sudo apt install curl git -y
+sudo apt install git curl -y
```
mpscholten, line 31: Should we add make as well here?

nezbite: wouldn't hurt, I'll add it real quick

PR #143, opened 2020-06-25
added js for os specific spoilers in documentation

nezbite:


mpscholten:
image

Looks like the `Windows Systems Paths` is off

mpscholten:
This fails when opening http://localhost:3000/installation.html and then clicking `Your First Project` and then going back via clicking `Installing IHP`.

PR #150, opened 2020-06-25
Fix a few small typos in installation.markdown

moodmosaic:


mpscholten:
Good catch! Thanks :)

PR #163, opened 2020-06-26
Update your-first-project.markdown

waddlaw:
The message is a bit different.


mpscholten:
Good catch! :) Thanks

PR #168, opened 2020-06-26
use session store, even on Linux

hendi:
Apparently there was a segfault on Linux earlier, so a memory cache was used for the session store on Linux. This causes frequent logouts though, so I removed it. No segfaults so far. If they happen again, let's fix the root cause :-)


mpscholten:
Just for documenting this: The related issue is https://github.com/yesodweb/clientsession/issues/35

PR #191, opened 2020-06-29
Reserved haskell keywords

nezbite:


IHP/IDE/SchemaDesigner/Controller/Tables.hs
```diff
@@ -37,7 +37,7 @@ instance Controller TablesController where
         render NewTableView { .. }
 
     action CreateTableAction = do
-        let tableName = param "tableName"
+        let tableName = compileIdentifier (param "tableName")
```
mpscholten, line 40: IMO this should not be happen in the Schema Designer itself, because you can also edit the sql file manually and it should then still work. We should instead put it into the `IHP.SchemaCompiler`

IHP/IDE/SchemaDesigner/Compiler.hs
```diff
@@ -76,12 +76,17 @@ compilePostgresType PTime = "TIME"
 compilePostgresType (PCustomType theType) = theType
 
 compileIdentifier :: _ -> Text
-compileIdentifier identifier = if identifierNeedsQuoting then tshow identifier else identifier
+compileIdentifier identifier = if isHaskellKeyword
```
mpscholten, line 79: The sql field name should still be the "normal" thing. Only the haskell field should be named differently.

nezbite: Haskell Keywords are fixed in the parser now. Manual SQL changes will now work

IHP/IDE/SchemaDesigner/View/Layout.hs
```diff
@@ -76,10 +76,10 @@ databaseControls = [hsx|
...
 findStatementByName statementName statements = find pred statements
     where
-        pred CreateTable { name } | (toUpper name) == (toUpper statementName) = True
-        pred CreateTable { name } | (toUpper name) == (toUpper (tshow statementName)) = True
-        pred CreateEnumType { name } | (toUpper name) == (toUpper statementName) = True
-        pred CreateEnumType { name } | (toUpper name) == (toUpper (tshow statementName)) = True
+        pred CreateTable { name } | (compileIdentifier name) == (compileIdentifier statementName) = True
+        pred CreateTable { name } | (parseKeyword name) == (parseKeyword statementName) = True
+        pred CreateEnumType { name } | (compileIdentifier name) == (compileIdentifier statementName) = True
+        pred CreateEnumType { name } | (parseKeyword name) == (parseKeyword statementName) = True
```
mpscholten, line 82: Why these changes?

IHP/NameSupport.hs
```diff
@@ -80,8 +80,71 @@ lcfirst = applyFirst toLower
...
+
+-- | Add '_' to the end of a name if it is a reserved haskell keyword
+--
+-- >>> parseKeyword "test"
+-- "test"
+--
+-- >>> parseKeyword "type"
+-- "type_"
+parseKeyword :: Text -> Text
+parseKeyword name = if toLower name `elem` haskellKeywords then name <> "_" else name
```
mpscholten, line 97:
```suggestion
escapeHaskellKeyword :: Text -> Text
escapeHaskellKeyword name = if toLower name `elem` haskellKeywords then name <> "_" else name
```

IHP/NameSupport.hs
```diff
@@ -80,8 +80,71 @@ lcfirst = applyFirst toLower
...
 ucfirst :: Text -> Text
 ucfirst = applyFirst toUpper
 {-# INLINE ucfirst #-}
+
+-- | Add '_' to the end of a name if it is a reserved haskell keyword
+--
+-- >>> parseKeyword "test"
+-- "test"
+--
+-- >>> parseKeyword "type"
```
mpscholten, line 94:
```suggestion
-- >>> escapeHaskellKeyword "type"
```

IHP/NameSupport.hs
```diff
@@ -80,8 +80,71 @@ lcfirst = applyFirst toLower
...
--- >>> ucfirst "Alread uppercase"
+-- >>> ucfirst "Already uppercase"
 -- "Already uppercase"
 ucfirst :: Text -> Text
 ucfirst = applyFirst toUpper
 {-# INLINE ucfirst #-}
+
+-- | Add '_' to the end of a name if it is a reserved haskell keyword
+--
+-- >>> parseKeyword "test"
```
mpscholten, line 91:
```suggestion
-- >>> escapeHaskellKeyword "test"
```

IHP/IDE/SchemaDesigner/View/Layout.hs
nezbite: This was necessary when the escaping was done in the parser. I'll revert this.

PR #193, opened 2020-06-30
Action Generator

Laobiz:


mpscholten:
When typing `ShowAction` into the generator, I get `ShowActionAction` as the generated data constructor. Can we fix this? When I type just `Show` it should still generate `ShowAction` after changing this :)

mpscholten:
Also can we add special case handling, like when generating a `ShowAction`, it should look like the show action generated by the controller generator?

IHP/IDE/CodeGen/ActionGenerator.hs
```diff
@@ -43,7 +43,7 @@ generateGenericAction schema config =
             singularName = config |> get #modelName
             singularVariableName = lcfirst singularName
             pluralVariableName = lcfirst controllerName
-            nameWithSuffix = name <> "Action" --e.g. "TestAction"
+            nameWithSuffix = if ("Action" `isSuffixOf` name) then name else name <> "Action" -- e.g. TestAction
```
mpscholten, line 46: IMO reads better written like this:

```suggestion
            nameWithSuffix = if "Action" `isSuffixOf` name
                    then name
                    else name <> "Action" -- e.g. TestAction
```

IHP/IDE/CodeGen/ActionGenerator.hs
```diff
@@ -0,0 +1,131 @@
...
+            deleteContent =
+                ""
+                <> "    action Delete" <> singularName <> "Action { " <> idFieldName <> " } = do\n"
+                <> "        " <> modelVariableSingular <> " <- fetch " <> idFieldName <> "\n"
+                <> "        deleteRecord " <> modelVariableSingular <> "\n"
+                <> "        setSuccessMessage \"" <> model <> " deleted\"\n"
+                <> "        redirectTo " <> name <> "Action\n"
+
+            typesContent = 
+                   "    | " <> nameWithSuffix
```
mpscholten, line 126: When I generate a `ShowUser` action, this will generate the data constructor:

```
| ShowUserAction
```

But it should be the following:

```
| ShowUserAction { userId :: !(Id User) }
```

IHP/IDE/CodeGen/ActionGenerator.hs
```diff
@@ -0,0 +1,141 @@
...
+
+            typesContentGeneric = 
+                   "    | " <> nameWithSuffix
+
+            typesContentWithParameter = 
+                   "    | " <> nameWithSuffix <> " { " <> idFieldName <> " :: !(" <> idType <> ") }\n"
+            
+
+            chosenContent = fromMaybe actionContent (lookup nameWithSuffix specialCases)
+            chosenType = if chosenContent `elem` [actionContent, newContent, createContent]
```
Laobiz, line 134:
```suggestion
            chosenType = if chosenContent `elem` [actionContent, newContent, createContent, indexContent]
```

mpscholten:
image

When you use the generator like this (entering `test`), the name is not normalized. We should use `ucfirst` on the input to fix that.

PR #196, opened 2020-06-30
Fix typo: automtaically -> automatically

stianlagstad:


mpscholten:
Thank you, good catch! :)

PR #203, opened 2020-07-02
Check if a name already exists in CodeGen

### LINKED ISSUE #173: Codegenerator check if name exists
Running the codegenerator for a second time does not check for name uniqueness. 

What happens is it:
- for files where it appends, it appends more code
- for new files, it overwrites entire file

Preferably, it should warn that this name already exists. If the warning is ignored, I would expect to overwrite in both cases.

Laobiz:
This closes #173


IHP/IDE/CodeGen/Controller.hs
```diff
@@ -27,8 +27,16 @@ instance Controller CodeGenController where
 
     action NewControllerAction = do
         let controllerName = paramOrDefault "" "name"
-        plan <- ControllerGenerator.buildPlan controllerName
-        render NewControllerView { .. }
+        controllerAlreadyExists <- case Text.splitOn "." controllerName of
+            [applicationName, controllerName'] -> doesFileExist $ cs applicationName <> "/Controller/" <> cs controllerName' <> ".hs"
+            [controllerName'] -> doesFileExist $ "Web/Controller/" <> cs controllerName' <> ".hs"
```
mpscholten, line 32: Can we move this into a helper function `doesControllerExists name` (at the bottom of the controller file)? Then we could do just `controllerAlreadyExists <- doesControllerExists name`.

Also right now we have duplicated the controller name handling stuff. Can we have a function `normalizeControllerName :: Text -> (Text, Text)`, which returns `("Admin", "Users")` for e.g. `"admin.users"`. Then this function can be used by this controller and also by the `ControllerGenerator`

IHP/IDE/CodeGen/Controller.hs
```diff
@@ -27,8 +27,16 @@ instance Controller CodeGenController where
...
+        if controllerAlreadyExists
```
mpscholten, line 33: We could write this less nested by using `when`:

```haskell
when controllerAlreadyExists do
    setErrorMessage "..."
    redirect ...
```

IHP/IDE/CodeGen/Controller.hs
Laobiz: I think a complete separate function would be of no use because it is and only will be used inside of this action. For better readability we could put it in the where clause.

mpscholten: Yeah this is omly for readability :)

IHP/IDE/CodeGen/Controller.hs
Laobiz: Done :)

IHP/IDE/CodeGen/Controller.hs
Laobiz: Done :)

Issue #182, opened 2020-06-29
Fix ClientSession Issue


mpscholten:
https://github.com/yesodweb/clientsession/issues/35

PR #204, opened 2020-07-02
Disabled external interpreter to avoid issues on certain Linux systems

mpscholten:
Additonally this improves the performance on VMs a lot when loading IHP as a source module.
Especially fixes the issue with clientsession (#182).


hendi:
Great, this seem to fix some random instability issues I've been having, and also make sessions work on Linux finally :tada:

PR #205, opened 2020-07-02
Open the local dev url in an async way

### LINKED ISSUE #202: Fails to start on KDE / Fedora 32
On a clean KDE install on Fedora 32 neither the DevServer nor the application server start up.

The issue seems to be that `xdg-open` uses KDE's `kioclient`, which checks if the parsed URL is reachable before passing it along to the browser. Since IHP calls `xdg-open` before :8000 is ready, this stalls.

kioclient's message is `localhost contacted. Waiting for reply...`

This is obviously stupid behaviour from KDE's and/or Fedora's side.

Workaround for IHP: support an env flag to disable auto-start of the browser.

mpscholten:
Fixes #202


hendi:
I've tested this on Fedora 32 with KDE, didn't help.

mpscholten:
@hendi could you quickly take a look again? :)

PR #212, opened 2020-07-04
[#211] Normalise field names and type arguments in SchemaCompiler

### LINKED ISSUE #211: Snake-caps Tables with Foreign Key Constraints Break Generated Types
If we follow the tutorial example with a `posts` table, but replace `comments` with `user_comments`, we see two issues in `build/Generated/Types.hs`.

First, `type instance Include "userComments" (Posts' user_comments) = Posts' user_comments` instead of `Posts' [UserComment]`. This breaks `fetchRelated`.

Second, the field in `Posts'` type is called `user_comments`, rather than `userComments` as is the naming convention. This then propagates to instances in the file such as `UpdateField "user_comments"...` and also breaks `fetchRelated`.

This will happen with any second table with a name that has multiple words in snake caps and a foreign key constraint.

ruhatch:
Variations in naming of field names and type arguments were breaking the logic
for generating instances in the SchemaCompiler. This led to broken fetchRelated
functionality when referencing tables had multiple words in their names. This is
a relatively simple and neat way to clean up those issues.

Fixes #211

### DIFF
--- a/IHP/SchemaCompiler.hs
+++ b/IHP/SchemaCompiler.hs
@@ -189,7 +189,7 @@ compileData table@(CreateTable { name, columns }) =
 
 -- | Returns all the type arguments of the data structure for an entity
 dataTypeArguments :: (?schema :: Schema) => Statement -> [Text]
-dataTypeArguments table = belongsToVariables <> hasManyVariables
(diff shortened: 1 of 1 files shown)


mpscholten:
Just tested this and seems to work well :-) Thanks a lot for fixing this!

PR #216, opened 2020-07-06
Added remainingText function which replaces the old getInput

### LINKED ISSUE #213: Implement getInput
It seems like the `getInput` function referenced in the [guide](https://ihp.digitallyinduced.com/Guide/routing.html#autoroute-beautiful-urls) would be a helpful utility function to include in IHP.RouterSupport.

mpscholten:
Fixes #213 

The "old" `getInput` has been renamed to `takeByteString` in the latest attoparsec version. We also usually want the result as a `Text` value, therefore a new `remainingText` has been added :)


mpscholten:
@nghamilton could you take a look?

nghamilton:
lgtm, thanks!

PR #218, opened 2020-07-06
Fix bootstrap table-responsive in generated Index view

ruhatch:
Update the generated Index view to match the [bootstrap documentation of `.table-responsive`](https://getbootstrap.com/docs/4.5/content/tables/#always-responsive).


mpscholten:
Thanks :)

PR #220, opened 2020-07-08
added tstamp, varying(n), numeric(x,y)

nezbite:


IHP/IDE/SchemaDesigner/Types.hs
```diff
@@ -64,10 +64,14 @@ data PostgresType
...
     | PBoolean
     | PTimestampWithTimezone
+    | PTimestamp
     | PReal
     | PDouble
     | PDate
     | PBinary
     | PTime
+    | PNumeric
+    | PNumericP Int Int
```
mpscholten, line 74: Can't we unify these types by using maybe fields?

```suggestion
    | PNumericP { precision :: Maybe Int, scale :: Maybe Int }
```

IHP/IDE/SchemaDesigner/Parser.hs
```diff
@@ -194,6 +212,25 @@ sqlType = choice
                     try (symbol' "TIME")
                     pure PTime
 
+                numericPS = do
+                    try (symbol' "NUMERIC(")
+                    values <- between (space) (char ')' >> space) (varExpr `sepBy` (char ',' >> space))
+                    case values of
+                        [VarExpression precision, VarExpression scale] -> pure (PNumericP (tryTextToInt precision) (tryTextToInt scale))
+                        [VarExpression precision] -> pure (PNumericP (tryTextToInt precision) 0)
+                        _ -> pure PNumeric
```
mpscholten, line 221: This will silently fail here, we should call `fail` here instead:
```suggestion
                        _ -> fail "Failed to parse NUMERIC(..) expression"
```

IHP/IDE/SchemaDesigner/Parser.hs
```diff
@@ -229,3 +266,6 @@ comment = do
     lexeme "--" <?> "Line comment"
     content <- takeWhileP Nothing (/= '\n')
     pure Comment { content }
+
+tryTextToInt :: Text -> Int
+tryTextToInt input = fromMaybe 0 (textToInt input)
```
mpscholten, line 271: This should not silently fail

IHP/IDE/SchemaDesigner/Parser.hs
```diff
@@ -194,6 +212,25 @@ sqlType = choice
...
+                        [VarExpression precision, VarExpression scale] -> pure (PNumericP (tryTextToInt precision) (tryTextToInt scale))
+                        [VarExpression precision] -> pure (PNumericP (tryTextToInt precision) 0)
+                        _ -> pure PNumeric
+
+                numeric = do
+                    try (symbol' "NUMERIC")
+                    pure PNumeric
+
+                varyingN = do
+                    try (symbol' "VARYING(") <|> try (symbol' "VARCHAR(") <|> try (symbol' "CHAR(") <|> try (symbol' "CHARACTER(")
```
mpscholten, line 228: Isn't `character` different from `varchar`? This would turn any `CHARACTER(1)` to `VARCHAR(1)` when parsing and then resaving the schema.

mpscholten:
The tests are still failing

PR #224, opened 2020-07-09
Action Generator option for corresponding view

### LINKED ISSUE #200: ActionGenerator should have an option to auto generate the corresponding view
* If the option is ticked the action should redirect to the created view
* The view should be created through the ViewGenerator and the plan for it should be appended to the actionGenerator plan

Laobiz:
closes #200


mpscholten:
image

Right now the view is at the top of my preview, IMO it should be at the bottom

mpscholten:
Can we have the view checkbox on the preview as well (then resubmit the preview form when it's changed)?

PR #225, opened 2020-07-10
Fix date-times to correctly show abbreviated timezone suffix.

nghamilton:
It looks like 'Uhr' abbreviated suffix is hard coded into the date-time JavaScript helper.


mpscholten:
Thanks :) `Uhr` is the german word for `Time`. So actually this is not a timezone suffix. Maybe the fix could be to just remove the word `Uhr`?

mpscholten:
Thanks 💯

PR #226, opened 2020-07-11
Accept empty strings in optional parameter parsers, returning Nothing

nghamilton:
Nothing will only be returned after the read is attempted, which will continue to allow empty stringy types to parse successfully as `Just ""`


mpscholten:
Thank you! Very good change 🚀

PR #231, opened 2020-07-13
Issue 199/generators application aware

### LINKED ISSUE #199: Generators should be application-aware
Currently generators only work with the `Web` application. This should be selectable when there exist multilple apps.

Laobiz:
closes #199

### DIFF
--- a/IHP/IDE/CodeGen/Controller.hs
+++ b/IHP/IDE/CodeGen/Controller.hs
@@ -88,15 +88,16 @@ instance Controller CodeGenController where
 
     action NewActionAction = do
         let actionName = paramOrDefault "" "name"
-        let applicationName = "Web"
+        let applicationName = paramOrDefault "Web" "applicationName"
         let controllerName = paramOrDefault "" "controllerName"
-        controllers <- findWebControllers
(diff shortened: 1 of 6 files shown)


mpscholten:
Can we also add this to the controller generator?

mpscholten:
Where is the application argument passed from the preview to the actual code generation? I could not find a hidden input for this, and it's also not an url query param

Laobiz:
@mpscholten There is no argument that is passed. The path of the generated files is altered and so the plan which is displayed in the preview is another than the one with another application.

I will apply the Select Form to the controller generator as well :)

mpscholten:
But when clicking Generate (after looking at the preview), where is the application passed?

Laobiz:
Oh yeah good point. There is a hidden field missing I think. Good catch 👍

mpscholten:
Can we hide the application selector when there is only a single application? Because then it's basically just noise

Laobiz:
Good point :)

Marc Scholten <notifications@github.com> schrieb am Di., 14. Juli 2020,
11:55:

PR #234, opened 2020-07-15
Fixed controller generator generating code without table fields when …

### LINKED ISSUE #194: Missing fields in generated Edit.hs and New.hs views
Hi, I was trying to experiment with a many-to-many relation, but something is wrong in the view generations of the connecting table. See schema below.

## Steps to reproduce

1. ihp-new test
2. cd test and ./start
3. recreate schema in IHP
4. update and push to db
5. create new codegen/Controller for `ProjectProducts`
6. open `http://localhost:8000/NewProjectProduct`

## Expected

A form with two UUID fields and a quantity field

## Bug

The form turns out empty, also in the controller the build statement seems empty.
```
buildProjectProduct projectProduct = projectProduct
    |> fill @'[]
```
I verified that generally when I add an `Int` column to a model it works fine, it almost seems as if the double foreign key is breaking this.

```
CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
CREATE TABLE products (
    title TEXT NOT NULL,
    description TEXT NOT NULL,
    created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL,
    updated_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL,
    project_id UUID NOT NULL
);
CREATE TABLE projects (
    title TEXT NOT NULL,
    author TEXT NOT NULL,
    created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL
);
CREATE TABLE project_products (
    project_id UUID NOT NULL,
    product_id UUID NOT NULL,
    quantity INT NOT NULL
);
ALTER TABLE products ADD CONSTRAINT products_ref_project_id FOREIGN KEY (project_id) REFERENCES projects (id) ON DELETE NO ACTION;
ALTER TABLE project_products ADD CONSTRAINT project_products_ref_project_id FOREIGN KEY (project_id) REFERENCES projects (id) ON DELETE NO ACTION;
ALTER TABLE project_products ADD CONSTRAINT project_products_ref_product_id FOREIGN KEY (product_id) REFERENCES products (id) ON DELETE NO ACTION;
```

mpscholten:
Fixed controller generator generating code without table fields when  there is an "_" in the table name. This fixes #194.

### DIFF
--- a/IHP/IDE/CodeGen/ControllerGenerator.hs
+++ b/IHP/IDE/CodeGen/ControllerGenerator.hs
@@ -150,7 +150,7 @@ generateController schema config =
             <> "        render EditView { .. }\n"
 
         modelFields :: [Text]
-        modelFields = fieldsForTable schema modelVariablePlural
+        modelFields = fieldsForTable schema (modelNameToTableName modelVariableSingular)
 
         updateAction =
             ""
--- a/IHP/NameSupport.hs
+++ b/IHP/NameSupport.hs
@@ -3,9 +3,19 @@ Module: IHP.NameSupport
 Description:  Transforms names, e.g. table names to model names
 Copyright: (c) digitally induced GmbH, 2020
 -}
-module IHP.NameSupport (tableNameToModelName, columnNameToFieldName, humanize, ucfirst, lcfirst, fieldNameToColumnName, escapeHaskellKeyword) where
+module IHP.NameSupport
+( tableNameToModelName
+, columnNameToFieldName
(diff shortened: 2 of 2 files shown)


Laobiz:
When I generate a controller I get an error in the Routes.hs file, cause the generation looks like this:

image

mpscholten:
I guess the new line is missing in the generated code. But this is not related to this issue IMO, should still be fixed nonetheless

PR #238, opened 2020-07-16
Use "prinf %q" instead of "echo" in Makefile.dist

adithyaov:
$PATH might contain spaces and other special characters. We need to quote them else the script will fail.


mpscholten:
Thanks 🎉

PR #246, opened 2020-07-18
Fixed quickjump index not build by haddock causing quickjump to fail in api docs

### LINKED ISSUE #245: Quick Jump not working on API docs
I get this error when trying to use quick jump on the [API docs](https://ihp.digitallyinduced.com/api-docs/):

image

mpscholten:
Fixes #245


mpscholten:
Once https://github.com/digitallyinduced/ihp/actions/runs/173583676 has completed, the fix will be live :)

PR #248, opened 2020-07-18
Improve Create and UpdateRecord

ruhatch:
We now:
- Keep track of fields that are set in the `touchedFields` attribute of the `meta` field
- Use the `touchedFields` to correctly use the `DEFAULT` SQL value
- Use the `touchedFields` to only update new values in `updateRecord`


mpscholten:
Very good job 💯 Works very well! :)

Issue #249, opened 2020-07-18
Improve <script> tag handling in HSX


mpscholten:
Right now when having HSX code like this:

```html
<script type="text/javascript">
// Load google charts
google.charts.load('current', {'packages':['corechart']});
google.charts.setOnLoadCallback(drawChart);

// Draw the chart and set the chart values
function drawChart() {
  var data = google.visualization.arrayToDataTable([
  ['Task', 'Hours per Day'],
  ['Work', 8],
  ['Friends', 2],
  ['Eat', 2],
  ['TV', 2],
  ['Gym', 2],
  ['Sleep', 8]
]);

  // Optional; add a title and set the width and height of the chart
  var options = {'title':'My Average Day', 'width':550, 'height':400};

  // Display the chart inside the <div> element with id="piechart"
  var chart = new google.visualization.PieChart(document.getElementById('piechart'));
  chart.draw(data, options);
}
</script>
```

The HSX parser will fail with a bad error message. This is because the `{` in `google.charts.load('current', {'packages':['corechart']});` starts a new inline haskell expression (like in `<b>{my haskell expr}</b>`).

We should add special handling for `<script>` tags to not handle `{}` expressions at all.

cc: @chrislf

PR #250, opened 2020-07-19
Hsx script and style

mpscholten:
Fixes https://github.com/digitallyinduced/ihp/issues/249

### DIFF
--- a/IHP/HtmlSupport/Parser.hs
+++ b/IHP/HtmlSupport/Parser.hs
@@ -26,7 +26,8 @@ data Attribute = StaticAttribute !Text !AttributeValue | SpreadAttributes Text d
 
 data Node = Node !Text ![Attribute] ![Node]
     | TextNode !Text
-    | SplicedNode !Text
+    | PreEscapedTextNode !Text -- ^ Used in @script@ or @style@ bodies
+    | SplicedNode !Text -- ^ Inline haskell expressions like @{myVar}@ or @{f "hello"}@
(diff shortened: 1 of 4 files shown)


mpscholten:
@ruhatch maybe you could take a look over this PR and do a quick review? :)

IHP/HtmlSupport/Parser.hs
```diff
@@ -126,13 +146,16 @@ hsxClosingElement name = do
...
-    let isWhitespaceTextOnly = Text.null (Text.strip value)
-    pure if isWhitespaceTextOnly
-        then TextNode ""
-        else TextNode (collapseSpace value)
+hsxText = buildTextNode <$> takeWhile1P (Just "text") (\c -> c /= '{' && c /= '}' && c /= '<' && c /= '>')
+
+-- | Builds a TextNode and strips all surround whitespace from the input string
+buildTextNode :: Text -> Node
+buildTextNode value | Text.null (Text.strip value) = TextNode ""
+buildTextNode value = TextNode (collapseSpace value)
```
ruhatch, line 158: Bit of a nit-pick, but I find it easier to parse the logic here if you just stick to guards:

```suggestion
buildTextNode value 
    | Text.null (Text.strip value) = TextNode ""
    | otherwise = TextNode (collapseSpace value)
```

mpscholten: Agree :) Will change this

PR #253, opened 2020-07-21
[#223] Add support for void tags without trailing slash

### LINKED ISSUE #223: HSX input tag
When using an `<input>` tag without closing (as seen on Bootstrap Documentation) HSX gives an error, searching for the wrong closing tag. 
In my case, using the navbar template from Bootstrap, the issue was fixed with closing input like so `<input type=""/>` but the error was:
```
   |
34 |     </form>
   |      ^
unexpected '/'
expecting "</div>", "</form>", "</input>", "</nav>", identifier, or text
```

ruhatch:
Closes #223


mpscholten:
Works great :-) Thanks 🎉

Issue #255, opened 2020-07-22
toHtml dateField Could not deduce (InputValue Day)


Kaeptenblaubaer:
[14 of 17] Compiling Web.View.Partners.Edit ( Web/View/Partners/Edit.hs, interpreted )

Web/View/Partners/Edit.hs:19:43: error:
• Could not deduce (InputValue Day)
arising from a use of ‘dateField’
from the context: ?viewContext::ViewContext
bound by the type signature for:
renderForm :: Partner -> Html
at Web/View/Partners/Edit.hs:(19,1)-(25,2)
or from: (?viewContext::ViewContext,
?formContext::FormContext Partner)
bound by a type expected by the context:
(?viewContext::ViewContext, ?formContext::FormContext Partner) =>
Text.Blaze.Html.Html
at Web/View/Partners/Edit.hs:(19,38)-(25,2)
• In the first argument of ‘toHtml’, namely
‘(dateField (fromLabel @"validfrom"))’
In the expression: toHtml (dateField (fromLabel @"validfrom"))
In the third argument of ‘Data.MonoTraversable.Unprefixed.foldl'’, namely
‘[toHtml (textField (fromLabel @"vorname")),
toHtml (textField (fromLabel @"name")),
toHtml (dateTimeField (fromLabel @"createdat")),
toHtml (dateField (fromLabel @"validfrom")), ....]’
|
19 | renderForm partner = formFor partner [hsx|
| ^...
Failed, 13 modules loaded.

mpscholten:
Thanks for reporting :) I've prepared a fix in the linked PR. Once this is merged you could update to the latest Git master of IHP and then the issue should  be solved 👍

PR #258, opened 2020-07-23
Added missing InputValue and ParamReader instances for Day type

mpscholten:
Fixes https://github.com/digitallyinduced/ihp/issues/255


mpscholten:
Thanks for reivew 👍

Issue #265, opened 2020-07-25
ihp-new runs out of memory on small hosts (eg EC2 micro instances)


chrislf:
`ihp-new` calls `nix-env -i $package`, which does a large evaluation of nix expressions, which can run out of memory on resource-constrained hosts.

Using `nix-env -iA nixpkgs.$package` is faster and uses much less memory.

chrislf:
I have added a PR which should address this.

I only changed the invocations in ihp-new; there are instances in the github actions config which might benefit as well. I can add these if that is considered a good idea.

mpscholten:
> I only changed the invocations in ihp-new; there are instances in the github actions config which might benefit as well. I can add these if that is considered a good idea.

Would be happy to merge this 👍

mpscholten:
This is fixed via https://github.com/digitallyinduced/ihp/pull/266

PR #266, opened 2020-07-25
use qualified nixpkgs attribute paths to avoid large evaluations

chrislf:
This is to address https://github.com/digitallyinduced/ihp/issues/265

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,6 +16,6 @@ jobs:
         authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
 
     - run: |
-          nix-env -i direnv
+          nix-env -iA nixpkgs.direnv
           nix-env -f https://ihp.digitallyinduced.com/ihp-new.tar.gz -i ihp-new
           ihp-new blog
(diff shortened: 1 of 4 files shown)


mpscholten:
Hey Chris, thanks for the PR 👍 We've tried to similiar changes to the docs a while ago and then found out that `nix-env -iA nixpkgs....` fails on NixOS systems. I just tried it out again on one of our NixOS systems and got the following error:

```bash
[root@haskell:~]# nix-env -iA nixpkgs.cachix
error: attribute 'nixpkgs' in selection path 'nixpkgs.cachix' not found
```

When using `nixos.cachix` instead of `nixpkgs.cachix`, this works:
```bash
[root@haskell:~]# nix-env -iA nixos.cachix
installing 'cachix-0.3.0'
```

Maybe we can have some variable `NIXPKGS_ATTR = if is_nixos then "nixos" else "nixpkgs"` and then use it like `nix-env -iA $NIXPKGS_ATTR.cachix`.

chrislf:
Interesting, I didn't realise it was different on NixOS. I will install on a VM and test with your suggested approach.

mpscholten:
Any updates? :)

chrislf:
I got steamrollered this week: going to give this a shot this evening, however.

chrislf:
I've tested these changes on NixOS (20.03 on VirtualBox), Amazon Linux and Ubuntu 18.04 and they seem to work fine.

For the CI changes, I didn't have a Mac to test the on but I understand the nixpkgs namespace works as for other non-NixOS OSes.

mpscholten:
Looks great, thank you! 🙏

PR #267, opened 2020-07-26
datetime-local field change missing on renderBootstrapFormField

nghamilton:
renderBootstrapFormField updated to use the new datetime-local field type.


mpscholten:
Thanks :)

PR #270, opened 2020-07-27
'make' only shows help; 'make all' needed to create lib directory to ensure GHCI loads Main correctly

nghamilton:
Not sure if this is the intended process, but it was what I had to do to get GHCI loading Main correctly.


mpscholten:
Good catch :)

Issue #276, opened 2020-07-29
Docu Section: Editors & IDEs


mpscholten:
What is the recommended choice?
How to set up common plugins?

PR #281, opened 2020-07-29
Added new docu section "Editors & Tooling"

mpscholten:
Fixes https://github.com/digitallyinduced/ihp/issues/276


mpscholten:
@hendi maybe you could add some words on the emacs setup :)

Issue #285, opened 2020-08-01
Show server errors also in console


mpscholten:
Right now the server errors are only displayed in the http response. The console output just says a general "Internal Server Error" has happend.

PR #287, opened 2020-08-01
Print exception to stdout additionally to rendering it into the browser

mpscholten:
Fixes https://github.com/digitallyinduced/ihp/issues/285


mpscholten:
Good point, haven't thought about that :) Just pushed the change.

Issue #232, opened 2020-07-13
Unique Together is interpreted as a syntax error


kigsmtua:
While editing the raw SQL  on the schema designer, a table defined as follows, fails with a syntax error

```sql
CREATE TABLE user_followers (
    id UUID DEFAULT uuid_generate_v4() PRIMARY KEY NOT NULL,
    user_id UUID NOT NULL,
    follower_id UUID NOT NULL,
    created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL,
    updated_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL,
    UNIQUE(user_id, follower_id)
);

```
Seemingly we cannot add the UNIQUE constraints for two fields as it fails with the below error

```haskell

Application/Schema.sql:18:11:
   |
18 |     UNIQUE(user_id, follower_id)
   |           ^^^^^^^^^^^^^^^^^^
unexpected "(user_id, follower"
expecting "BIGINT", "BINARY", "BOOL", "BOOLEAN", "CHAR(", "CHARACTER VARYING(", "CHARACTER(", "DATE", "DOUBLE PRECISION", "FLOAT4", "FLOAT8", "INT", "INT4", "INT8", "INTEGER", "NUMERIC", "NUMERIC(", "REAL", "TEXT", "TIME", "TIMESTAMP", "TIMESTAMPZ", "UUID", "VARCHAR(", Custom type, identifier, or white space


```

I am not entirely sure if I am missing something.

mpscholten:
Hey, defining a unique constraint this way is not supported at the moment. We can fix this by updating the [SchemaParser](https://github.com/digitallyinduced/ihp/blob/master/IHP/IDE/SchemaDesigner/Parser.hs) :)

kigsmtua:
Aaah I see.. am guessing a P.R for this would be welcome ?

mpscholten:
Yes, would be happy to merge this :)

mpscholten:
@kigsmtua the issue should be fixed with the next IHP release (next release is coming this friday) ✅

PR #290, opened 2020-08-01
Unique constraints

mpscholten:
Fixes https://github.com/digitallyinduced/ihp/issues/232


mpscholten:
We could add a test case for an empty `UNIQUE()`. This should trigger a syntax error. Do you have other cases in mind?

ruhatch:
@mpscholten nope, just asking questions :)

mpscholten:
just added the empty unique test case. I guess then this is ready to go to the master :)

PR #301, opened 2020-08-05
Serial and Bigserial column type

nezbite:


mpscholten:
I think we also need to make changes to the `SchemaCompiler` as the generated code expects the `id` column to be of type `UUID` (you can most likely trigger an error by creating a table with an `SERIAL` column as the primary key). We also need to keep in mind that using `SERIAL` means that the column implicitly gets a `DEFAULT` value (the default value is the next value of the serial). Maybe we can do this in a new seperate PR so that this one can be merged already.

PR #310, opened 2020-08-08
Added ParamReader Float and InputValue Float instances

mpscholten:
This enables float values to be used everywhere where double values previously were possible already

### DIFF
--- a/IHP/Controller/Param.hs
+++ b/IHP/Controller/Param.hs
@@ -18,6 +18,7 @@ import qualified Data.ByteString.Char8 as Char8
 import IHP.ValidationSupport
 import GHC.TypeLits
 import qualified Data.Attoparsec.ByteString.Char8 as Attoparsec
+import qualified GHC.Float as Float
 
 -- | Returns a query or body parameter from the current request. The raw string
 -- value is parsed before returning it. So the return value type depends on what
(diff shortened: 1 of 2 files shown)


IHP/Controller/Param.hs
```diff
@@ -203,6 +204,13 @@ instance ParamReader Double where
             Right value -> Right value
             Left error -> Left ("ParamReader Dobule: " <> cs error)
 
+instance ParamReader Float where
+    {-# INLINE readParameter #-}
+    readParameter byteString =
+        case Attoparsec.parseOnly (Attoparsec.double <* Attoparsec.endOfInput) byteString of
+            Right value -> Right (Float.double2Float value)
```
ruhatch, line 211: Are there situations where this conversion could catch people out?

mpscholten: I've been thinking about this too. It could happen that a user types in a number with lots of precision which can only be represented as a double. In that case a bit of the precision would be lost during conversion. Practically I don't think will cause problems, as usually when you have lots of precision as a requirement you will use double anyway.

Attoparsec sadly doesn't provide a float parser, and we also cannot write the parser ourselves as the required helper function for that is not exported by the attoparsec module.

Issue #302, opened 2020-08-05
SERIAL/BIGINT in SchemaCompiler


nezbite:
I think we also need to make changes to the `SchemaCompiler` as the generated code expects the `id` column to be of type `UUID` (you can most likely trigger an error by creating a table with an `SERIAL` column as the primary key). We also need to keep in mind that using `SERIAL` means that the column implicitly gets a `DEFAULT` value (the default value is the next value of the serial). Maybe we can do this in a new seperate PR so that this one can be merged already.

_Originally posted by @mpscholten in https://github.com/digitallyinduced/ihp/pull/301#issuecomment-669153609_

PR #312, opened 2020-08-09
Added support for SERIAL primary keys

mpscholten:
UUIDs are still the default and recommended way for the id column. For cases where UUIDs don't work, you can now change the column type to SERIAL or BIGSERIAL to work with integers.

Fixes https://github.com/digitallyinduced/ihp/issues/302


mpscholten:
cc @Kaeptenblaubaer @andreasrueb SERIAL PKs are working now on master :)

PR #325, opened 2020-08-12
Github Actions should build latest hash

Laobiz:


mpscholten:
@Laobiz any updates?

Laobiz:
Not yet. I need to test out if the pulling of another repository works without having the tree element error.

PR #329, opened 2020-08-14
[#291] Allow composite primary keys

### LINKED ISSUE #291: Composite primary keys break generation

ruhatch:
Fixes #291

### DIFF
--- a/IHP/IDE/Data/Controller.hs
+++ b/IHP/IDE/Data/Controller.hs
@@ -13,9 +13,10 @@ import IHP.IDE.Data.View.EditValue
 import qualified Database.PostgreSQL.Simple as PG
 import qualified Database.PostgreSQL.Simple.FromField as PG
 import qualified Database.PostgreSQL.Simple.FromRow as PG
+import qualified Database.PostgreSQL.Simple.ToField as PG
 import qualified Database.PostgreSQL.Simple.Types as PG
 import qualified IHP.FrameworkConfig as Config
-import qualified Data.UUID as UUID
+import qualified Data.Text as T
 
 instance Controller DataController where
     action ShowDatabaseAction = do
@@ -27,8 +28,9 @@ instance Controller DataController where
     action ShowTableRowsAction { tableName } = do
         connection <- connectToAppDb
         tableNames <- fetchTableNames connection
+        primaryKeyFields <- tablePrimaryKeyFields connection tableName
 
-        rows :: [[DynamicField]] <- PG.query connection "SELECT * FROM ? ORDER BY id" (PG.Only (PG.Identifier tableName))
+        rows :: [[DynamicField]] <- fetchRows connection tableName
 
         tableCols <- fetchTableCols connection tableName
 
@@ -45,12 +47,13 @@ instance Controller DataController where
         PG.close connection
         render ShowQueryView { .. }
 
-    action DeleteEntryAction { fieldValue, tableName } = do
+    action DeleteEntryAction { primaryKey, tableName } = do
         connection <- connectToAppDb
         tableNames <- fetchTableNames connection
-        let (Just id) = UUID.fromText fieldValue
-        let query = "DELETE FROM " <> tableName <> " WHERE id = ?"
-        PG.execute connection (PG.Query . cs $! query) (PG.Only id)
+        primaryKeyFields <- tablePrimaryKeyFields connection tableName
+        let primaryKeyValues = T.splitOn "---" primaryKey
+        let query = "DELETE FROM " <> tableName <> " WHERE " <> intercalate " AND " ((<> " = ?") <$> primaryKeyFields)
+        PG.execute connection (PG.Query . cs $! query) primaryKeyValues
         PG.close connection
         redirectTo ShowTableRowsAction { .. }
 
@@ -58,7 +61,7 @@ instance Controller DataController where
         connection <- connectToAppDb
         tableNames <- fetchTableNames connection
 
-        rows :: [[DynamicField]] <- PG.query connection "SELECT * FROM ? ORDER BY id" (PG.Only (PG.Identifier tableName))
+        rows :: [[DynamicField]] <- fetchRows connection tableName
 
         tableCols <- fetchTableCols connection tableName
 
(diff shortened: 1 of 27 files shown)


IHP/IDE/SchemaDesigner/Types.hs
```diff
@@ -8,9 +8,9 @@ module IHP.IDE.SchemaDesigner.Types where
 import IHP.Prelude
 
 data Statement
-    = 
+    =
     -- | CREATE TABLE name ( columns );
-    CreateTable { name :: Text, columns :: [Column], constraints :: [Constraint] }
+      StatementCreateTable { unsafeGetCreateTable :: CreateTable }
```
mpscholten, line 13: What is the reason behind this refactoring? :) IMO this new level of indirection makes the code a bit more complex

Test/IDE/SchemaDesigner/ParserSpec.hs
```diff
@@ -253,6 +253,25 @@ tests = do
...
+                    { name = "user_followers"
+                    , columns =
+                        [ col { name = "user_id", columnType = PUUID, notNull = True }
+                        , col { name = "follower_id", columnType = PUUID, notNull = True }
+                        ]
+                    , constraints = [ PrimaryKeyConstraint { columnNames = [ "user_id", "follower_id" ] } ]
+                    }
+
+        -- Should we also have the following failure or is it better to check it outside of the parser?
+        -- it "should fail to parse a CREATE TABLE statement with PRIMARY KEY column and table constraints" do
```
mpscholten, line 267: I would be fine with testing it here :)

IHP/IDE/SchemaDesigner/Compiler.hs
```diff
@@ -22,13 +22,17 @@ compileSql statements = statements
...
 compileStatement CreateEnumType { name, values } = "CREATE TYPE " <> compileIdentifier name <> " AS ENUM (" <> intercalate ", " (values |> map TextExpression |> map compileExpression) <> ");"
 compileStatement CreateExtension { name, ifNotExists } = "CREATE EXTENSION " <> (if ifNotExists then "IF NOT EXISTS " else "") <> "\"" <> compileIdentifier name <> "\";"
 compileStatement AddConstraint { tableName, constraintName, constraint } = "ALTER TABLE " <> compileIdentifier tableName <> " ADD CONSTRAINT " <> compileIdentifier constraintName <> " " <> compileConstraint constraint <> ";"
 compileStatement Comment { content } = "-- " <> content
 compileStatement UnknownStatement { raw } = raw
 
+compilePrimaryKeyConstraint :: PrimaryKeyConstraint -> Text
+compilePrimaryKeyConstraint PrimaryKeyConstraint { primaryKeyColumnNames } =
+    "PRIMARY KEY(" <> intercalate ", " primaryKeyColumnNames <> ")"
```
mpscholten, line 34: Maybe we can also add some test cases to the CompilerSpec to cover this new function :)

IHP/SchemaCompiler.hs
```diff
@@ -49,36 +50,41 @@ fullCompileOptions = CompilerOptions { compileGetAndSetFieldInstances = True }
...
-        actualType =
-            case findForeignKeyConstraint table column of
-                Just (ForeignKeyConstraint { referenceTable }) -> "(" <> primaryKeyTypeName' referenceTable <> ")"
-                _ -> atomicType
-    in
-        if not notNull
-            then "(Maybe " <> actualType <> ")"
-            else actualType
+atomicType :: PostgresType -> Text
+atomicType = \case
```
mpscholten, line 54: 👍 good job here :) much cleaner now

IHP/SchemaCompiler.hs
```diff
@@ -444,38 +454,36 @@ toDefaultValueExpr Column { columnType, notNull, defaultValue = Just theDefaultV
...
-        (Just idColumn) = find (get #primaryKey) columns
-
         idType :: Text
-        idType = case get #columnType idColumn of
-                PUUID -> "UUID"
-                PSerial -> "Int"
-                PBigserial -> "Integer"
-                otherwise -> error ("Unexpected type for primary key column in table" <> cs name)
+        idType = case primaryKeyColumns table of
+            [] -> error "Impossible happened in compilePrimaryKeyInstance"
```
mpscholten, line 468: Maybe we can improve this error message a little bit `Impossible happened in compilePrimaryKeyInstance. No primary keys found for table $tableName. Atleast one primary key is required.`. The impossible always happens at some point :D 

Btw: I've written down a few notes on IHP error message design goals here https://github.com/digitallyinduced/ihp/wiki/Design-Goals an hour ago. Let's strive to be the haskell package with the best error messages 👍

IHP/SchemaCompiler.hs
```diff
@@ -532,6 +540,31 @@ compileUpdateFieldInstances table@(CreateTable { name, columns }) = unlines (map
                 compileTypePattern' ::  Text -> Text
                 compileTypePattern' name = tableNameToModelName (get #name table) <> "' " <> unwords (map (\f -> if f == name then name <> "'" else f) (dataTypeArguments table))
 
+compileHasFieldId :: (?schema :: Schema) => CreateTable -> Text
+compileHasFieldId table@CreateTable { primaryKeyConstraint } = cs [i|
+instance HasField "id" Picture (Id' "pictures") where
```
mpscholten, line 545: I think we need to replace `Picture` here with the model name :)

IHP/SchemaCompiler.hs
ruhatch: @mpscholten Hmm, I've actually found out that SQL doesn't require primary keys on all tables. Is that something we want to allow?

mpscholten: For now it's easier to just disallow this. Maybe we can skip tables without primary key in the code generation. This way you can still define tables without PK in your `Schema.sql`, but have to use plain sql queries to access them (like using  `sqlQuery "SELECT * FROM my_table_without_pk"`)

ruhatch:
Phew, that was a big one, but it's ready for a final look through @mpscholten

IHP/IDE/Data/Controller.hs
```diff
@@ -27,8 +28,9 @@ instance Controller DataController where
     action ShowTableRowsAction { tableName } = do
         connection <- connectToAppDb
         tableNames <- fetchTableNames connection
+        primaryKeyFields <- tablePrimaryKeyFields connection tableName
 
-        rows :: [[DynamicField]] <- PG.query connection "SELECT * FROM ? ORDER BY id" (PG.Only (PG.Identifier tableName))
+        rows :: [[DynamicField]] <- fetchRows connection tableName
```
mpscholten, line 33: By using `fetchRows` here, we lose the `ORDER BY id`. This has been added here to avoid rows jumping around when using the data editor when updating a value of a row.

When we don't order by the primary key, we will get the results in the order they're read by postgres from the disk. This order usually changes when doing running an `UPDATE` statement. Maybe we can query the primary keys before doing this query and then add them in a `ORDER BY` clause?

mpscholten: Forgot what I've written above. Just saw that `fetchRows` is already doing that 👍  very good job

Test/IDE/SchemaDesigner/CompilerSpec.hs
```diff
@@ -200,77 +194,90 @@ tests = do
...
-                            , primaryKey = False
                             }
                         ]
+                    , primaryKeyConstraint = PrimaryKeyConstraint []
                     , constraints = []
                     }
             compileSql [statement] `shouldBe` sql
 
         it "should compile a CREATE TABLE statement with a multi-column UNIQUE (a, b) constraint" do
-            let sql = cs [plain|CREATE TABLE user_followers (\n    id UUID DEFAULT uuid_generate_v4() PRIMARY KEY NOT NULL,\n    user_id UUID NOT NULL,\n    follower_id UUID NOT NULL,\n    UNIQUE(user_id, follower_id)\n);\n|]
```
mpscholten, line 248: I'd rather like to keep the existing behavior for single primary key columns. Can we add a special case for that so that it stays like `id UUID ... PRIMARY KEY` instead of appending `PRIMARY KEY (id)` at the end? The first approach feels more "human" to me (meaning it would more fit the code a normal human would write). The new approach feels a bit too much like a code generator :D I know this is very subjective, but I think we need to get the small details right :)

mpscholten (commented):
Looks very good! :) Thanks for your work 👍

Test/IDE/SchemaDesigner/CompilerSpec.hs
ruhatch: Done!

mpscholten: Cool thanks :) Just in time for the next IHP release today 👍

mpscholten (approved):

mpscholten:
Very good job, thanks again 👍

PR #338, opened 2020-08-17
Column deletion removes assigned foreign key constraints

nezbite:


IHP/IDE/SchemaDesigner/Controller/Columns.hs
```diff
@@ -101,6 +101,9 @@ instance Controller ColumnsController where
         statements <- readSchema
         let tableName = param "tableName"
         let columnId = param "columnId"
+        let columnName = param "columnName"
+        case findForeignKey statements tableName columnName of
+            Just AddConstraint { constraintName, .. } -> updateSchema (deleteForeignKeyConstraint constraintName)
```
mpscholten, line 106: This will fail when there is not foreign key constraint as the `case` is only handling the `Just AddConstraint` case

mpscholten (approved):

PR #344, opened 2020-08-20
Danielsonnenberg/point bytea tsvector

nezbite:


IHP/IDE/SchemaDesigner/View/Columns/Edit.hs
```diff
@@ -113,6 +113,9 @@ typeSelector postgresType enumNames = [hsx|
             {option selected "Time" "Time"}
             {option selected "SERIAL" "Serial"}
             {option selected "BIGSERIAL" "Bigserial"}
+            {option selected "POINT" "Point"}
```
mpscholten, line 116: While testing I got this error in the generated haskell code:

Not in scope: type constructor or class ‘Point’

fegu:
Is still PR stalled? I will be needing both point and tsvector in two different projects soon. Should we fix this PR or make a new one?

mpscholten:
I think we need to update this, then we could merge

mpscholten:
Point has been merged to master via https://github.com/digitallyinduced/ihp/commit/958d553c028b583d7a057dc66f4ee305569ee621

PR #345, opened 2020-08-21
HSX attribute names are now parsed differently for improved error mes…

### LINKED ISSUE #343: dashes in html/xml attribute names break hsx parsing
For example `fill-rule=""` in the following piece of SVG, breaks the parser

```
...
  <path fill-rule="evenodd" d="M16 8A8 8 0 1 1 0 8a8 8 0 0 1 16 0zM8.5 4a.5.5 0 0 0-1 0v3.5H4a.5.5 0 0 0 0 1h3.5V12a.5.5 0 0 0 1 0V8.5H12a.5.5 0 0 0 0-1H8.5V4z"/>
</svg>
```

mpscholten:
…sages

We also added lots of svg attribute names to the attribute name whitelist.
Fixes #343.

### DIFF
--- a/IHP/HtmlSupport/Parser.hs
+++ b/IHP/HtmlSupport/Parser.hs
@@ -124,15 +124,20 @@ hsxNodeAttribute = do
     pure (StaticAttribute key value)
 
 hsxAttributeName :: Parser Text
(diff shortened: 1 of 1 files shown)


mpscholten:
@themmes this should fix the svg issue in #343 :)

Issue #333, opened 2020-08-16
uuid-ossp missing - Update DB fails


2mol:
[The documentation](https://ihp.digitallyinduced.com/Guide/your-first-project.html) mentions the line

```
CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
```

in `Application/Schema.sql `

This was not there by default when I first clicked "Update DB", which resulted in an error along the lines of

```
ERROR: function uuid_generate_v4() does not exist HINT: No function matches the given name and argument types
```

I'm using the latest IHP version. Adding it manually fixed the problem.

mpscholten:
Hey, I cannot reproduce this issue with the latest IHP version. To avoid this kind of issue happening again we've moved this DB init step into the framework code (it's a technical implementation detail anyway, doesn't really belong to the app code). Therefore we can be safe that this setup code is not lost while making changes to the file :)

PR #354, opened 2020-08-24
Moved uuid-ossp extension setup to IHP from the app specific Schema.sql

mpscholten:
Fixes https://github.com/digitallyinduced/ihp/issues/333

This fixes the above issue by introducing a new `IHPSchema.sql` which is loaded before the app specific `Schema.sql`. This way we can be sure the extension is set.

After the new IHP version with this is released, [we can remove the uuid setup code from the boilerplate](https://github.com/digitallyinduced/ihp-boilerplate/blob/master/Application/Schema.sql#L1).


ruhatch (approved):
Looks good to me

PR #362, opened 2020-08-30
Allow dotless email addresses in `isEmail`

tchoutri:
Although their usage are prohibited by ICANN for the wider internet, dotless email addresses are perfectly valid as per the standards. I included a warning leading to the announcement made in 2013 by ICANN regarding the usage of dotless email addresses.

references:
* https://en.wikipedia.org/wiki/Email_address#Examples
* https://www.icann.org/news/announcement-2013-08-30-en


mpscholten:
Good catch 👍

PR #363, opened 2020-08-30
Improved error message when 'param ".."' is called and the param is not found

mpscholten:
This PR improves the error message when a call to param fails. The new error screen looks like this:

image

### DIFF
--- a/IHP/ErrorController.hs
+++ b/IHP/ErrorController.hs
@@ -148,8 +148,7 @@ patternMatchFailureHandler exception controller additionalInfo = do
 paramNotFoundExceptionHandler :: (Show controller, ?requestContext :: RequestContext) => SomeException -> controller -> Text -> Maybe (IO ResponseReceived)
 paramNotFoundExceptionHandler exception controller additionalInfo = do
     case fromException exception of
-        Just (exception :: Param.ParamNotFoundException) -> Just do
-            let (Param.ParamNotFoundException paramName) = exception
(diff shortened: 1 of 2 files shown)


IHP/ErrorController.hs
```diff
@@ -143,6 +144,51 @@ patternMatchFailureHandler exception controller additionalInfo = do
             respond $ responseBuilder status500 [(hContentType, "text/html")] (Blaze.renderHtmlBuilder (renderError title errorMessage))
         Nothing -> Nothing
 
+-- Handler for 'IHP.Controller.Param.ParamNotFoundException'
+paramNotFoundExceptionHandler :: (Show controller, ?requestContext :: RequestContext) => SomeException -> controller -> Text -> Maybe (IO ResponseReceived)
+paramNotFoundExceptionHandler exception controller additionalInfo = do
+    case fromException exception of
+        Just (exception :: Param.ParamNotFoundException) -> Just do
+            let (Param.ParamNotFoundException paramName) = exception
```
ruhatch, line 152: Does this work:
```suggestion
        Just (exception@(Param.ParamNotFoundException paramName)) -> Just do
```
or is this a quirk with `fromException`?

ruhatch (approved):
Looks really pretty :+1:

IHP/ErrorController.hs
mpscholten: Good point, just changed it 👍

Issue #361, opened 2020-08-28
js-delete breaks when html anchor is not the leaf element


themmes:
I just stumbled upon this problem when creating a delete <button>

```
<a href="DeleteTodoAction (get #id todo)">
  <button>
    Delete
  </button>
</a>
```

In this case actually the `event.target` is not the `anchor`, but the `button`. I verified this by removing the HTMLAnchor check (here: ) and adding a `js-delete` to the button.

I was going to write a PR to remove the HTMLAnchor check, but that does not work. It uses the href from that specific element which does not exist on the button. 

Alternatively, we could recursively check parents for the `js-delete` class, however maybe you guys know of a better way to solve this?

themmes:
I think I wrote a fix using `event.target.closest()`, however I am not sure how to apply this change to my project. Is there some documentation on how to contribute on the javascript code(and haskell, documentation for that matter)?

mpscholten:
Good catch :) I think there's [`event.currentTarget`](https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget) which always refers to the element the event listener was registered to (in our case the `a`).

To make changes to the JS, follow the steps in https://github.com/digitallyinduced/ihp/blob/master/CONTRIBUTING.md#running-an-application-against-a-local-copy-of-the-framework

In short it should be like this:

```bash
cd $YOUR_PROJECT
git clone git@github.com:digitallyinduced/ihp.git IHP
make -B .envrc
make -B build/ihp-lib

# Now it should be ready to pick up the local IHP source instead of the release version
./start
```

Then you can make changes to `$YOUR_PROJECT/IHP/lib/IHP/static/helpers.js` :)

themmes:
Hmm.. I might be mixing up stuff here. When I run `./start` after following your instructions I get a compilation error 

```
build/Generated/Types.hs:51:14
    • The constructor ‘ModelContext’ should have 2 arguments, but has been given 1
    • In the pattern: ModelContext conn
      In a pattern binding: (ModelContext conn) = ?modelContext
```

Is this an error because I'm now running the latest master or because I am mixing up versions?

themmes:
Have not been able to finish this, after my holidays (2 weeks) I'll pick this up!

mpscholten:
> Hmm.. I might be mixing up stuff here. When I run `./start` after following your instructions I get a compilation error
> 
> ```
> build/Generated/Types.hs:51:14
[...]

PR #364, opened 2020-08-30
Attempt to fix js-delete

themmes:
Fix for https://github.com/digitallyinduced/ihp/issues/361, changes..

- add `handleClick` only to elements with `js-delete` to limit scope to `<a>` (instead of `document`)
- change `target` to `currentTarget`

This seems to work, however I get issues when trying to delete right after being routed from `New` or `Edit`. For some reason it requires a refresh of the page before working correctly. Maybe you can verify whether that is due to my project?


mpscholten:
Thanks! Looks good so far :)

By using `document.querySelectorAll(".js-delete")` the listener on attachs to elements available at the first page load. When using a form or link, we're using turbolinks for navigation (it's part of the default IHP setup), therefore we have to reattach the listener after turbolinks applied changes to the DOM. We're doing this for all the other helpers [already here in line 11](https://github.com/digitallyinduced/ihp/blob/master/lib/IHP/static/helpers.js#L11). Due to use of morphdom we also need to make sure that we don't attach our event listener twice (the DOM node might still be the same because of the diff approach by morphdom). This might be tricky :)

lib/IHP/static/helpers.js
```diff
@@ -42,20 +42,20 @@ function initTime() {
 }
 
 function initDelete() {
-    document.addEventListener('click', handleClick);
+    document.querySelectorAll(".js-delete").forEach(e => e.addEventListener('click', handleClick));
```
mpscholten, line 45: Could we use boring JS here? :) This way these helpers support even older browsers. (Only in the JS files used by the dev tools, we can use more modern JS)

```suggestion
    document.querySelectorAll(".js-delete").forEach(function (e) { e.addEventListener('click', handleClick) });
```

themmes: Makes sense!

themmes:
That should do it!

themmes:
There might still be something weird, if you cancel the confirmation prompt it asks you a second time. Any clue why that happens?

mpscholten:
I guess the callback is registered multiple times on the link due to turbolinks events being triggered multiple times. You could add a `console.log` to the callback in  `document.querySelectorAll(".js-delete").forEach(function(e) {` to verify this :)

themmes:
Have not been able to finish this, after my holidays (2 weeks) I'll pick this up!

themmes:
@mpscholten I'm sorry, I have tried but I am not able to confirm or exclude it's caused by turbolinks.

themmes:
@mpscholten I feel like handing in my programmers-license, but I really can't lay my finger on this issue. It's black magic to me 🧙 !

mpscholten:
:) no worries, will take a look on monday and maybe can provide a fix 👍

PR #374, opened 2020-09-06
Update controller.markdown

fegu:
Typo in name of function for responding with a file as a controller action.


mpscholten (approved):
Good catch :) Thanks 👍

PR #376, opened 2020-09-06
Fix Grammar

CharlesTaylor7:
This is a very interesting framework!
I'm just getting started with it this afternoon, but figured I could still make this minor fix.


mpscholten (approved):
Good catch :) Thanks 🎉

Issue #93, opened 2020-06-05
Error installing Nix in Mac with FileVaut encrypted volumes


pupoxcu:
Following the install guide executed: `sh <(curl https://nixos.org/nix/install) --darwin-use-unencrypted-nix-store-volume` then it returned: `error: refusing to create Nix store volume because the boot volume is FileVault encrypted, but encryption-at-rest is not available. Manually create a volume for the store and re-run this script. See https://nixos.org/nix/manual/#sect-macos-installation`

mpscholten:
What macbook version are you using?

Can you try these steps (take from https://github.com/NixOS/nix/issues/2925#issuecomment-539490866):

```
sudo mkdir /System/Volumes/Data/opt/nix
sudo chown {your_user} /System/Volumes/Data/opt/nix
sudo sh -c "echo 'nix\tSystem/Volumes/Data/opt/nix' >> /etc/synthetic.conf"
```
Reboot mac. Then:
```
export NIX_IGNORE_SYMLINK_STORE=1
curl https://nixos.org/nix/install | sh     
```

After that, add the following to `~/.zprofile`:

```
. ~/.nix-profile/etc/profile.d/nix.sh
export NIX_IGNORE_SYMLINK_STORE=1
```

I've just tried these steps on a clean 2018 macbook pro (where the above nix installation error message also appeared), and this fixed the problem.

ajkovar:
This didn't work for me as is.  I had to add the -L argument to curl.  I guess the page moved? 🤷‍♂️

   export NIX_IGNORE_SYMLINK_STORE=1
   curl https://nixos.org/nix/install | sh

jantebeest:
Doesn't work on MacOS 11, the check looks to be insufficient as it tells me to _upgrade_ to 10.x

mpscholten:
What commands/approach did you try? And what's the error message?

jantebeest:
```➜  ~ sh <(curl -L https://nixos.org/nix/install) --darwin-use-unencrypted-nix-store-volume
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:00:02 --:--:--     0
100  2490  100  2490    0     0    771      0  0:00:03  0:00:03 --:--:--   771
[...]

PR #377, opened 2020-09-06
Replace make .envrc with lorri

### LINKED ISSUE #339: Try to replace our .envrc approach with lorri

ruhatch:
Closes #339


lib/IHP/Makefile.dist
```diff
@@ -88,11 +88,6 @@ endif
 
 all: build/ihp-lib build/Generated/Types.hs
 
-.envrc: default.nix ## Rebuild nix packages and .envrc
```
mpscholten, line 91: Maybe we can add helpful fallback like

```make
.envrc: 
    echo "The '.envrc' approch has been replaced with 'lorri'. Use 'lorri watch --once' instead." 
```

lib/IHP/Makefile.dist
```diff
@@ -134,7 +129,6 @@ clean: ## Resets all build and temporary files
 	rm -rf build/bin
 	rm -rf IHP/IHP/static/node_modules
 	rm -rf build/db
-	rm -f .envrc
```
mpscholten, line 137: I'd still keep this line for a while. Otherwise when migrating from an older IHP version a clean rebuild using `make clean` will not fix all problems.

ruhatch:
@mpscholten this is ready to review, but see digitallyinduced/ihp-boilerplate#5 for details before merging!

mpscholten:
Code looks very good so far :)

I'm currently hit by https://github.com/target/lorri/issues/248 

This is related to having `/nix` symlinked on macOS. The IHP documentation refers to https://github.com/digitallyinduced/ihp/issues/93#issuecomment-639611313 for the macOS installation, where I recommend the symlinking approach. So I guess a lot of people using IHP on macOS will be hit with that issue. I think we also need to solve that somehow by migrating people of the symlink approach to something else.

lib/IHP/prepare-lorri
```diff
@@ -0,0 +1,54 @@
...
+    case "$choice" in
+        y|Y ) cp master-boilerplate/start .;;
+        * ) echo -e "\e[31mPlease update start manually from https://github.com/digitallyinduced/ihp-boilerplate and then re-run this program.\e[0m"; exit 1;;
+    esac
+fi
+
+rm -rf master-boilerplate
+
+lorri init
+direnv allow
```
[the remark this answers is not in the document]
ruhatch: We can run `lorri watch --once`, which is what will happen in the `ihp-new` path. I'll move the call to `lorri watch --once` from `ihp-new` into the `prepare-lorri` script.

Can you try the above steps with `lorri watch --once` after `lorri init` to see if that is better?

ruhatch:
> Code looks very good so far :)
> 
> I'm currently hit by [target/lorri#248](https://github.com/target/lorri/issues/248)
> 
> This is related to having `/nix` symlinked on macOS. The IHP documentation refers to [#93 (comment)](https://github.com/digitallyinduced/ihp/issues/93#issuecomment-639611313) for the macOS installation, where I recommend the symlinking approach. So I guess a lot of people using IHP on macOS will be hit with that issue. I think we also need to solve that somehow by migrating people of the symlink approach to something else.

Ah, that's annoying. Is there another approach to use for now or do we have to wait for better Catalina support?

mpscholten:
I see two possible solutions:
1. we build a migrate script that migrates the symlink approach on macos to something else which works better. There was a ruby script in the nix github issue related to macos which got it to work reliable (i think the ruby script was build and used by shopify). This way we could also fix https://github.com/digitallyinduced/ihp/issues/93 It's a hard problem to solve.
2. we change lorri to work differently on macos (e.g. maybe we can patch lorri so that lorri somehow rewrites the nix file that causes the error)

ruhatch:
> I see two possible solutions:
> 
>     1. we build a migrate script that migrates the symlink approach on macos to something else which works better. There was a ruby script in the nix github issue related to macos which got it to work reliable (i think the ruby script was build and used by shopify). This way we could also fix #93 It's a hard problem to solve.
> 
>     2. we change lorri to work differently on macos (e.g. maybe we can patch lorri so that lorri somehow rewrites the nix file that causes the error)

Okay, I work on Linux and don't really have the energy for this bit. Can I leave it with you or someone else at di?

mpscholten:
Yes I’ll take care of this. I definitely want this to be merged as I think lorri is a lot better than our current workflow.

Thanks again for your work so far :-)

mpscholten:
We can likely replace lorri with nix flakes soon. See https://github.com/digitallyinduced/ihp/issues/1552

PR #381, opened 2020-09-07
Update database.markdown

fegu:
Two things:
1: Add manual invocation of psql for all those (like me) where make psql does not work (and for those who like to see how stuff works).
2: Removed reference to create uuid extension line in Schema.sql which since 20200904 is not there any more (now automatically loaded by IHP).


mpscholten (approved):
Thanks, good addition 🙌

PR #382, opened 2020-09-07
Update your-first-project.markdown

fegu:
1: The tutorial telles the user to remove "Show" link in section "Display all posts", but then refers to the "Show" link a bit further down. Fixed by changing the latter text to not refer to the Show link, but to the now clickable title.
2: Text of query (technical representation of query) is now shown as SQL, looks sufficiently different to warrant a helpful parenthesis for newbies.


mpscholten (approved):
Perfect, thanks! :)

PR #386, opened 2020-09-11
Support for nested ORDER BYs

fegu:
Currently this code:
|> orderBy #lastname
|> orderBy #firstname
has a bug: it will order by the last clause (#firstname) only.
This commit adds support for such nested order bys.


mpscholten (approved):
Thanks, looks good :)

PR #388, opened 2020-09-13
Fix paramOrDefault documentation

ivanov:
Thanks for working on IHP!

It seems like perhaps the default value used to come after the parameter name and the docs still have this old flipped order.


mpscholten (approved):
Thanks, good catch :)

Issue #390, opened 2020-09-15
static/helpers.js dateTime producing incorrect output for locales with 12-hour time


dansvo:
I'm in a locale where `.toLocaleTimeString()` produces output like `9:20:50 PM`.

Currently at the tip of master, dateTime produces a time stamp without seconds by cutting down to 5 characters
https://github.com/digitallyinduced/ihp/blob/e6f0bcd037e8c3cbe8a74c4bf8db663d2def14ef/lib/IHP/static/helpers.js#L33-L35

This produces the output `9:20:` for this locale. It should instead read something like `9:20 PM`.

mpscholten:
Fixed via https://github.com/digitallyinduced/ihp/pull/391

PR #391, opened 2020-09-15
update dateTime so that it works correctly for 12-hour locales

dansvo:
Hi! I'd like to propose this fix for https://github.com/digitallyinduced/ihp/issues/390, an issue I encountered when using dateTime in a 12-hour locale. This change will produce canonical time stamps with hours, minutes, and without seconds, for both 12- and 24-hour locales.


mpscholten (approved):

mpscholten:
Thanks for the fix :) good job 👍

PR #403, opened 2020-09-19
Auth: normalize login error message

### LINKED ISSUE #402: Authentication failure message should not be different for wrong password/username
## Description

Currently, the `createSessionAction` in the default Sessions Controller (`IHP/AuthSupport/Controller/Sessions.hs`) differentiates between the case that the password is wrong and the user not being found. This differentiation is also communicated to the client via an on page error message.

## Possible fix

Set the error message in both cases to something identical such as `Invalid Credentials`

## Why is this a problem?
An attacker could use the error message to enumerate email addresses which is a privacy concern as well as a minor security concern.

HugoPeters1024:
closes #402


mpscholten (approved):

mpscholten:
Thanks :)

PR #404, opened 2020-09-19
codegen: import Generated.Types in layout view

HugoPeters1024:
By importing that module by default, developers can easily use types like the User when rendering the Layout view.


mpscholten (approved):
Thanks :) 👍

PR #405, opened 2020-09-20
Disable all fancy debugging error handlers in production

mpscholten:
We don't want to display all the fancy error messages which sometimes include sql queries for normal users.


ruhatch (approved):
Nice, looks good!

mpscholten:
Thx for fast review 👍

Issue #389, opened 2020-09-14
Add a paramList function


mpscholten:
See https://stackoverflow.com/questions/63875081/how-can-i-pass-list-params-in-ihp-forms/63879113#63879113

```
        let paramList name =
                allParams
                |> filter (\(paramName, paramValue) -> paramName == name)
                |> mapMaybe (\(paramName, paramValue) -> paramValue)
                |> map (readParameter @p)
                |> map (Either.fromRight (error $ "param: Parameter " <> name <> " is invalid"))
```

PR #411, opened 2020-09-21
Added paramList function to deal with multiple params with the same name

mpscholten:
Fixes https://github.com/digitallyinduced/ihp/issues/389


HugoPeters1024 (approved):
Haven't actually tested it but the implementation looks very sound.

mpscholten:
Thanks for review :) 👍

PR #414, opened 2020-09-22
WIP: add customisable 404 page renderer

HugoPeters1024:
## Summary

**This is a work in progress**

Here is a proposition for a method of allowing for custom 404 pages addressing issue #367. The newly created typeclass could be extended in the future to support custom views for other error cases as well.

## What still needs to be done

When generating a new project, `Config/Config.hs` should add the following import:

```haskell
import IHP.ErrorController (ErrorController, renderNotFound)
```

and the following instance declaration (will result in default behavior):

```haskell
instance ErrorController
```

## Example of the functionality

If the developer is so inclined he/she can now override the default renderer in `Config/Config.hs`. For example:

```haskell
instance ErrorController where
    renderNotFound title body = H.docTypeHtml ! A.lang "en" $ [hsx| 
        <h1>Welcome to Minimalism Inc.</h1>

        <p>This page does not seem to exist, our software says: </p>

        <p class="error">{body}</p>
    |]
```

## Extra notes

- This update would not be backwards compatible with existing projects (although it is an easy fix)
- I decided to add the render function as an argument to the ErrorController handler, directly binding the typeclass to the context would work but required a cascading number of context binds in parent functions that generated a lot of extra code.
- Ideally the ErrorController would fall back on an existing view, but I did not see how to implement such functionality at this time.

Please let me know if you see any improvements or have any feedback.

### DIFF
--- a/IHP/AuthSupport/Controller/Sessions.hs
+++ b/IHP/AuthSupport/Controller/Sessions.hs
@@ -93,7 +93,7 @@ createSessionAction = do
                     redirectUrl <- getSessionAndClear "IHP.LoginSupport.redirectAfterLogin"
                     redirectToPath (fromMaybe (afterLoginRedirectPath @record) redirectUrl)
                 else do
-                    setErrorMessage "Incorrect Password"
+                    setErrorMessage "Invalid Credentials"
                     user :: record <- user
                             |> incrementField #failedLoginAttempts
                             |> updateRecord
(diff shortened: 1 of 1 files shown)


mpscholten:
To support all other common operations you do in a controller (like e.g. working with the layouts), I'd rather make this a real `Controller`, like:

```haskell
data ErrorController = NotFoundAction | AccessDeniedAction | GenericErrorAction
instance Controller ErrorController where
    notFoundAction = render NotFoundView { .. }
```

This way the error controller would be similiar to all existing controllers and feels more consistent :)

PR #423, opened 2020-09-24
Custom error page (requires discussion)

HugoPeters1024:
### Summary
- A default error controller is generated as part of a new project (using the default error view)
- The runApp function tries to lookup the /NotFound action before defaulting to previous behavior

### Known problems
- [ ] Fix the layout (the bg color is not working, probably through the addition of the layout default css)

### Discussion points
- A default project now contains extra files. Is this desirable or is there a way to implement default behavior in the package rather than a fresh project?
- This change is not backwards compatible, is that a problem? What is our policy on this matter?

### DIFF
--- a/IHP/ErrorController.hs
+++ b/IHP/ErrorController.hs
@@ -8,6 +8,7 @@ module IHP.ErrorController
 , handleNoResponseReturned
 , handleNotFound
 , handleRouterException
+, renderErrorView
 ) where
 
 import IHP.Prelude hiding (displayException)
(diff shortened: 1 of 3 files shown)


mpscholten:
I've been thinking about this a bit more from a user perspective. Basically the use case we want to solve is to customize the 404 or 501 error pages in production.

I've been looking into how rails does this. They just have a 404.html or 501.html in the `public` directory. IMO we should follow a similiar approach. Having a new error controller in each new app is a very complex solution for most people who just want to have a nice 404 page :)

So I suggest we should move the `ErrorController` from the app into the framework itself. Then inside the view we should check whether `static/404.html` exists and then render that file (using `renderFile`). We can do similiar for a custom internal server error page.

This also largely solves the discussion points.

> A default project now contains extra files. Is this desirable or is there a way to implement default behavior in the package rather than a fresh project?

IMO we should strive to keep the IHP project boilerplate as small as necessary for a normal project.

> This change is not backwards compatible, is that a problem? What is our policy on this matter?

It should always be easy to update to the latest IHP version. Atleast at the current two-week release pace. At di we're currently running multiple different IHP versions in prod. So we're a bit biased into that direction :D  We should only do major BC breaks when absolutely necessary or when we're doing some kind of major release.

What do you think of this?

HugoPeters1024:
Now that I think about it, it probably is backwards compatible. But that does not take away the fact that a default 404.html in the public folder is a far more simple idea. Moving forward it might be nice to set up a policy on what guarantees minor and major version changes have.

PR #437, opened 2020-10-01
Added support for postgres array types

### LINKED ISSUE #431: Improved support for array types
> I wasn't able to save Schema.sql without an error whether I use, e.g., pay_by_quarter integer[] or pay_by_quarter integer ARRAY

mpscholten:
Fixes #431


HugoPeters1024 (commented):
Would it make sense to add a test that retrieves an array value from the database as well?

mpscholten:
Agree, but for that we'd need a new test suite. Basically we would need to create a test project that uses most IHP features. Then we could try to curl a few endpoints and check that they work. Just created https://github.com/digitallyinduced/ihp/issues/439 for this

PR #442, opened 2020-10-02
Added support for Postgresql OFFSET clause

fegu:
The OFFSET clause is useful for implementing paging of large datasets. It is most often used together with the LIMIT clause.


fegu:
missed a clause in isSimpleQuery. Would probably not cause problems but will fix.

PR #443, opened 2020-10-02
Adding support for Postgres OFFSET clause

fegu:
The OFFSET clause is useful for implementing paging of large datasets. It is most often used together with the LIMIT clause.


mpscholten (approved):
Perfect, thanks 👍  :)

PR #444, opened 2020-10-02
Fix/bool params

nghamilton:
A bool param with a value of 'true' fails to parse successfully; this could be considered unexpected behaviour?


mpscholten (approved):
Yes 👍  good addition, thanks! :)

Issue #401, opened 2020-09-19
SESSION cookie should be http-only by default


HugoPeters1024:
By setting up authentication as described by the guide, a SESSION cookie is used. However, this cookie is not marked as http-only. Enabling that option prevents the value of this cookie to be queried by any javascript, limiting the attack surface for any cross site scripting attacks. 

some more detailed information about the http-only flag: https://owasp.org/www-community/HttpOnly

mpscholten:
Good point 👍 The cookie is currently configured here https://github.com/digitallyinduced/ihp/blob/master/IHP/Server.hs#L42

mpscholten:
We could also move the cookie config code into a new function `buildSessionCookie` into [`FrameworkConfig`](https://github.com/digitallyinduced/ihp/blob/master/IHP/FrameworkConfig.hs#L42) (like we do with `requestLoggerMiddleware`). Then inside the application config users could configure the session cookie (e.g. setting a custom lifetime).

HugoPeters1024:
Ideally, the secure flag should also be set to enforce an https connection before sending the value of the cookie. But that should be enabled in production only. Possibly it could be enabled in development but that would require messing around with self signed developer certificates or something of the like, not sure if that should be a priority.

PR #445, opened 2020-10-03
Session Cookie: HttpOnly + secure

mpscholten:
Fixes https://github.com/digitallyinduced/ihp/issues/401

@HugoPeters1024 could you take a look? :)


HugoPeters1024 (approved):
That does the job :) ( can't test it currently because my laptop is grinding compiling that haskell language server)

mpscholten:
Perfect, thanks for fast review :)

PR #449, opened 2020-10-05
updated picture upload docs to handle null cases.

Montmorency:
Small change to docs for recipe on upload picture so that the picture_url is Nullable.


mpscholten (approved):
Thanks :)

PR #453, opened 2020-10-08
Adding instances for Ord and Hashable to PrimaryKeys

fegu:
This allows for use of PrimaryKeys as keys in data structures such as Data.Map and Data.Hashable. When working on older databases without defined relations, or for cases where native joins (or fetchRelated) performs poorly, the ability to manually look up data based on key is useful.


fegu:
Got mangled with an unrealted commit

PR #454, opened 2020-10-08
Adding instances for Ord and Hashable to PrimaryKeys

fegu:
This enables use of Data.Map and Data.Hashable on primary keys. This is handy for working with older DBs, or where joins (or fetchRealted) performs badly.


mpscholten (approved):
Thanks :)

PR #457, opened 2020-10-10
Bugfix for creating view together with action

fegu:
Creating a view together with an action does not actually work. The view is not created even though the preview shows it should. These changes fixes the bug. And a related bug giving the view the wrong name (missing call to ucfirst).


fegu:
I added two other things as well: generated Show.hs now actually shows data (used to be empty). Generated Edit.hs and New.hs now has parenthesis around text fields because it is easy to forget these when adding field modifiers later.

mpscholten (approved):
good job, thanks :)

PR #459, opened 2020-10-11
Added more tests

mpscholten:


Test/HaskellSupportSpec.hs
```diff
@@ -0,0 +1,18 @@
...
+import IHP.HaskellSupport
+
+tests = do
+    describe "HaskellSupport" do
+        describe "stripTags" do
+            it "should deal with empty input" do
+                stripTags "" `shouldBe` ""
+
+            it "should strip html tags and return the plain text" do 
+                stripTags "This is <b>Bold</b>" `shouldBe` "This is Bold"
```
HugoPeters1024, line 18: Is this functionality only meant as a utility, or is it used as a actual sanitizer?

Test/HtmlSupport/QQSpec.hs
```diff
@@ -0,0 +1,72 @@
...
+module Test.HtmlSupport.QQSpec where
+
+import Test.Hspec
+import IHP.Prelude
+import IHP.HtmlSupport.QQ
+import qualified Text.Blaze.Renderer.Text as Blaze
+
+tests = do
+    describe "HSX" do
+        it "should work with static html" do
```
HugoPeters1024, line 14: are there any meaningful cases in which we don't want hsx to work but fail in a specific way?

Test/HaskellSupportSpec.hs
mpscholten: This function is used as a utility in the IHP email system. We use it to generate a plaintext email from the html email by removing all the html tags. [Like `text mail = stripTags (cs $ Blaze.renderHtml (html mail))`](https://github.com/digitallyinduced/ihp/blob/master/IHP/Mail/Types.hs#L45).

Test/HtmlSupport/QQSpec.hs
mpscholten: Yes, some cases that come to my mind:
- invalid element name (like `<my-invalid-el>`)
- invalid attribute name (like `<div invalid-attribute>`)

We cannot test these cases inside the QQ module as having code such as `[hsx|<my invalid>|]` will fail at compile time. We'd need to add ParserSpec for this that manually runs the HSX parser and then checks the result.

Do you have more than the above two cases in mind?

HugoPeters1024: I'm not sure how far you want to push the correctness, but unmatched tags comes to mind

mpscholten: Good idea. We should test these as well. This way we can ensure that the parser error messages are good.

Will add this to the PR tomorrow.

mpscholten: Added it now :)

Issue #413, opened 2020-09-21
IHP Repair Script


mpscholten:
We should build a small script that works similiar to `brew doctor`.

It should check:
- that the current directory looks like an ihp project
- that the `.ghci` exists and has the right permissions to be loaded
- that `build/ihp-lib` exists and references a valid IHP build
- that direnv is loaded correctly into the shell
- that direnv can correctly load the `.envrc`
- that the digitallyinduced cachix cache is installed
- check for old key in `.config/nix/nix.conf` (`digitallyinduced.cachix.org-1:3mGU1b6u5obFp2VUfI55Xe8/+mawl7y9Eztu3rb94PI= digitallyinduced.cachix.org-1:y+wQvrnxQ+PdEsCt91rmvv39qRCYzEgGQaldK26hCKE=`)

PR #460, opened 2020-10-11
Added first version of ihp-troubleshoot command

mpscholten:
Fixes https://github.com/digitallyinduced/ihp/issues/413

### DIFF
--- a/Troubleshoot/ihp-troubleshoot
+++ b/Troubleshoot/ihp-troubleshoot
@@ -115,6 +115,6 @@ echo "$direnv_output"
 step "which direnv Output"
 echo "$which_direnv_output"
 step "default.nix"
-echo "$(cat default.nix|head -n5)"
+head -n5 default.nix
 step "OS"
 uname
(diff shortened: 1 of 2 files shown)


HugoPeters1024:
This is a cool idea!

Troubleshoot/ihp-troubleshoot
```diff
@@ -0,0 +1,120 @@
...
+step "GHCI Output"
+echo "$ghci_output"
+step "build/ihp-lib Target"
+echo "$ihplib_link"
+step "Direnv Output"
+echo "$direnv_output"
+step "which direnv Output"
+echo "$which_direnv_output"
+step "default.nix"
+echo "$(cat default.nix|head -n5)"
```
hendi, line 118: No need for the `echo`, just do `cat default.nix|head -n5`

hendi: Also, no need for `cat`, just do `head -n5 default.nix` :wink:

mpscholten: Good catch :) Just removed it 👍

PR #462, opened 2020-10-14
guide: Address confusion with PostsController / HTTP POST

HugoPeters1024:
There was a discussion in the gitter about how the name PostsController can be confusing, especially in the context of routing. I added a small note to elaborate the difference


mpscholten (approved):
Good addition, thanks :)

PR #466, opened 2020-10-15
Re-exporting Only from postgresql-simple

fegu:
The type Only is not available in controllers. Strange as there is an example in the documentation showing use of it, but it does not actually compile. Fixed with this PR.


mpscholten:
Good point. I'd rather add it to `IHP.ControllerPrelude` to keep things clean :)

fegu:
Good idea, done.

mpscholten (approved):

PR #473, opened 2020-10-18
Add support for timestamp without time zone (i.e. localtime)

fegu:
There was some support already (seemed to be started but not finished) in the code generator. In addition to adding support for localtime this PR also solves an unrelated bug: editing a primary key (.e.g. setting unique on/off) always deselects it as a primary key. Also added a couple of clarifications to contributing doc.


mpscholten (approved):

mpscholten:
Looks good, thanks! :)

PR #477, opened 2020-10-20
Document differences between user vs admin auth

unhammer:
----

This worked for me, would love to find out if it's wrong :-)


mpscholten (approved):
Good additions, thanks :)

PR #485, opened 2020-10-21
Refactored FrameworkConfig to a datatype

HugoPeters1024:
## Summary

As discussed with @mpscholten a while ago, I propose to implement the FrameworkConfig typeclass as a datatype instead. This significantly reduces type headers in many of the framework's functions and exposes the config to the entire pipeline via either the ApplicationContext as well as the RequestContext. This means that users will still have access to their own configuration in controllers.

## TODO
- [x] Update the boilerplate to reflect the configuration initialization (https://github.com/digitallyinduced/ihp-boilerplate/pull/7)
- [x]  create a proxy getter for all properties that uses an implicit FrameworkConfig argument so we don't have to pass the instance explicitly everytime. (Perhaps FrameworkConfigProxy typeclass?)
- [x] Ensure the documentation is updated
- [x] Lift the FrameworkConfig argument of runServer into IO? This way users can do arbitrary queries to build their framework config. If we also lift defaultFrameworkConfig we are released from an unsafePerformIO.
- [x] ~~Confirm that the subset equality property of the IDE Context datatype is a valid assumption and the best way to implement this.~~ (No longer relevant)

## Considerations
- Is not backwards compatible with existing projects

Premature feedback is appreciated :)


IHP/ErrorController.hs
```diff
@@ -48,14 +47,14 @@ handleNoResponseReturned controller = do
             
         |]
     let title = [hsx|No response returned in {tshow controller}|]
-    let (RequestContext _ respond _ _ _) = ?requestContext
+    let (RequestContext _ respond _ _ _ _) = ?requestContext
```
mpscholten, line 50: Maybe when we're on this already we could replace this with a named pattern match:
```suggestion
    let RequestContext { respond } = ?requestContext
```

This it's more robust against future changes when we want to add more fields :)

IHP/Mail.hs
```diff
@@ -29,8 +31,8 @@ buildMail mail = let ?mail = mail in simpleMail (to mail) from subject (cs $ tex
 -- | Sends an email
 --
 -- Uses the mail server provided in the controller context, configured in Config/Config.hs
-sendMail :: (Config.FrameworkConfig, BuildMail mail) => mail -> IO ()
-sendMail mail = buildMail mail >>= sendWithMailServer Config.mailServer
+sendMail :: (BuildMail mail, ?requestContext :: RequestContext ) => mail -> IO ()
```
mpscholten, line 34: I'm a bit worried about this change. When using e.g. a Script (or generally anything running outside the request-response cycle, such as a worker queue) this will not be callable as the `?requestContext` is of course not available.

The only solution that came to my mind was to have some kind of stub request context for Scripts and the job queue.

mpscholten (commented):
Did a first quick overview :) Looks good in generally. Will dig deeper into this tomorrow and try running it with some real world ihp apps to see whether we're breaking something.

IHP/Mail.hs
HugoPeters1024: This is indeed one of the most blatant assumptions I made. I was developing on a fresh project so I can't foresee if this breaks anything. Would be nice to have an all out project that uses all features, that makes developing easier since things break more often.

HugoPeters1024: The example given in the documentation only shows using this function from within a controller action at which point you do have access to the RequestContext. Or did you have plans of providing a more generic use case?

IHP/IDE/Types.hs
```diff
@@ -147,8 +152,49 @@ data Context = Context
...
+    { actionVar_ :: MVar Action
+    , portConfig_ :: PortConfig
+    , appStateRef_ :: IORef AppState
+    , isDebugMode_ :: Bool
     } deriving (Eq)
 
+toEqualitySubset :: Context -> ContextEqualitySubset
+toEqualitySubset (Context var pconf appstate mode _) = ContextEqualitySubset var pconf appstate mode
+
+instance Eq Context where
```
mpscholten, line 195: Where's the need for this Eq instance coming from? :) Feels super hacky, but I guess there's a good reason for this.

IHP/Mail.hs
HugoPeters1024: I think I found a solution by requiring only a FrameworkConfig that gets injected when setting up the GHCI environment

mpscholten: Yes, we have plans to add a email queue, so that the emails would be send in the background (this makes e.g. actions that send emails way faster). For that we just need to merge our internal job queue into IHP. The job runner is build using an IHP script.

For now the `sendMailFromScript` solution looks good :)

exe/IHP/CLI/run-script
```diff
@@ -28,10 +28,11 @@ fi
 
 echo "import IHP.ScriptSupport" >> "$GHCI_SCRIPT"
 echo ":set -i." >> "$GHCI_SCRIPT"
+echo ":l Config/Config.hs" >> "$GHCI_SCRIPT"
 echo ":l Config/Config.hs Application/Script/$TASK_MODULE.hs" >> "$GHCI_SCRIPT"
 echo "import qualified Application.Script.$TASK_MODULE as Script" >> "$GHCI_SCRIPT"
-echo "IHP.ScriptSupport.runScript Script.run" >> "$GHCI_SCRIPT"
-echo ":quit" >> "$GHCI_SCRIPT"
+echo "let ?frameworkConfigIO = config in IHP.ScriptSupport.runScript Script.run" >> "$GHCI_SCRIPT"
```
mpscholten, line 34: We also need to update this make task here:

https://github.com/digitallyinduced/ihp/blob/master/lib/IHP/Makefile.dist#L179

It builds a script to a binary. You can run it via `make build/bin/Script/SomeScript` :)

IHP/FrameworkConfig.hs
```diff
@@ -18,18 +17,26 @@ import IHP.Mail.Types
 defaultPort :: Int
 defaultPort = 8000
 
-portRef :: IORef Int
-portRef = unsafePerformIO (newIORef defaultPort)
+portRef :: IO (IORef Int)
```
mpscholten, line 20: I think we could actually now get completly rid of this `portRef` by merging `initAppPort` into `defaultFrameworkConfig` :)

IHP/IDE/Types.hs
HugoPeters1024: I'm not sure where or why, but equality was previously on Contex. However with the addition of the FrameworkConfig it can't be derived (because Middleware is a function for starters). I'll dig deeper to see if Eq can be omitted.

HugoPeters1024:
> How do I integrate this into my `Config/Config.hs`?
> 

You could leverage the default values of existing method and then override what you want, something like:

```haskell
config :: IO FrameworkConfig
config = setMailServer <$> defaultFrameworkConfig "localhost" Development
  
-- not sure if this syntax works but you get the gist
setMailServer :: FrameworkConfig -> FrameworkConfig
setMailServer FrameworkConfig {..} = let mailServer = SES { accessKey = "..", secretKey = "..", region = "eu-west-1" } in FrameworkConfig {..}
```

IHP/IDE/Types.hs
HugoPeters1024: It's because the Context is part of the `ToolServerApplication` of which `frontControllerToWaiApp` requires an Eq instance. However, if I remove that constraint from `frontControllerToWaiApp` everything seems to be fine. Any reason that it is there?

exe/IHP/CLI/run-script
HugoPeters1024: It might also be a little bit deceiving that the configuration does not change for scripts that are already built. We should either add a good warning about this in the guide or think about a way to dynamically load the config.

HugoPeters1024: https://purelyfunctional.org/posts/2016-05-20-dynamic-loading-haskell-module.html It is possible, but it's not pretty :/

mpscholten: > It might also be a little bit deceiving that the configuration does not change for scripts that are already built

I would be ok with that. Scripts are usually only build to binary when running in productions (in dev you just call them using `./Application/Script/SomeScript.hs`).

fegu:
I always thought that compiling scripts to a binary is overcomplicating things. Why not just fork a thread in the main process to run these at user-configurable intervals?

HugoPeters1024:
> I always thought that compiling scripts to a binary is overcomplicating things. Why not just fork a thread in the main process to run these at user-configurable intervals?

For the purpose of a background job this makes sense, but I can also foresee some devops tasks that you might want to call manually from time to time.

mpscholten:
> For the purpose of a background job this makes sense, but I can also foresee some devops tasks that you might want to call manually from time to time.

Yes, exactly. Scripts are basically the lowest level api for defining other entry points into an IHP app.

For user-configurable intervals etc. we use the job queue. Our job queue system runner is also just a IHP script actually :)

IHP/FrameworkConfig.hs
HugoPeters1024: How do we deal with the portRef write in startToolServer in `IHP/IHP/IDE/ToolServer.hs`? Can it just be removed?

mpscholten: It does that to set the port at which the tool server runs. Usually port 8001, but when port 8001 is busy it decides to user other free ports. The tool server is also built with IHP. The `startToolServer'` is similiar to the `IHP.Server` module. Previously the only way for getting this to work with the FrameworkConfig type class was by using an IORef. Now we can just build a right FrameworkConfig value where we just set the port.

mpscholten: Best way to test that this still works is by having two different ihp projects and running them at the same time. This should then still work.

If you're curious: here's the code that find a set of ports which IHP is going to use to start all it's services: https://github.com/digitallyinduced/ihp/blob/master/IHP/IDE/PortConfig.hs#L67 :)

HugoPeters1024: It seems to me like that the value that ToolServer writes to the IORef doesn't ever get read. I just removed it and starting a second project nicely picks a different set of ports.

mpscholten:
We've added new config attributes yesterday with https://github.com/digitallyinduced/ihp/commit/1095914191ebfe5a28df8f0f5a5c5dac34ef2b93 :) Can we add them here as well? :)

HugoPeters1024:
> We've added new config attributes yesterday with [1095914](https://github.com/digitallyinduced/ihp/commit/1095914191ebfe5a28df8f0f5a5c5dac34ef2b93) :) Can we add them here as well? :)

I did a rebase last night, I remember adding those two properties :)

IHP/FrameworkConfig.hs
```diff
@@ -19,18 +18,26 @@ import IHP.Mail.Types
...
-portRef :: IORef Int
-portRef = unsafePerformIO (newIORef defaultPort)
+defaultFrameworkConfig :: Text -> Environment -> IO FrameworkConfig
+defaultFrameworkConfig appHostname environment = do
+    appPort <- defaultAppPort
+    databaseUrl <- defaultDatabaseUrl
+    let
+        baseUrl = let port = appPort in "http://" <> appHostname <> (if port /= 80 then ":" <> tshow port else "")
+        requestLoggerMiddleware = RequestLogger.logStdoutDev
+        sessionCookie = defaultIHPSessionCookie baseUrl
```
mpscholten, line 28: I think there's an issue here. The `baseUrl` is passed here to check whether the session cookie needs to be `Secure` (meaning only available for https connections). Previously with the class based approach the `baseUrl` from the `FrameworkConfig` instance was passed.

An example: Let's say I'm running an IHP app on `https://example.com`. In the old `FrameworkConfig` I override the `baseUrl` to be `https://example.com`. Then the sessionCookie will be secure because `https://example.com` is passed to the `defaultIHPSessionCookie`.


But now with this change it would pass `let port = appPort in "http://" <> appHostname <> (if port /= 80 then ":" <> tshow port else "")` to the `defaultIHPSessionCookie` instead of the overriden `https://example.com`. Then the cookie would not be secure unless I override the `sessionCookie` function as well.

mpscholten: Basically what we're missing is late binding. Will have to think a little bit more about how we can easily solve this here.

HugoPeters1024: Maybe just expand defaultFrameworkConfig into developer-land?

Guide/controller.markdown
```diff
@@ -159,6 +159,9 @@ action ExampleAction = do
 This will render `Hello World, Unnamed!` when the `ExampleAction` is called without a `firstname` parameter.
 
 
+### Accessing the FrameworkConfig inside Controllers and Views.
+The instance of the `FrameworkConfig` that is defined in `Config/Config.hs` will be part of the `RequestContext` and is thus available in controllers. To make it more convienient a  helper function called `getConfig` is available. This function takes a record from the FrameworkConfig and evaluates it. For example: `getConfig environment`.
```
mpscholten, line 163:
```suggestion
The instance of the `FrameworkConfig` that is defined in `Config/Config.hs` will be part of the `RequestContext` and is thus available in controllers. To make it more convienient a  helper function called `getConfig` is available. This function takes a record from the FrameworkConfig and evaluates it.

For example:
```haskell
action MyAction = do
    let env = getConfig environment
    when (isDevelopment env) (putStrLn "Running in dev mode")
```

Guide/mail.markdown
```diff
@@ -89,6 +89,8 @@ action MyAction = do
     sendMail ConfirmationMail { user }
 ```
 
+Note that emails can also be send from within scripts but only with the slightly modified `sendmailFromScript` function.
```
mpscholten, line 92:
```suggestion
Note that emails can also be send from within scripts but only with the slightly modified `sendMailFromScript` function.
```

mpscholten (approved):
Looks good to me. Small improvements to the docu :)

PR #486, opened 2020-10-22
IDE support for array types

fegu:
Postgres Array types have been supported in Schema.sql and codegen for a few weeks now, but there hasn't been any support in the IDE. I also lumped in a bugfix for the duplicate null/NULL occasionally showing in the default dropdown.

### DIFF
--- a/IHP/IDE/Data/Controller.hs
+++ b/IHP/IDE/Data/Controller.hs
@@ -29,21 +29,16 @@ instance Controller DataController where
         connection <- connectToAppDb
         tableNames <- fetchTableNames connection
         primaryKeyFields <- tablePrimaryKeyFields connection tableName
-
         rows :: [[DynamicField]] <- fetchRows connection tableName
-
         tableCols <- fetchTableCols connection tableName
-
         PG.close connection
         render ShowTableRowsView { .. }
 
     action ShowQueryAction = do
         connection <- connectToAppDb
         let query = (param @Text "query")
-        when (query == "") do
-            redirectTo ShowDatabaseAction
-        rows :: [[DynamicField]] <- PG.query_ connection (fromString (cs query))
-
+        when (query == "") $ redirectTo ShowDatabaseAction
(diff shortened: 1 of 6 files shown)


IHP/IDE/SchemaDesigner/Controller/Columns.hs
```diff
@@ -228,3 +228,9 @@ isCreateEnumType CreateEnumType {} = True
 isCreateEnumType _ = False
 
 nameList statements = map (get #name) statements
+
+arrayifytype :: Bool -> PostgresType -> PostgresType
+arrayifytype False   (PArray coltype) = coltype
+arrayifytype True  a@(PArray coltype) = a
```
HugoPeters1024, line 234: Is this an expected case? I feel like this should never happen. (if so, an error of sorts might be better)

HugoPeters1024 (commented):
Nice!

IHP/IDE/SchemaDesigner/Controller/Columns.hs
fegu: Yes. This is when editing. It could be an array type or not before edit, and the checkbox could turn it on or off. So all cases are viable.

fegu:
Added a quick fix for something that has been bugging me: the data view custom query input will crash on any DDL, INSERT; UPDATE (i.e. stuff not returning rows). Now it works (but should be made nicer such as returning affected rows etc).

IHP/IDE/Data/Controller.hs
```diff
@@ -195,3 +190,6 @@ parseValues False True text = text
 parseValues True True text = text
 
 updateValues list = map (\elem -> fst elem <> " = " <> snd elem) list
+
+isQuery sql = T.isInfixOf "SELECT" u
```
mpscholten, line 194: Finally fixed haha, good job :)

mpscholten:
Looks good to me. @nezbite can you also take a quick look?

nezbite (approved):

mpscholten (approved):

Issue #365, opened 2020-08-31
Custom css framework


themmes:
Hey!

In my tiny test project I was trying to make IHP work with tailwindcss (https://github.com/themmes/ihp-todo). I thought I'd simply go the easy route first, so added the CDN to the `Layout.hs` and started styling the `hsx`. This works well until I want to style `flashMessages` or form related items. The documentation is still `WIP` on this, so I thought I might ask you folks for advice and then attempt to write it.

The docs 
point me to: 
1. https://ihp.digitallyinduced.com/Guide/form.html#working-with-other-css-frameworks
2. https://ihp.digitallyinduced.com/api-docs/IHP-View-Form.html#v:horizontalFormFor
3. https://ihp.digitallyinduced.com/api-docs/src/IHP.View.Form.html#renderHorizontalBootstrapFormField

Brings up the following questions:

- What I dont really understand is why there exist custom functions specific to Bootstrap? 
- Why bootstrap classes on e.g. the button are default, why can't I just customise from scratch?
- Is there a way to add children elements? e.g. an SVG icon to the button
- Maybe the CSS section should get a more appropriate location in the docs, instead of being under `Forms`?

Would love to hear your thoughts!

themmes:
To clarify, this relates to https://github.com/digitallyinduced/ihp/issues/72 but isn't necessarily limited to forms.

mpscholten:
Hey Tom,

thanks for opening an issue on this 👍 


> What I dont really understand is why there exist custom functions specific to Bootstrap?

IHP is optimized to be used with bootstrap because that was what we needed at di for building our apps. We've basically optimized it for ourselves.

> Why bootstrap classes on e.g. the button are default, why can't I just customise from scratch?

I think we should make it possible to completely customize the default classes used. It would be cool to later have something like a `ihp-tailwindcss` package which one could use to have everything default to e.g. `tailwindcss` instead of bootstrap.

> Is there a way to add children elements? e.g. an SVG icon to the button

Currently not supported, we should add that 👍 

> Maybe the CSS section should get a more appropriate location in the docs, instead of being under Forms?

Agree, we could add a new `Styling` section below the `Frontend` headline

![image](https://user-images.githubusercontent.com/2072185/91978114-e764c280-ed23-11ea-8dca-20e4df87b8a3.png)

themmes:
Thanks for your response @mpscholten, after my holidays (2 weeks) I'll pick this up!

mpscholten:
[...]

PR #487, opened 2020-10-22
Introduced CSSFramework

mpscholten:
This allows us to add support for more css frameworks beyond bootstrap. E.g. tailwind.

First draft for https://github.com/digitallyinduced/ihp/issues/365

To use this branch inside an app, a `cssFramework` field needs to be added to the `ViewContext`.

In `Web/Types.hs` add:

```haskell
data ViewContext = ViewContext
    { 
    -- .......
    , cssFramework :: CSSFramework
    }
```

In `Web.View.Context` add:

```haskell
import IHP.View.CSSFramework

-- ....
        let viewContext = ViewContext {
                -.....
                cssFramework = bootstrap
            }
```

Once https://github.com/digitallyinduced/ihp/pull/485 is merged we can add this to the FrameworkConfig. Then the css framework could be changed inside `Config.hs` like this:

```haskell
instance FrameworkConfig where 
    environment = Development
    appHostname = "localhost"
    cssFramework = tailwind
```

We can also customize the default frameworks:

```haskell
instance FrameworkConfig where 
    environment = Development
    appHostname = "localhost"
    cssFramework = tailwind |> set #styledSubmitButtonClass "my-app-button"
```

----
 

Todos:
- [x] Add more tests for bootstrap
- [x] Make sure that select works
- [x] Make sure that checkboxes work
- [ ] Add tests for tailwind
- [ ] Make sure that the tailwind is actually correct (I have no experience with tailwind, so maybe someone else can confirm the css classes are correct)
- [x] Integrate this into the Config.hs


kodeFant:
Just a general note, I don't feel qualified to review the code yet at my level :)

Bundling the whole tailwind css can be a really large file, a couple of MB I think. Will this support the processing of tailwind so that only the utilized classes are being built into the production output?

I have made a fairly simple gist of a tailwind script using the Tailwind CLI, one build to dev to support the dev environment with all the tailwind classes, and one that builds the output in an optimized build. Just running it locally as a bash script whenever I need it.

The `tailwind.config.js` has this purge property that tells Tailwind CLI to search through the relevant files for classnames and only include those that are mentioned. 

Not sure if it's usable for this application, but feel free to take from if it's of any use:
https://gist.github.com/kodeFant/57e6fb8917e834feb052d63affc132cb

mpscholten:
Thanks for sharing the script. I think this can be useful later on. This PR mainly adds the ability to use the tailwind classes with `formFor` and flash messages. Actually adding the tailwind.css is something I'd do in a separate PR later on. Then we can wire up your script to do the bundling 👍

PR #492, opened 2020-10-22
troubleshoot script: use absolute path when verifying link

HugoPeters1024:
When developing on IHP itself the link will be a relative path starting with `../`, making the script think it's broken. Using the `-f` flag on readlink we get an absolute path which works in all cases.


mpscholten (approved):
Thanks :)

PR #497, opened 2020-10-23
Database refactoring docs

fegu:
Expanded on the current docs to help people move their projects forward through database changes. Chose steps that points contributors to how we should automate this in the IDE, using DDL which avoids manually updating Fixtures.sql, is a lot less involved, and now can be done without errors in the IDE. The "old way" is still there as an alternate route.


mpscholten (approved):
Thanks :) Very good additions 👍

PR #499, opened 2020-10-24
Fix shellcheck warnings

unhammer:
`BIN_DIR/LIB_DIR` were not exported nor used, so deleted.

----

Nothing serious, but it's always a good idea to avoid warning-blindness :)


mpscholten (approved):
Thanks 👍

PR #500, opened 2020-10-24
Set right .ghci permissions for user in ihp-new

unhammer:
(But keep going if chmod fails for whatever reason)


mpscholten (approved):
👍 also a very good change :) Thanks

Issue #96, opened 2020-06-08
Basic Auth


mpscholten:
For very simple apps and MVPs it would be great to have a `basicAuth` function which can be used like this:

```haskell
instance Controller UsersController where
    beforeAction = basicAuth "beta" "hunter2"
```

https://app.clubhouse.io/digitallyinduced/story/1440/basic-auth

fegu:
Implemented in pull request #504

PR #504, opened 2020-10-25
basicAuth support

fegu:
Adds http basic authentication support (including docs). Also removes some duplicate entries in the cabal file.
Solves issue #96

### DIFF
--- a/IHP/Controller/Session.hs
+++ b/IHP/Controller/Session.hs
@@ -7,12 +7,15 @@ import qualified Data.Text.Read
 import qualified Data.UUID
 import           IHP.Controller.RequestContext
 import           IHP.HaskellSupport
-import           Network.HTTP.Types                   (status200, status302)
+import           IHP.ControllerSupport
+import           Network.HTTP.Types                   (status200, status302, status401)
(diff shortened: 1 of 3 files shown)


IHP/Controller/Session.hs
```diff
@@ -53,6 +56,12 @@ getSessionUUID name = do
             Just (Just value) -> Just value
             _                 -> Nothing
 
+basicAuth :: (?requestContext::RequestContext) => Text -> Text -> Text -> IO ()
```
mpscholten, line 59: Can we add some haddock documentaton `| -- ...` to this function? This is then visible when using e.g. haskell language server.

mpscholten (commented):
Thanks, looks good 👍

IHP/Controller/Session.hs
fegu: Added now.

fegu: ..and that hunter2 reference still makes me chuckle :)

mpscholten: finally someone got it haha :D

mpscholten (approved):
Thanks :)

PR #508, opened 2020-10-26
support for custom text on js-delete

fegu:
Adds support for data-confirmtext="Your custom are you sure text here" on any link having class js-delete.

The standard text bugs me for two reasons:
1) too generic (warning should confirm exactly what is about to be deleted)
2) language is english. While users will tolerate english error messages, messages in ordinary usage should be localized.

If the approach taken here is accepted (add a comment), I will also update the docs with this feature in a separate commit (same PR)

### DIFF
--- a/lib/IHP/static/helpers.js
+++ b/lib/IHP/static/helpers.js
@@ -50,7 +50,7 @@ function initDelete() {
 
             if (!event.target.classList.contains('js-delete-no-confirm')) {
                 var confirmText = event.target.dataset.confirm || 'Are you sure you want to delete this?';
-                if (!confirm(txt ? txt : 'Are you sure you want to delete this?')) {
+                if (!confirm(confirmText)) {
                     return;
                 }
             }
--- a/lib/IHP/static/helpers.js
+++ b/lib/IHP/static/helpers.js
@@ -49,7 +49,8 @@ function initDelete() {
             event.preventDefault();
 
             if (!event.target.classList.contains('js-delete-no-confirm')) {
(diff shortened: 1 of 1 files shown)


lib/IHP/static/helpers.js
```diff
@@ -49,7 +49,8 @@ function initDelete() {
             event.preventDefault();
 
             if (!event.target.classList.contains('js-delete-no-confirm')) {
-                if (!confirm('Are you sure you want to delete this?')) {
+                let txt = event.target.dataset.confirmtext;
```
mpscholten, line 52: Can we use `var` here instead. I'd also shorten the name to just `data-confirm` instead of `data-confirmtext`.

```suggestion
                var confirmText = event.target.dataset.confirm || 'Are you sure you want to delete this?';
```

lib/IHP/static/helpers.js
```diff
@@ -49,7 +49,8 @@ function initDelete() {
             event.preventDefault();
 
             if (!event.target.classList.contains('js-delete-no-confirm')) {
-                if (!confirm('Are you sure you want to delete this?')) {
+                var confirmText = event.target.dataset.confirm || 'Are you sure you want to delete this?';
+                if (!confirm(txt ? txt : 'Are you sure you want to delete this?')) {
```
mpscholten, line 53:
```suggestion
                if (!confirm(confirmText)) {
```

mpscholten (commented):
We also need to change this one then :)

PR #515, opened 2020-10-28
Move wiki pages to guide

### LINKED ISSUE #513: Move the IHP Wiki into the Guide
Currently the IHP Wiki can only be updated by people with write access to the repo. It's not possible to make a PR to the Wiki. I think it's also not very clear that we use the wiki for some things (it's not really visible anywhere).

To fix this we should move all the pages from the wiki into it's own markdown files in the IHP Guide.

The markdown files need to be placed here: https://github.com/digitallyinduced/ihp/tree/master/Guide Also needs to be added to the Makefile: https://github.com/digitallyinduced/ihp/blob/master/Guide/Makefile

bulters:
This moves the two wiki pages into the guide, allowing external
contributions to these by PRs.

This fixes #513

First PR

### DIFF
--- a/Guide/Makefile
+++ b/Guide/Makefile
@@ -27,6 +27,8 @@ HTML_FILES+= editors.html
 HTML_FILES+= deployment.html
 HTML_FILES+= auto-refresh.html
 HTML_FILES+= mail.html
+HTML_FILES+= design-goals.html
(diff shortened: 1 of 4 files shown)


mpscholten (approved):
Perfect thanks for the fast PR :)

PR #518, opened 2020-10-30
Always use ?context with ConfigProvider typeclass

HugoPeters1024:
This is the continuation of #485.

### Changes
- `FrameworkConfig` is now a datatype
- `FrameworkConfig` is now build by a state monad, removing unsafe operations and resolving inter-dependency between fields
- The `ConfigProvider` typeclass has been created, which is a forces a type to be able to produce an instance of `FrameworkConfig`
- All instances of `?requestContext` and `?viewContext` have been replaced by `?context`, this way generic helpers can capture either of those under `ConfigProvider` typeclass.

- UPGRADE.md is introduced to track breaking changes between releases
- UPGRADE.md is populated with fixes to these changes.


HugoPeters1024:
Replaced by #519

PR #524, opened 2020-11-01
remove redudant ConfigProvider instance in codegen and small guide fix

HugoPeters1024:
These are two small cleanups of artifacts from refactoring the FrameworkConfig.


mpscholten (approved):
Good catch, thanks :)

PR #531, opened 2020-11-04
Actually use user-specified labelClass on checkboxField

unhammer:
(Or, if this shouldn't be possible, perhaps the doc's should mention that where checkboxField is mentioned.)


mpscholten (approved):
Good point, thanks 👍

PR #536, opened 2020-11-05
Doc how to (not) preload links

unhammer:
This is how I read the code anyway.

Btw, is there any reason `data-turbolinks-preload` isn't in use? turbolinksInstantClick.js has

```
const preloadAttribute = function(link) {
return true;
…
}
```

It'd be nice if we could have preloading unless that attribute is explicitly false, but perhaps there's a good reason for the early `return true` there?

### DIFF
--- a/Guide/view.markdown
+++ b/Guide/view.markdown
@@ -160,6 +160,14 @@ You can disable this behavior by removing the following code from your `Web/Layo
         |]
 ```
 
+Preloading with InstantClick on hover will only happen with links that
+
+1. Use the GET method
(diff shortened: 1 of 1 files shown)


mpscholten:
Thanks 👍 

> It'd be nice if we could have preloading unless that attribute is explicitly false, but perhaps there's a good reason for the early return true there?

Cannot remember why we put it there. But looking at the indentation of that line of code it looks like it was just put there for quick testing. The git history is also not very helpful for that file.

Maybe you can drop the early return and make a second PR for that? :)

PR #544, opened 2020-11-11
Respect `data-turbolinks-preload='false'` attribute

unhammer:
(But still preload if that attribute does not exist at all!)

See #536. Just removing the `return true` would turn off preloading if
the attribute wasn't there at all – pretty sure we don't want
that. Now it should only turn off preloading if it's explicitly there
and false.


mpscholten (approved):
Thanks :)

PR #545, opened 2020-11-11
doc: Add example of how to have a certain select option preselected

unhammer:


Guide/form.markdown
```diff
@@ -395,6 +395,15 @@ Given the above example, the rendered form will look like this:
 </form>
 ```
 
+If you want a certain value to be preselected, change the `fieldValue` of the returned `selectField`. For example, to have Andreas preselected in the above example:
+
+```haskell
+formFor project [hsx|
```
mpscholten, line 401: Actually the suggested way is to set this on the record for which the form is generated, like this:

```
action MyAction = do
    users <- query @User |> fetch
    let project = newRecord @User |> set #userId (users !! 1)
    render NewProjectView { .. }
```

The `fieldValue` is already converted to a HTML string, so you loose type safety by overriding it that way.

unhammer: Oh, that makes things a lot simpler! Thanks, updated the PR – how does that look?

mpscholten (approved):
Perfect, thanks :)

PR #546, opened 2020-11-13
Signed Int readParam bugfix, added JSONB column in IDE and two other minor fixes

fegu:
- FIX: numberField with negative number gives `ParamReader Int: Failed reading: takeWhile1` because `readParam Int` does not allow minus sign.
- ADD: JSONB column was missing from IDE (but otherwise supported).
- FIX: Timezone without timezone field did not have NOW() available as default in IDE
- FIX: Spawning browser on `./start`does not work in windows due to missing package in install guide.
- Added hint of nix gc in install docs. Saves people many GB.

### DIFF
--- a/IHP/Controller/Param.hs
+++ b/IHP/Controller/Param.hs
@@ -276,7 +276,7 @@ instance ParamReader Integer where
     readParameter byteString =
         case Attoparsec.parseOnly ((Attoparsec.signed Attoparsec.decimal) <* Attoparsec.endOfInput) byteString of
             Right value -> Right value
-            Left error -> Left ("ParamReader Int: " <> cs error)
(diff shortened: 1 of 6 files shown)


IHP/Controller/Param.hs
```diff
@@ -261,7 +261,7 @@ instance ParamReader ByteString where
 instance ParamReader Int where
     {-# INLINE readParameter #-}
     readParameter byteString =
```
mpscholten, line 263: Could add a short testcase to https://github.com/digitallyinduced/ihp/blob/master/Test/Controller/ParamSpec.hs#L26 :)

fegu: will do

mpscholten (approved):
Thanks :)

PR #547, opened 2020-11-13
Binary equals to BYTEA in postgres (UI)

nezbite:


fegu:
Should add `BYTEA` in isIllegalKeyword in `View\Layout.hs`

mpscholten (approved):

PR #569, opened 2020-11-23
Add mocking framework for controllers

nghamilton:
Is this useful to incorporate into IHP framework? Is this the right spot for it?

Example usage:
```
import Test.Controller.Mocking

spec :: Spec
spec = beforeAll (configBuilder >>= mockContext >>= \c -> withContext c (pure ()) do
  describe "Run mock test" do

    it "checks headers" $ c -> withParams [("param","val"] c do
      heads <- headers (mockAction SomeAction)
      let loc = lookup "Location" heads
      loc `shouldNotBe` Nothing

    it "checks status" $ c -> withParams [("param","val"] c do
      mockActionStatus AnotherAction `shouldReturn` status200
```

### DIFF
--- a/Guide/Makefile
+++ b/Guide/Makefile
@@ -9,6 +9,7 @@ HTML_FILES+= installation.html
 HTML_FILES+= index.html
 HTML_FILES+= database.html
 HTML_FILES+= debugging.html
+HTML_FILES+= testing.html
 HTML_FILES+= querybuilder.html
 HTML_FILES+= view.html
 HTML_FILES+= form.html
--- a/Guide/layout.html
+++ b/Guide/layout.html
@@ -48,6 +48,7 @@
                 <a class="nav-link secondary" href="javascript:alert('This cookie section is not ready yet')">Cookies</a>
                 <a class="nav-link secondary" href="scripts.html">Scripts</a>
                 <a class="nav-link secondary" href="debugging.html">Debugging</a>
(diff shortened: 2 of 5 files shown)


mpscholten:
Yes, let's add this to IHP :) If we want this to be callable from userland we should propably name the module something like `IHP.Test.Controller`. We should also add some documentation on this to the Guide.

nghamilton:
Cool, I can do that.

Is there a neat way to get rid of the userland dependencies? `WebApplication`, `User`, and the `RootApplication` instance of `FrontController`? I reckon they need to be taken out.

mpscholten:
The `RootApplication` can stay as it's defined in the `IHP.FrameworkConfig` module. We could pass the `WebApplication` as an argument to `withContext`. This would also help to solve the `User` issue, as we could now just call the `initContext` of the `InitControllerContext WebApplication` instance (like https://github.com/digitallyinduced/ihp-blog-example-app/blob/master/Web/FrontController.hs#L20). That `initContext` would then set the user to `Nothing`. If a specific test case needs a custom user, it could always call `putContext` to override the `Nothing` value with a mock user.

nghamilton:
I have tidied up the mocking context, if we are happy with it I'll write up some docs.

mpscholten:
Looks good to me 👍

nghamilton:
Updated and added docs.

mpscholten (approved):
Looks good, thanks 👍

PR #573, opened 2020-11-26
Added guide on how to use tailwind

mpscholten:
You can find my first tailwind app here: https://veghhvypzdbnfpjgomywejhyzagkqmuo.ihpapp.com/NewPost Yes it looks very raw :D 

@kodeFant can you do a review? I've taken some of the code from your gist :)

### DIFF
--- a/Guide/Makefile
+++ b/Guide/Makefile
@@ -31,6 +31,7 @@ HTML_FILES+= design-goals.html
 HTML_FILES+= troubleshooting.html
 HTML_FILES+= modal.html
 HTML_FILES+= database-migrations.html
+HTML_FILES+= tailwindcss.html
 
 all: $(HTML_FILES) bootstrap.css instantclick.js
 
--- a/Guide/layout.html
+++ b/Guide/layout.html
@@ -62,7 +62,8 @@
                 <a class="nav-link secondary" href="design-goals.html">IHP Design Goals</a>
 
 
-                <a class="nav-link headline" href="javascript:alert('The frontend section is coming soon')">Frontend</a>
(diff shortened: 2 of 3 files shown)


kodeFant:
I think this looks really decent 😊

If you would prefer not to have a package.json file in the project, you could run all scripts with `npx` without a package.json as it will do whatever it takes to run the tailwindcli even if not installed.

On the other side, package.json is nice for getting to control the version, and probably a better practice in the long run

Will the `autoprefixer` and `postcss` actually be run somewhere under the hood? AFAIK the tailwindcss script does not run them automatically. If they are redundant as of now, maybe these docs can be clarifying on using preprocessors:
https://tailwindcss.com/docs/using-with-preprocessors

kodeFant:
By looking at the tailwind source code I think I stand corrected. When using the package.json approach, it looks like autoprefixer and postcss are tailwind deps. I have just not noticed since I have been running it directly from npx with no package.json so far :)

mpscholten:
Thanks for taking a look :)

> If you would prefer not to have a package.json file in the project, you could run all scripts with npx without a package.json as it will do whatever it takes to run the tailwindcli even if not installed.

Yes, I tried that initially as well (followed the commands from your gist basically :) ). But that failed when deploying it to IHP Cloud because somehow tailwind needs to have autoprefixer installed. With the package.json it should also be more reliable.


> Conclusion and afterthought: I think the setup you have outlined looks good, I will switch to this as I now also have a good recipe for upgrading to the brand new Tailwind 2 🥇

Awesome :) Then I'll merge it 👍

PR #574, opened 2020-11-29
Update tailwindcss.markdown

OhCoder:
whenver -> whenever


mpscholten (approved):
Good catch :) Thanks!

PR #576, opened 2020-11-29
Add a hint that one may need to click Update Db

stianlagstad:
Add a hint that one may need to click Update Db in order to see the insert-statement.


mpscholten (approved):
Perfect, thanks! :)

PR #579, opened 2020-12-01
Fix Path To Welcome.hs

s0kil:
Tiny Fix (:


mpscholten (approved):
Good catch, thanks! :)

PR #581, opened 2020-12-02
Bug fix + different message suggestion for column name

adithyaov:
This PR has 2 commits,

Commit 1:
Bugfix: Consider Haskell keywords as illegal keywords as well.

Commit 2:
Change suggested column name from `*reserved keyword_` to `*singular table name*_*reserved keyword*`

### DIFF
--- a/lib/IHP/static/ihp-schemadesigner.js
+++ b/lib/IHP/static/ihp-schemadesigner.js
@@ -8,7 +8,11 @@ function initSchemaDesigner() {
         if (reservedKeywords.includes(this.value)) {
             this.classList = "form-control is-invalid"
             var warning = document.createElement("div");
-            warning.innerHTML = "'" + this.value + "' is a reserved keyword in Haskell. Your element will be named '" + this.value + "_'.";
+            warning.innerHTML =
+		"'" + this.value + "' is a reserved keyword in Haskell. "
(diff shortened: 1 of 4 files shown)


mpscholten:
@nezbite will do the final review as he wrote most of that code here. For me this looks fine already.

lib/IHP/static/ihp-schemadesigner.js
```diff
@@ -1,3 +1,11 @@
+function toSingular(word) {
```
nezbite, line 1: When using a reserved keyword, the `is-invalid` class is added to the text field, but displaying the error message fails while using `toSingular(word)`:
```
Uncaught TypeError: word is undefined
```
In general, we use `singularize` from Text.Countable. Using this as the data attribute directly would allow to remove the JS singularize function.

nezbite (changes requested):

lib/IHP/static/ihp-schemadesigner.js
adithyaov: Avoiding JS is a good idea in general.

nezbite: Avoiding JS is everything we stand for. I just tested everything and it seems fine; Good job!

nezbite (approved):

PR #583, opened 2020-12-02
 Fix .js-delete Behavior

s0kil:
Continuation of pr https://github.com/digitallyinduced/ihp/pull/364 fixing issue https://github.com/digitallyinduced/ihp/issues/361

- add event listener only to elements with `.js-delete` class
- change `target` to `currentTarget`

Also fixes the [issue](https://github.com/digitallyinduced/ihp/pull/364#issuecomment-683948383) original author had
> There might still be something weird, if you cancel the confirmation prompt it asks you a second time. Any clue why that happens?


lib/IHP/static/helpers.js
```diff
@@ -42,33 +43,54 @@ function initTime() {
 }
 
 function initDelete() {
-    document.addEventListener('click', handleClick);
+    document.querySelectorAll('.js-delete').forEach(function (elem) {
+        if (elem.hasAttribute('jsDeleteInitialized') === false) {
```
mpscholten, line 47: image

The attribute is visible from the browser dev tools. Can't we also do `elem.jsDeleteInitialized = true`? Should work similiar to the attribute but is invisible. In case that doesn't work we should rename the attribute to something like `data-js-delete-initialized` to be more consistent with HTML attribute conventions

s0kil: `elem.jsDeleteInitialized = true` worked like a charm.

mpscholten (approved):
Perfect, thanks :)

PR #591, opened 2020-12-07
More Guide Updates

s0kil:


s0kil:
I want to also fix the `viewContext` references in Guide:
- https://ihp.digitallyinduced.com/Guide/recipes.html#making-a-dynamic-loginlogout-button
- https://ihp.digitallyinduced.com/Guide/session.html#rendering-a-message
however, I'm not exactly sure how to translate to the new API.

From what I understand, in the old API `get #user viewContext`, we are getting the `user` field from the `viewContext` record.
What does this mean: `fromFrozenContext @(Maybe User)`, I get the `Maybe User` part, but what about `@()`? And Also Why Are we passing in the `User` type, instead of a field as previously `#user`?

mpscholten:
The new controller context is key value map where the keys are types and the values are values of the type. E.g. think of a json like `{ "Maybe User": { id: .., name: .. }, .. }`. Previously the old view context was a key value map where the key was basically a data structure field. Therefore when calling `fromFrozenContext` we need to pass a type instead of a value to the function. [The `@` syntax is called type application](https://stackoverflow.com/questions/40275080/how-do-you-use-typeapplications-in-haskell).

The `@` syntax from `fromFrozenContext @(Maybe User)` is just a syntax sugar for `let maybeUser :: Maybe User = fromFrozenContext`.

mpscholten:
Any updates on this PR?

s0kil:
@mpscholten I'm also looking at https://ihp.digitallyinduced.com/Guide/session.html#rendering-a-message
And trying to translate `viewContext |> #flashMessages` into the new API, So far I have:
```haskell
setSuccessMessage "Testing Context"

flashMessages :: [FlashMessage] <- fromContext
let flashMessagesText =
      map flashMessageText flashMessages
      where
        flashMessageText fm =
          case fm of
            SuccessFlashMessage message ->
              "Success: " ++ message
            ErrorFlashMessage message ->
              "Got an Error: " ++ message

putStrLn $ tshow $ flashMessagesText
```

Is this the way to work with the context, or is there a better, maybe cleaner way?

mpscholten:
```haskell
flashMessages :: [FlashMessage] <- fromContext
let flashMessagesText =
      map flashMessageText flashMessages
      where
        flashMessageText fm =
          case fm of
            SuccessFlashMessage message ->
              "Success: " ++ message
            ErrorFlashMessage message ->
              "Got an Error: " ++ message

putStrLn $ tshow $ flashMessagesText
```
This will work when called from inside an action.

When calling this from the view, it needs to be like this:
```haskell
let flashMessages :: [FlashMessage] = fromFrozenContext
```

mpscholten (approved):
Thanks :)

PR #599, opened 2020-12-09
adding haskell units definitions for extending type safety to physica…

Montmorency:
Adding the units and units-defs packages for type safe physical dimensions and polymorphic units support. This bring in some dependencies which inflate build times.


mpscholten (approved):
Perfect thx :)

Issue #608, opened 2020-12-18
IHP Telemetry should detect WSL as an OS


mpscholten:

s0kil:
https://github.com/microsoft/WSL/issues/423#issuecomment-221627364

mpscholten:
Looks good, in case someone already wants to tackle this, we need to place the detection logic around here https://github.com/digitallyinduced/ihp/blob/master/IHP/Telemetry.hs#L40

PR #609, opened 2020-12-19
Made user record updateable in beforeLogin

fegu:
Very convenient. Also updated docs.

### DIFF
--- a/Guide/authentication.markdown
+++ b/Guide/authentication.markdown
@@ -260,7 +260,7 @@ You can also access the user using `currentUser` inside your views:
 |]
 ```
 
-## Performing actions after login
+## Performing actions on login
 
 The sessioncontroller has a convenient `beforeLogin` which is run on login after the user is authenticated, but before the target page is rendered. This can be useful for updating last login time, number of logins, aborting the login when the user is blocked etc. Add code for it in your `Web/Controller/Sessions.hs`
 
--- a/Guide/authentication.markdown
+++ b/Guide/authentication.markdown
@@ -262,7 +262,7 @@ You can also access the user using `currentUser` inside your views:
 
 ## Performing actions after login
 
-The sessioncontroller has a convenient beforeLogin which is run on login after the user is authenticated, but before the target page is rendered. This can be useful for updating last login time, number of logins etc. Add code for it in your `Web/Controller/Sessions.hs`
+The sessioncontroller has a convenient `beforeLogin` which is run on login after the user is authenticated, but before the target page is rendered. This can be useful for updating last login time, number of logins, aborting the login when the user is blocked etc. Add code for it in your `Web/Controller/Sessions.hs`
 
 ```haskell
 instance Sessions.SessionsControllerConfig User where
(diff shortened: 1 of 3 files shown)


fegu:
Added support for identifying running under WSL (as discussed)

fegu:
closes issue #608

Guide/authentication.markdown
```diff
@@ -259,6 +260,22 @@ You can also access the user using `currentUser` inside your views:
...
+
+```haskell
+instance Sessions.SessionsControllerConfig User where
+    beforeLogin = updateLoginHistory
+
+updateLoginHistory user = do
+    let n = get #logins user
+    user
+        |> set #logins (n+1)
+        |> updateRecord
```
mpscholten, line 275:
```suggestion
    user
        |> modify #logins (\count -> count + 1)
        |> updateRecord
```

Guide/authentication.markdown
```diff
@@ -259,6 +260,22 @@ You can also access the user using `currentUser` inside your views:
 |]
 ```
 
+## Performing actions after login
```
mpscholten, line 263: The function is called `beforeLogin` and the headline says `after login`, this might be a bit confusing :) Maybe we can change the headline to this:

```suggestion
## Performing actions on login
```

Guide/authentication.markdown
```diff
@@ -259,6 +260,22 @@ You can also access the user using `currentUser` inside your views:
...
+
+The sessioncontroller has a convenient beforeLogin which is run on login after the user is authenticated, but before the target page is rendered. This can be useful for updating last login time, number of logins etc. Add code for it in your `Web/Controller/Sessions.hs`
```
mpscholten, line 265:
```suggestion
The sessioncontroller has a convenient `beforeLogin` which is run on login after the user is authenticated, but before the target page is rendered. This can be useful for updating last login time, number of logins, aborting the login when the user is blocked etc. Add code for it in your `Web/Controller/Sessions.hs`
```

Guide/authentication.markdown
```diff
@@ -259,6 +260,22 @@ You can also access the user using `currentUser` inside your views:
...
+    beforeLogin = updateLoginHistory
+
+updateLoginHistory user = do
+    let n = get #logins user
+    user
+        |> set #logins (n+1)
+        |> updateRecord
+    pure ()
+```
+
```
mpscholten, line 278: Maybe we could also add a second example for the `Disallow login until user is confirmed` case here. We already have that inside the API docs, so basically it's just copy paste :)

https://ihp.digitallyinduced.com/api-docs/IHP-AuthSupport-Controller-Sessions.html#v:beforeLogin

mpscholten (commented):
Thanks! Looks good, just a few small points :)

Guide/authentication.markdown
fegu: Yes! I looked for that but did not find it. I remembered you had an example with blocking login until user has confirmed his email. Will add that later today.

mpscholten (approved):
Perfect, thanks :)

PR #621, opened 2020-12-22
Fixed typo in JobGenerator error message.

### LINKED ISSUE #620: Typo in JobGenerator error message
In `IHP/IDE/CodeGen/JobGenerator.hs`, the error message in the [case of an empty job name](https://github.com/digitallyinduced/ihp/blob/c44cfa56c7e1826c42859d20914cc8655b1f0412/IHP/IDE/CodeGen/JobGenerator.hs#L27) is

> Job name can be empty

This should clearly be

> Job name cannot be empty

instead.

zacwood9:
See title. Closes #620.


mpscholten (approved):
Good catch 👍 Thanks :)

PR #628, opened 2020-12-26
adds safe tail and init to prelude. Closes #618

### LINKED ISSUE #618: `tail` Does not return a Maybe
`head :: [a] -> Maybe a` returns a `Maybe a` type, while `tail :: [a] -> [a]` throws an runtime error if the list is empty, when it should also return a `Maybe [a]`.

fegu:


fegu:
Need to fix tests first

PR #635, opened 2020-12-27
Remove code already present in schemaDesignerLayout

sebastian-palma:
When some error is raised after reading the schema file, the header (toggle and update db buttons) are duplicated,  because that HTML is in the [`schemaDesignerLayout`](https://github.com/digitallyinduced/ihp/blob/master/IHP/IDE/SchemaDesigner/View/Layout.hs#L12:L29) function, resulting in something like this:

Screen Shot 2020-12-27 at 19 34 37

I made the less changes possible to keep the `bg-white visual-error` div inside a `container`.


mpscholten (approved):
Good catch, thanks :)

Issue #642, opened 2021-01-01
Syntax error when creating an object with text field containing quotes


k-bx:
I'm trying to add a record putting this text into a field's value:

```
In 1973, a young filmmaker named George Lucas scribbled some notes for a far-fetched space-fantasy epic. Some forty years and 37 billion later, Star Wars -- related products outnumber human beings, a growing stormtrooper army spans the globe, and "Jediism" has become a religion in its own right. Lucas's creation has grown into far more than a cinematic classic; it is, quite simply, one of the most lucrative, influential, and interactive franchises of all time. Yet incredibly, until now the complete history of Star Wars -- its influences and impact, the controversies it has spawned, its financial growth and long-term prospects -- has never been told.
```

Getting an error:

```
SqlError {sqlState = "42601", sqlExecStatus = FatalError, sqlErrorMsg = "syntax error at or near \"s\"", sqlErrorDetail = "", sqlErrorHint = ""}
```

Looks like some kind of escaping problem. Would be great if IHP would print the full SQL upon error if possible.

mpscholten:
Thanks for reporting! :) Do you do this via the Database Editor in the dev tools?

k-bx:
Ok, this might be a different one (I'll try to reproduce the one with `near \"s\"` later). The error happens when you have a schema with nullable UUID field, and you try to edit a row, IHP will by default show you a text field input with an empty string as the value.

What should happen (ideally): you shouldn't be allowed to input an empty string in that field, asked to provide a valid UUID value.

Alternatively: IHP should put NULL into that field if it's nullable.

What happens now: `SqlError {sqlState = "22P02", sqlExecStatus = FatalError, sqlErrorMsg = "invalid input syntax for type uuid: \"\"", sqlErrorDetail = "", sqlErrorHint = ""}`

k-bx:
Yeah, seems like there's a potential for an SQL injection. When you're using single-quotes in your text field, it'd make a query like this (printed via `Prelude.print`):

```
[...]

PR #645, opened 2021-01-02
Escape single quotes

k-bx:
Not 100% where `parseValues` is used, so this might be completely wrong.

Fixes https://github.com/digitallyinduced/ihp/issues/642 (partially)

### DIFF
--- a/IHP/IDE/Data/Controller.hs
+++ b/IHP/IDE/Data/Controller.hs
@@ -182,7 +182,7 @@ fetchRows connection tableName = do
 parseValues :: Bool -> Bool -> Text -> Text
 parseValues _ True "on" = "true"
 parseValues _ True "off" = "false"
-parseValues False False text = "'" <> text <> "'"
+parseValues False False text = "'" <> (T.replace "'" "''" text) <> "'"
 parseValues True False text = text
 parseValues False True text = text
 parseValues True True text = text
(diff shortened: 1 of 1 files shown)


k-bx:
not sure that error is mine ^

mpscholten:
I'd rather do this a little bit different and use the Postgresql Parameter Substitution.

There's this action constructor here: https://hackage.haskell.org/package/postgresql-simple-0.6.2/docs/Database-PostgreSQL-Simple-ToField.html

We could change `parseValues` to return a `Action` instead of text. The function then could look like this:

```haskell
parseValues :: Bool -> Bool -> Text -> Action
parseValues _ True "on" = toField True
parseValues _ True "off" = toField False
parseValues False _ text = toField text
parseValues _ _ text = Plain text
```

Then at all call-sites of the `parseValues` function we need to change the calls from `PG.execute_` to `PG.execute`. The `PG.execute` takes a list of parameters to substitute as a additional argument (the query should have a `?` for every parameter in the sql string).

So bascially it should look like this at the call site:
```haskell
let values :: [Action] = ...
let query = "INSERT INTO " <> tableName <> " VALUES (" <> map (const "?") values <> ")"
PG.execute connection (PG.Query . cs $! query) values
```

k-bx:
Good point! Substitution params are definitely the way to go, I wasn't familiar enough with IHP codebase thus proposed a minimal change, but thanks to your comment it wasn't hard to re-do via the param substitution.

Should be good to go now!

p.s.: the other thing with validation and errors needs quite a bit of work and I don't know if I'll have time for that, maybe at some point in future :)

mpscholten (approved):
Perfect, thanks :)

Issue #528, opened 2020-11-01
Default value Enum is missing quotes '


themmes:
Generated tables with a column with Enum type can have a default value, however the default value is missing single quotes.

themmes:
Hmm.. Now that I think of this, I remember seeing some ticket related to showing the user to add these quotes themselves...

mpscholten:
Yeah we had some issue regarding this already. But I think this was only related to the Data Editor (not the schema designer). As long as it's still confusing we should improve this

mpscholten:
Wrong ticket closed

Issue #571, opened 2020-11-24
Enum values of multiple words are incorrectly interpreted


themmes:
In translation to Haskell the enums are capitalised, however, they should probably be CamelCased and spaces removed?

Schema
```sql
CREATE TYPE mood AS ENUM ('happy', 'very happy', 'sad', 'very sad');
```

build/Generated/Types.hs
```haskell
data Mood = Happy | Very happy | Sad | Very sad deriving (Eq, Show, Read, Enum)
instance FromField Mood where
    fromField field (Just value) | value == (Data.Text.Encoding.encodeUtf8 "happy") = pure Happy
    fromField field (Just value) | value == (Data.Text.Encoding.encodeUtf8 "very happy") = pure Very happy
    fromField field (Just value) | value == (Data.Text.Encoding.encodeUtf8 "sad") = pure Sad
    fromField field (Just value) | value == (Data.Text.Encoding.encodeUtf8 "very sad") = pure Very sad
    fromField field (Just value) = returnError ConversionFailed field ("Unexpected value for enum value. Got: " <> Data.String.Conversions.cs value)
    fromField field Nothing = returnError UnexpectedNull field "Unexpected null for enum value"
instance Default Mood where def = Happy
instance ToField Mood where
    toField Happy = toField ("happy" :: Text)
    toField Very happy = toField ("very happy" :: Text)
    toField Sad = toField ("sad" :: Text)
    toField Very sad = toField ("very sad" :: Text)
instance InputValue Mood where
    inputValue Happy = "happy" :: Text
    inputValue Very happy = "very happy" :: Text
    inputValue Sad = "sad" :: Text
    inputValue Very sad = "very sad" :: Text
```

Resulting in error
```
build/Generated/Types.hs:1118:40
    Multiple declarations of ‘Very’
```

themmes:
Aha, so if I use underscores it does actually work as intended

mpscholten:
Good catch, we should improve the code gen to map `very happy` to `VeryHappy` 👍

PR #646, opened 2021-01-02
Enum values can contain spaces now

mpscholten:
Fixes https://github.com/digitallyinduced/ihp/issues/528

### DIFF
--- a/IHP/NameSupport.hs
+++ b/IHP/NameSupport.hs
@@ -13,6 +13,7 @@ module IHP.NameSupport
 , fieldNameToColumnName
 , escapeHaskellKeyword
 , tableNameToControllerName
+, enumValueToControllerName
 , toSlug
 ) where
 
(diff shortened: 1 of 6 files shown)


mpscholten:
Fixes https://github.com/digitallyinduced/ihp/issues/571

IHP/NameSupport.hs
```diff
@@ -57,6 +61,31 @@ tableNameToControllerName tableName = do
...
+        words = 
+                enumValue
+                |> splitOn " "
+                |> List.map (Inflector.parseSnakeCase [])
+                |> List.map (\case
+                        Left failed -> error (cs $ "enumValueToControllerName failed for " <> show failed)
+                        Right result -> result)
+                |> join
+    in
+        Inflector.camelizeCustom True words
```
nghamilton, line 87: This PR results in camel-case data constructors for enums that are all-caps. As an example, an enum of `'USD'` is now produces a `Usd`, whereas the generated data constructor was previously `USD`. 

Is this the intended behaviour?

nghamilton (commented):
@mpscholten Is this expected behaviour of the code gen for all-caps enums, or would you like me to create an issue?

IHP/NameSupport.hs
mpscholten: Yes, this is expected behavior 👍

PR #647, opened 2021-01-02
Make tailwindcss styled flash messages consistent

zarak:
Teal is no longer in the default colour scheme of **tailwindcss** v2 (@kodeFant thanks for the tip!) so I'm using green instead.
I also made the styling of `SuccessFlashMessage` otherwise identical to `ErrorFlashMessage`.


mpscholten (approved):
Perfect, thanks :)

PR #648, opened 2021-01-03
support LocalTime for dateField and dateTimeField

fegu:
While LocalTime was supported in the IDE for DB columns, it was not supported in the input fields.
Since localtime vs utc is just a flag in postgresql (just different field types, identical storage) no time zones are involved. It is just a matter of interpretation (local time or utc). Hence, we don't need any change in settings in the flatpickr and we happily parse Z(ulu) just as for UTCTime.

### DIFF
--- a/IHP/Controller/Param.hs
+++ b/IHP/Controller/Param.hs
@@ -399,6 +399,24 @@ instance ParamReader UTCTime where
     readParameterJSON (Aeson.String string) = readParameter (cs string)
     readParameterJSON _ = Left "ParamReader UTCTime: Expected String"
 
+-- | Accepts values such as @2020-11-08T12:03:35Z@ or @2020-11-08@
+instance ParamReader LocalTime where
+    {-# INLINABLE readParameter #-}
+    readParameter "" = Left "ParamReader LocalTime: Parameter missing"
+    readParameter byteString =
+        let
+            input = (cs byteString)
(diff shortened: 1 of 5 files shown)


IHP/Controller/Param.hs
```diff
@@ -399,6 +399,24 @@ instance ParamReader UTCTime where
     readParameterJSON (Aeson.String string) = readParameter (cs string)
     readParameterJSON _ = Left "ParamReader UTCTime: Expected String"
 
+-- | Accepts values such as @2020-11-08T12:03:35Z@ or @2020-11-08@
+instance ParamReader LocalTime where
```
mpscholten, line 403: Could you add a test for this? [Likely we can just copy the UTCTime tests](https://github.com/digitallyinduced/ihp/blob/master/Test/Controller/ParamSpec.hs#L249)

fegu: done

fegu:
btw, for the last few weeks doing this in the IHP dir:
nix-shell
ghci
:l Test/Main.hs

gives:

Test/Controller/ContextSpec.hs:7:1: error:
    Could not find module ‘Test.Hspec’

mpscholten:
I think that issue has been fixed already on latest master. Can you make sure that your IHP dir is up to date? I just tested it with `f687c75e9f39e060042cabf35114d3f4d99ff973` and everything worked fine

mpscholten (approved):

PR #650, opened 2021-01-04
improvment

codemastersk:


s0kil (commented):
This change does not make it easier to understand.
"Contributing" is good enough on it's own.

mpscholten:
Agree with @s0kil We're happy to merge substantial changes that are useful users of IHP.

PR #661, opened 2021-01-09
Fix Haddock Script

s0kil:
When running `./build-haddock` it fails due to some Haskell packages not being loaded, so now we run it inside a `nix-shell`.


mpscholten (approved):
👍

PR #666, opened 2021-01-09
install and first project doc updates

fegu:
installation updates to fit with acctual flow of windows install
removed $ in bash prompts in first-project to conform with install doc
added .vscode to gitignore


fegu:
Removed {renderflashmessags} from login view because it causes them to render twice (first as part of the default layout).

mpscholten (approved):
Perfect, thanks :)

PR #667, opened 2021-01-10
Reduce Empty Space In Readme Logo

s0kil:
Demo: https://github.com/s0kil/ihp/blob/s0kil/readme-logo-update/README.md


mpscholten (approved):
Thanks :)

PR #669, opened 2021-01-10
added empty array '{}' in default-dropdown for array columns in IDE

fegu:


fegu:
Smallint is perhaps rarely used, but I needed it for an existing database. Noticed that bigint support was already there, but missing in the IDE. Now they are both fully supported. Smallint maps to Haskell int so the small part is just for the database.

fegu:
in the last commit, there should be 2 tests for bigint, the second one using INT8. I could not make that work. It fails with "unexpected '8'" even though INT8 is defined where it should (just like INT2 and INT4). This is not important, but would be nice to know why and to make it complete, let me know if you know.

IHP/IDE/SchemaDesigner/View/Columns/Edit.hs
```diff
@@ -114,7 +114,9 @@ typeSelector :: Maybe PostgresType -> [Text] -> Html
 typeSelector postgresType enumNames = [hsx|
         <select id="typeSelector" name="columnType" class="form-control select2-simple">
             {option selected "TEXT" "Text"}
-            {option selected "INT" "Int"}
+            {option selected "SMALLINT" "Int (16bit)"}
+            {option selected "INT" "Int (32bit)"}
```
mpscholten, line 118: Can we keep this one just `Int`? Otherwise new users might be confused on what Int type to pick :)

```suggestion
            {option selected "INT" "Int"}
```

Maybe we can also then move small and big ints near the end of the list (the list is sorted by general frequency of usage)?

IHP/IDE/SchemaDesigner/Parser.hs
```diff
@@ -187,8 +187,9 @@ sqlType :: Parser PostgresType
 sqlType = choice $ map optionalArray
         [ uuid
         , text
-        , bigint
+        , smallint
         , int
```
mpscholten, line 191: The `"unexpected '8'" ` error likely happens here. The `int` parser is called before the `bigint` parser. `int` has a `try (symbol' "INT")`. Given an input string `INT8`, this `int` parser will consume the `INT` prefix and then is left with just `8`. Therefore the above error.

We can fix this by calling the `bigint` parser before the `int` parser.
Like:
```haskell
, smallint
, bigint
, int
```

fegu: Of course, it was getting late last night. Fixed, with comment to help the next guy as well.

fegu: (github didn't allow committing suggested change after I had fixed INT8 (outdated diff))
I like the suggestion of having less-used types at the end of the list, in this way we can cater to everyone but not confuse newbies.

mpscholten (approved):
Perfect, thanks :)

PR #670, opened 2021-01-11
Fix distinctOn QueryBuilder Function

s0kil:
`#fullName` should map to `full_name` in the database.


mpscholten (approved):
👍 perfect, thanks :)

PR #672, opened 2021-01-11
categorized columntypes in IDE and fixed bug in telemetry WSL

fegu:
Two fairly random things to put in the same commit, I know.
The WSL bug: up to now only the older WSL1 was reported, now it reports WSL2 as well. Both are reported as WSL because we really don't care.

### DIFF
--- a/IHP/IDE/SchemaDesigner/View/Columns/Edit.hs
+++ b/IHP/IDE/SchemaDesigner/View/Columns/Edit.hs
@@ -113,31 +113,35 @@ instance View EditColumnView where
 typeSelector :: Maybe PostgresType -> [Text] -> Html
 typeSelector postgresType enumNames = [hsx|
         <select id="typeSelector" name="columnType" class="form-control select2-simple">
-            {option selected "TEXT" "Text"}
-            {option selected "INT" "Int"}
-            {option selected "UUID" "UUID"}
-            {option selected "BOOLEAN" "Bool"}
-            {option selected "TIMESTAMP WITH TIME ZONE" "Timestamp (UTCTime)"}
-            {option selected "TIMESTAMP WITHOUT TIME ZONE" "Timestamp (LocalTime)"}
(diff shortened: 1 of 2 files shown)


IHP/IDE/SchemaDesigner/View/Columns/Edit.hs
```diff
@@ -113,38 +113,47 @@ instance View EditColumnView where
...
+                {option isSelected "INT" "Int"}
+                {option isSelected "UUID" "UUID"}
+                {option isSelected "BOOLEAN" "Bool"}
+                {option isSelected "DATE" "Date / Day"}
+                {option isSelected "TIMESTAMP WITHOUT TIME ZONE" "Timestamp (LocalTime)"}
+                {option isSelected "SERIAL" "Serial"}
+            </optgroup>
+            {customenums enumNames}
+            <optgroup label="Other Types">
+                {option isSelected "TIMESTAMP WITH TIME ZONE" "Timestamp (UTCTime)"}
```
mpscholten, line 127: `TIMESTAMP WITH TIME ZONE` is the default type used for `created_at` fields in the schema designer (or generally any field whose name ends with `_at`). I think therefore we should have it in Common Types.

[I'm a big fan of storing all timestamps in the db in UTC](https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_timestamp_.28without_time_zone.29_to_store_UTC_times). Any reasons we cannot move the `TIMESTAMP WITHOUT TIME ZONE` into `Other Types` to `Timestamp UTCTime` the default suggested timestamp type?

fegu: No problem, pushed change now. As long as my types are there, it isn't important where they are in the list. I'm more of a localtime (or time with actual timezone) kind of guy. UTC has given me unnecessary complexity in the past. But UTC is a good choice for created_at etc.

mpscholten (approved):

PR #678, opened 2021-01-13
Inet

mpscholten:


fegu (approved):
Missing IDE support and docs. But can probably wait, a lot of type support has started out this way. Looks good. Is there a complete match between postgres INET type and the ip package (even for CIDR notation?). If not, should add comments (or docs).

mpscholten:
Thanks for the fast review :) Will add docu and IDE tomorrow 👍  CIDR is not supported (we'd need to map it to a different data type in that case). My main use case is storing ip's, so it's optimized for that.

PR #682, opened 2021-01-14
Add query builders for LIKE and ~ queries

jakosimov:
Add the functions `filterWhereLike` and `filterWhereMatches` for `WHERE a LIKE b` and `WHERE a ~ b` queries respectively. There are also case-insensitive versions of these functions.

### DIFF
--- a/IHP/QueryBuilder.hs
+++ b/IHP/QueryBuilder.hs
@@ -27,6 +27,10 @@ module IHP.QueryBuilder
 , DefaultScope (..)
 , filterWhereIn
 , filterWhereNotIn
+, filterWhereLike
+, filterWhereILike
+, filterWhereMatches
+, filterWhereIMatches
 , genericFetchId
 , genericfetchIdOneOrNothing
 , genericFetchIdOne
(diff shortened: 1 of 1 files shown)


IHP/QueryBuilder.hs
```diff
@@ -377,6 +389,22 @@ filterWhereNotIn (_, []) = id -- Handle empty case by ignoring query part: `WHER
 filterWhereNotIn (name, value) = FilterByQueryBuilder (name, NotInOp, toField (In value))
 {-# INLINE filterWhereNotIn #-}
 
+filterWhereLike :: forall name table model value. (KnownSymbol name, ToField value, HasField name model value, model ~ GetModelByTableName table) => (Proxy name, value) -> QueryBuilder table -> QueryBuilder table
```
mpscholten, line 392: Can we add haddock blocks to all these new functins, like this?

```haskell
-- | Adds a simple @WHERE x LIKE y@ condition to the query.
--
-- __Example:__ (we need to think of a real world like example here)
--
-- > activeProjects <- query @Project
-- >     |> filterWhere (#active, True)
-- >     |> fetch
-- > -- SELECT * FROM projects WHERE active = True
```

Those show up autocompletion when using haskell language server, so it's kind of useful :)

jakosimov:
I fixed some documentation for the functions!

Perhaps my examples could use some improvement though (:

mpscholten (approved):
Perfect, thanks :)

PR #686, opened 2021-01-15
Update "Installing IHP" with pointer to nix/zsh path issue on OSX

ijcd:
Nix install on OSX with zsh sometimes does not result in a path with nix executables present. Add doc pointer to cover this.

### DIFF
--- a/Guide/installation.markdown
+++ b/Guide/installation.markdown
@@ -24,6 +24,16 @@ After this restart your terminal.
 
 If you get an error like `error: refusing to create Nix store volume because the boot volume is FileVault encrypted, but encryption-at-rest is not available.`, follow the steps described [in this GitHub Issue](https://github.com/digitallyinduced/ihp/issues/93#issuecomment-639611313). We're working on improving this step.
 
+If `ihp-new` is not in your path using the zsh shell ([see this GitHub Issue](https://github.com/NixOS/nix/issues/1577), you may need to add the following to your dotfiles:
+
+```
+# Nix
+if [ -e '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh' ]; then
+  . '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
+fi
+# End Nix
+```
+
 ### Linux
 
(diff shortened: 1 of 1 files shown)


Guide/installation.markdown
```diff
@@ -24,6 +24,16 @@ After this restart your terminal.
 
 If you get an error like `error: refusing to create Nix store volume because the boot volume is FileVault encrypted, but encryption-at-rest is not available.`, follow the steps described [in this GitHub Issue](https://github.com/digitallyinduced/ihp/issues/93#issuecomment-639611313). We're working on improving this step.
 
+If `ihp-new` is not in your path using the zsh shell ([see this GitHub Issue](https://github.com/NixOS/nix/issues/1577), you may need to add the following to your dotfiles:
+
+```
+# Nix
+if [ -e '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh' ]; then
+  . '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
```
s0kil, line 32: Did the `/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh` path work on your machine? I have a different path on my machine (`/nix/var/nix/profiles/per-user/danielsokil/profile/etc/profile.d/nix-daemon.sh`), maybe it's best to stick with what [Nix recommends](https://nixos.org/manual/nix/stable/#sect-single-user-installation) `source ~/.nix-profile/etc/profile.d/nix.sh`.

ijcd: Ok, this is really weird. This did work the first time I ran it but now it's gone and the shell is working fine. I'm baffled for the moment... investigating.

ijcd: I'm seeing this in `zshenv` now, but not sure what added it... it wasn't there (even for new terminals) before I started poking around. 

I'm going to withdraw this PR for now.

```
if [ -e /Users/ijcd/.nix-profile/etc/profile.d/nix.sh ]; then . /Users/ijcd/.nix-profile/etc/profile.d/nix.sh; fi # added by Nix installer
```

I do think it's a bit rude to throw things into dotfiles without prompting if IHP is doing that.

ijcd:
Let's not do this for now. I'm not sure what ultimately fixed my shell.

Issue #384, opened 2020-09-09
Split nix derivations for library and executables


ruhatch:
While hacking on the IHP library, small changes to e.g. DevServer require rebuilding the entire IHPFramework library. It would be good to try and split these into smaller nix derivations to make them easier to work with.

mpscholten:
I'm always loading the dev server into my ghci. IMO this way you have the best feedback loop.

ruhatch:
> I'm always loading the dev server into my ghci. IMO this way you have the best feedback loop.

Yeah, I've been doing that as well, but sometimes you need to keep building the executable! It's not a huge deal though, so feel free to close this one :shrug:

mpscholten:
Ok, just wanted to make sure :) happy to keep it open 👍

PR #688, opened 2021-01-16
Faster builds

mpscholten:
@s0kil can you try this locally?

Also try to enable the fast build config in `ihp.nix` locally

Fixes https://github.com/digitallyinduced/ihp/issues/384


s0kil:
@mpscholten This is great! Compile times went from 30 minutes to 5 minutes.
Another thing that would help is telling haddock to skip processing the IDE modules, it takes quite some time to parse the files and results in `Missing documentation for ...` warnings.

s0kil:
I've also looked through the compiler flags, and seems we have every possible optimization enabled (:

There was another one that looked interesting [`-flate-specialise`](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/using-optimisation.html#ghc-flag--flate-specialise)

PR #693, opened 2021-01-17
Faster Builds

s0kil:
Intended to be merged back into https://github.com/digitallyinduced/ihp/pull/688

~~It seems flags defined in `*.cabal` [default to true](https://cabal.readthedocs.io/en/3.4/setup-commands.html#controlling-flag-assignments), so I've added `-FastBuild` to explicitly define it to false.~~
Also added the `-flate-specialise` flag

~~Since the flags default to true, now it makes more sense to have `OptimizedBuild` instead of `FastBuild`.~~

### DIFF
--- a/ihp.cabal
+++ b/ihp.cabal
@@ -132,6 +132,7 @@ common shared-properties
             -fconstraint-solver-iterations=100
             -fexpose-all-unfoldings
             -flate-dmd-anal
+            -flate-specialise
             -fmax-worker-args=200
             -fspec-constr-keen
             -fspecialise-aggressively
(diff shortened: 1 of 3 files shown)


s0kil:
I've just realized that it's possible to [set a default value](https://cabal.readthedocs.io/en/3.4/cabal-package.html#pkg-field-flag-default), If it's better I will revert the changes made here and use a default value for FastBuild instead.

mpscholten:
Then let's use a default value with FastBuild. Naming it `OptimizedBuild` could be confusing as cabal also has a built-in `--disable-optimizations` flag which basically just sets `-O0`

mpscholten (approved):
Thanks :)

PR #700, opened 2021-01-19
Add support for SVG elements in hsx

zarak:
I added all the elements that weren't already in the `parents` list from the link below. I wasn't sure whether to name a new list and/or sort alphabetically, but hopefully it at least checks out functionally.

Element index [reference](https://www.w3.org/TR/SVG2/eltindex.html)

W3C Candidate Recommendation
Appendix F: Element Index


mpscholten (approved):
Thanks :)

PR #701, opened 2021-01-19
Allow hsx to parse spaces in closing tags

zarak:
Some people structure their HTML code in a way that introduces spaces in
the closing tags. This is valid HTML, but the hsx parser will
complain when copying the HTML into hsx.

The following should be valid hsx after this change:
```html
<!-- Example 1 -->
<span>
    <i class="fab fa-html5"></i
                        >
</span>

<!-- Example 2 -->
<span>
    <i class="fab fa-html5"></i >
</span>
```


mpscholten:
Could you also add one test case for this newly allowed syntax?

mpscholten (approved):
Thanks :) Good job

PR #706, opened 2021-01-20
Add make command to start (only) postgres server

themmes:
After I asked about this in the Slack, @s0kil suggested to add it as a make command. I really liked that idea!

### DIFF
--- a/lib/IHP/Makefile.dist
+++ b/lib/IHP/Makefile.dist
@@ -106,6 +106,9 @@ build/bin:
 run-production: build/bin/RunProdServer build/Generated/Types.hs ## Run Production Server
 	build/bin/RunProdServer
 
+postgres: ## Starts the postgresql server
+	postgres -D build/db/state -k $PWD/build/db -c "listen_addresses="
+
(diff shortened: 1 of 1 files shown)


lib/IHP/Makefile.dist
```diff
@@ -106,6 +106,9 @@ build/bin:
 run-production: build/bin/RunProdServer build/Generated/Types.hs ## Run Production Server
 	build/bin/RunProdServer
 
+postgres: ## Starts the postgresql server
```
s0kil, line 109: Would be a bit better to have a message if the database is already running.
```
postgres: ## Starts the postgresql server
	@if [ -s build/db/state/postmaster.pid ]; then
		echo "Postgres is already running."
	else
		echo "Starting database"
		postgres -D build/db/state -k $$PWD/build/db -c "listen_addresses="
	fi
```

themmes: Ah thats even better, added it!

mpscholten (approved):

PR #709, opened 2021-01-24
Remove old "migration" docs

themmes:
With the new migration docs in `Guide/database-migrations.md` this part is redundant and potentially confusing

### DIFF
--- a/Guide/database.markdown
+++ b/Guide/database.markdown
@@ -503,66 +503,6 @@ In your views, use `inputValue` to get a textual representation for your enum wh
 |]
 ```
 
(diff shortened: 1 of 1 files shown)


mpscholten:
It's still the way I do migrations in dev mode. The other migration feature is mainly designed for production use. Maybe we need to make more clear that these processes are designed for different environments ?

themmes:
You're right. I removed this part as well, because the update/push/save are mentioned in multiple places. However, without this section it might lack a bit of detail. I'll see if I can reintroduce the detail in a place where they're already mentioned.

PR #713, opened 2021-01-25
Add basic documentation for jobs

zacwood9:
Alternative title: my Sunday afternoon figuring out how jobs work :-) 
I definitely plan on writing about this more in-depth in the future, but for the documentation I just included the basics that I got stuck on as I was trying to figure things out.

### DIFF
--- a/Guide/jobs.markdown
+++ b/Guide/jobs.markdown
@@ -37,7 +37,9 @@ import Web.Controller.Prelude
 instance Job EmailCustomersJob where
     perform EmailCustomersJob { .. } = do
       customers <- query @Customer |> fetch
-      mapM_ (\customer -> sendMail (MarketingMail customer)) customers
+      forEach customers sendToCustomer
+      where 
+        sendToCustomer customer = sendMail (MarketingMail customer)
 ```
 
 ### Running the job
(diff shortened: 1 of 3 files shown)


Guide/jobs.markdown
```diff
@@ -0,0 +1,89 @@
...
+A pseudocode implementation of a job that sends a marketing email to all of the customers:
+
+```haskell
+module Web.Job.EmailCustomers where
+import Web.Controller.Prelude
+
+instance Job EmailCustomersJob where
+    perform EmailCustomersJob { .. } = do
+      customers <- query @Customer |> fetch
+      mapM_ (\customer -> sendMail (MarketingMail customer)) customers
```
mpscholten, line 40: I'd use `forEach` here which is like mapM_ but arguments reversed. This feels more like the IHP approach :)

mpscholten (commented):
This is great, thanks :)

Guide/jobs.markdown
zacwood9: Never thought of using `forEach` outside of views! Thanks

mpscholten (approved):
Perfect, thanks :)

PR #717, opened 2021-01-26
forceRedirectToPath

fegu:
This addition has been discussed in slack a few times.
To keep the footprint small (it is a workaround, after all), I have added just one method instead of the full redirectX-set.
This workaround is mostly useful just for actions and paths anyway.
Very useful when you need it.

### DIFF
--- a/IHP/Controller/Redirect.hs
+++ b/IHP/Controller/Redirect.hs
@@ -3,7 +3,7 @@ Module: IHP.Controller.Redirect
 Description: redirect helpers
 Copyright: (c) digitally induced GmbH, 2020
 -}
-module IHP.Controller.Redirect (redirectTo, redirectToPath, redirectToUrl) where
+module IHP.Controller.Redirect (redirectTo, redirectToPath, redirectToUrl, forceRedirectToPath) where
 
 import IHP.Prelude
 import qualified Network.Wai.Util
@@ -12,9 +12,12 @@ import IHP.Controller.RequestContext
 import IHP.RouterSupport (HasPath (pathTo))
 import IHP.FrameworkConfig
 import qualified Network.Wai as Wai
+import qualified Data.Text.Encoding as TE
 import Data.String.Conversions (cs)
 import Data.Maybe (fromJust)
 import Network.HTTP.Types (status200, status302)
+import Network.HTTP.Types.Status
+import Network.HTTP.Types.Header (hLocation)
 import GHC.Records
 
 import IHP.Controller.Context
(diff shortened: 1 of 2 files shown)


IHP/Controller/Redirect.hs
```diff
@@ -44,6 +47,18 @@ redirectToPath :: (?context :: ControllerContext) => Text -> IO ()
...
+-- | like 'redirectToPath', but forcing full page reload
+--
+-- Forces reload by using a custom HTTP OK header mimicking a HTTP redirect
+-- which is used as a signal to the AJAX call to perform page reload.
+-- currently this is a workaround of last resort when you can't make your Javscript 
+-- code behave properly together with morphdom and/or turbolinks
+-- 
+-- use 'forceRedirectToPath (pathTo action)' if you want to redirect to a controller action
+forceRedirectToPath :: (?context :: ControllerContext) => Text -> IO ()
+forceRedirectToPath path = respondAndExit $ Wai.responseLBS (Status 280 "IHP ForceRedirect") [(hLocation,TE.encodeUtf8 (fromConfig baseUrl <> path))] ""
```
mpscholten, line 59: The more IHP-ish way would be to use `cs` instead of manually calling `encodeUtf8`. This also saves one import :)

```suggestion
forceRedirectToPath path = respondAndExit $ Wai.responseLBS (Status 280 "IHP ForceRedirect") [(hLocation,  cs (fromConfig baseUrl <> path))] ""
```

mpscholten (approved):

ByteEater-pl:
The release notes say:
> Added forceRedirectToPath to work around JS issues

Are they JS (the language) issues though? I haven't been able to confirm it from either the code changer or the above discussion and found nothing more on it through issue search.

mpscholten:
The issues have been related to our `helpers.js` that overrides the browser's default form submit. This sometimes causes issues if the form submissions leads to a redirect that e.g. causes a SPA to be started. In that case the best behavior is to disable the javascript form submission and rely on native browser form submission.

But previously this was only possible by removing the `helpers.js` `<script>` tag and so disabling the JS helpers. As a workaround the `forceRedirectToPath` was added. With `forceRedirectToPath` it was possible to trigger a native browser redirect even with `helpers.js` enabled.

[In the latest IHP version the `helpers.js` form submissions via JS can be disabled for a specific form](https://ihp.digitallyinduced.com/Guide/form.html#disable-form-submission-via-javascript). Therefore `forceRedirectToPath` has been removed in recent versions.

Issue #716, opened 2021-01-26
Proposal: IHP.Log, a unified logging solution for IHP apps


zacwood9:
# Introduction

This proposal introduces the `IHP.Log` module, a set of logging functions built on top of [fast-logger](https://hackage.haskell.org/package/fast-logger). Ideally, this would be the only way the framework logs information during its execution, allowing the user complete control over the format of logs, logging levels, and log destination.

## Logging in IHP today

Currently, IHP logs information about its execution through a variety of means.
- Database queries are logged in development mode only through the [`logQuery` function](https://github.com/digitallyinduced/ihp/blob/master/IHP/ModelSupport.hs#L355). This logs the query and its parameters through `putStrLn` to standard out.
- Requests are logged through `wai-extra` middleware, with defaults defined in [IHP configuration](https://github.com/digitallyinduced/ihp/blob/master/IHP/FrameworkConfig.hs#L94). By default this is logged to standard out but can also be directed to a file.
-  Exceptions during  the request cycle are [logged to standard error](https://github.com/digitallyinduced/ihp/blob/master/IHP/ErrorController.hs#L103).
- Logs in user application can be done through any means: `putStrLn`, or a custom logging framework.

## What's missing?

In order to collect and analyze logs in production applications, a common flow in other frameworks is to log into files that are collected and uploaded to external services. With current IHP logging, users would have to redirect stdin/stdout to a file to store logs. This solution makes rotating log files much harder and error prone.

Also, current IHP logging has no concept of log levels, a common idea in logging libraries. With log levels, users can specify what level of logs they would like output (i.e. in development all logs, in production just warnings and errors). This allows for logging flexibility -- for example, if a user would like to log query execution time in production to debug bad query performance, this would not be possible the built in IHP logging.

## Proposal: IHP.Log

I propose adding a module to the framework, `IHP.Log`, that provides functions to configure the application logger as well as functions to log at different log levels. By building on top of a battle-tested library such as `fast-logger`, we can inherit a lot of great features such as:
- automatic log rotating
- great performance, especially in multi-threaded environments
- extensive configuration options
- log formatting

I imagine the logging configuration be stored in `FrameworkConfig` and be configured in `Config.hs`.

```haskell
config :: ConfigBuilder
config = 
  ...
  option Production
  option $ LogFileTimedRotate "Log/production.log" rotateDaily
  option $ LogDateFormat "some time format string"
  option $ LogLevel Warning -- log warnings and greater
[...]

PR #718, opened 2021-01-27
IHP.Log module

zacwood9:
An initial implementation of the module described in the proposal (#716).

Two modules `IHP.Log.Types` and `IHP.Log.Logging` were added, and most internal logging was changed to use the new Logging system. Basic ideas and implementation so far

- in any `ApplicationContext` or `ModelContext`, you can call `Log.<debug|info|warn|error|fatal|unknown>` to log a message at that log level. 
- Configure the logging behavior in Config.hs. 
  - development logger defaults to Debug level, and production to Warn level.
  - logs default to stdout
  - can customize the format by providing a custom `LogFormatter`, takes the LogLevel and message currently, but time stamp should also be passed 

Example configuration:

```haskell
config :: ConfigBuilder
config = do
    ...
    logger <- liftIO $ case environment of
        Development -> defaultLogger
        Production -> newLogger Warn
    option logger
```

Some example code of a usage in my app:

```haskell
  action TopPerformancesAction {collection} = do
      Log.debug "starting TopPerformancesAction"
      let n = paramOrDefault 5 "numPerformances"
      band <- fetchBand collection
      topPerformances <- fetchTopPerformances collection n
      Log.debug $ show (length topPerformances) <> " top performances received."
      when (length topPerformances == 0) $ Log.warn "No performances found! Something might be wrong"
      render TopPerformancesView {..}
```

Still left to do
- Log exceptions at error level
- Log requests at info level
- provide timestamp to formatter
- tests of some sort
- provide access to all the fast-logger log types

## Feedback!!

Is this something y'all would consider merging once it's finished? I want to make sure there's interest in this before I put any serious time into it :)


mpscholten:
Off topic: In your example code the expression `when (length topPerformances == 0)` could be replaced with IHP's whenEmpty function `whenEmpty topPerformances` :) It's shorter.

On topic: I think this is valuable as we're going to need this at some point anyways. So in general I'd be happy to merge this once ready 👍

zacwood9:
Hey @mpscholten, this is finished now! Check out the included documentation page for a fuller explanation on how it works and can be configured.

mpscholten:
There seems to be a type error: I guess notConnectedModelContext needs to be passed a logger
```
IHP/IDE/IHP/IDE/ToolServer.hs:79:66: error:
    • Couldn't match expected type ‘ModelContext’
                  with actual type ‘IHP.Log.Types.Logger -> ModelContext’
    • Probable cause: ‘notConnectedModelContext’ is applied to too few arguments
      In the ‘modelContext’ field of a record
      In the expression:
        ApplicationContext
          {modelContext = notConnectedModelContext, session,
           autoRefreshServer, frameworkConfig}
      In an equation for ‘applicationContext’:
          applicationContext
            = ApplicationContext
                {modelContext = notConnectedModelContext, session,
                 autoRefreshServer, frameworkConfig}
   |
79 |     let applicationContext = ApplicationContext { modelContext = notConnectedModelContext, session, autoRefreshServer, frameworkConfig }
```

Guide/logging.markdown
```diff
@@ -0,0 +1,137 @@
...
+  Stderr          :: BufSize -> LogDestination
+  -- | Log messages to a file which is never rotated.
+  FileNoRotate    :: FilePath -> BufSize -> LogDestination
+  -- | Log messages to a file rotated automatically based on the critera in 'FileLogSpec'.
+  File            :: FileLogSpec -> BufSize -> LogDestination
+  -- | Log messages to a file rotated on a timed basis as defined in 'TimedFileLogSpec'.
+  FileTimedRotate :: TimedFileLogSpec -> BufSize -> LogDestination
+  -- | Send logged messages to a callback. Flush action called after every log.
+  Callback        :: (LogStr -> IO ()) -> IO () -> LogDestination
+```
```
mpscholten, line 130: Can we add examples here for:
- File
- FileNoRotate
- FileTimedRotate

These are most likely to be used by some user when the user is not happy with the stdout logging

Guide/logging.markdown
```diff
@@ -0,0 +1,137 @@
...
+
+#### Configuring timestamp format
+
+`timeFormat` expects a time format string as defined [here](https://man7.org/linux/man-pages/man3/strptime.3.html).
+
```
mpscholten, line 135: Can we also have a code example here?

Guide/logging.markdown
```diff
@@ -0,0 +1,137 @@
...
+- `withLevelFormatter` prepends the message's log level
+  - `[INFO] Server started`
+- `withTimeAndLevelFormatter` prepends both a timestamp and log level.
+  - `[INFO] [28-Jan-2021 10:07:58] Server started`
+
+You can also define you own formatter:
+
+```haskell
+type LogFormatter = FormattedTime -> LogLevel -> Text -> Text
+```
```
mpscholten, line 106: Can we add an example of a custom log formatter (with configuration) here?

Guide/logging.markdown
```diff
@@ -0,0 +1,137 @@
...
+}
+```
+
+#### Configuring log level
+
+Set `level` to one of the available constructors for the `LogLevel` type:
+
+```haskell
+data LogLevel =
+  Debug
```
mpscholten, line 81:
```suggestion
data LogLevel
  = Debug
```

Guide/logging.markdown
```diff
@@ -0,0 +1,137 @@
...
+
+**Note:** since the logging system is multi-threaded for optimal performance, it is not guaranteed that messages will be printed in order.
+If you need to know exact ordering it's recommended you rely on the timestamp.
+
+### Log levels
+IHP logging uses log levels to determine which messages should be printed.
+This way, you can log messages to help in development without flooding production logs.
+
+The available log levels are `debug`, `info`, `warn`, `error`, `fatal`, and `unknown`.
+Log messages will only be output if their log level is greater than or equal to the logger's configured log level.
```
mpscholten, line 19: What's the default log level here?

IHP/Fetch.hs
```diff
@@ -55,7 +55,6 @@ instance (model ~ GetModelByTableName table, KnownSymbol table) => Fetchable (Qu
     fetch !queryBuilder = do
         let !(theQuery, theParameters) = queryBuilder
                 |> toSQL
-        logQuery theQuery theParameters
```
mpscholten, line 58: Shouldn't this call the logger here?

IHP/FrameworkConfig.hs
```diff
@@ -83,18 +85,29 @@ ihpDefaultConfig = do
...
+
     environment <- findOption @Environment
+
+    maybeLogger <- findOptionMaybe @Logger
+    logger <- liftIO $ case maybeLogger of
+        Just logger -> pure logger
+        Nothing -> case environment of
+            Development -> defaultLogger
+            Production -> newLogger (def { level = Warn } :: LoggerSettings)
+    option logger
```
mpscholten, line 97: Actually the `option` function doesn't override a value if it has been set previously.

So e.g.

```haskell
option $
    case environment of
            Development -> defaultLogger
            Production -> newLogger (def { level = Warn } :: LoggerSettings)
```

would be equivalent to the code above

IHP/Log.hs
```diff
@@ -0,0 +1,77 @@
+{-|
+Module: IHP.Log.Logging
+Description:  Functions to write logs at all log levels.
+
+Import this module qualified!
```
mpscholten, line 5: Can we add a code example here?
```haskell
import qualified IHP.Log as Log
```

IHP/Log.hs
```diff
@@ -0,0 +1,77 @@
+{-|
+Module: IHP.Log.Logging
```
mpscholten, line 2:
```suggestion
Module: IHP.Log
```

IHP/Log.hs
```diff
@@ -0,0 +1,77 @@
...
+Description:  Functions to write logs at all log levels.
+
+Import this module qualified!
+
+-}
+module IHP.Log (
```
mpscholten, line 8: Most IHP code uses a different indentation layout here:
```suggestion
module IHP.Log
( module IHP.Log.Types
, debug
, info
...
```

IHP/Log.hs
```diff
@@ -0,0 +1,77 @@
...
+import Network.Wai.Middleware.RequestLogger (mkRequestLogger, RequestLoggerSettings, destination)
+import qualified Network.Wai.Middleware.RequestLogger as RequestLogger
+import Data.Default.Class (Default (def))
+import Data.String.Conversions (cs)
+import System.IO.Unsafe (unsafePerformIO)
+
+-- | Format a log and send it to the logger.
+log :: (?context :: context, LoggingProvider context) => LogLevel -> Text -> IO ()
+log level text = do
+  let logger = getLogger ?context
```
mpscholten, line 37: Can we indent this with 4 spaces?

IHP/Log.hs
```diff
@@ -0,0 +1,77 @@
...
+import Data.Text as Text
+import IHP.Log.Types
+import Network.Wai (Middleware)
+import Network.Wai.Middleware.RequestLogger (mkRequestLogger, RequestLoggerSettings, destination)
+import qualified Network.Wai.Middleware.RequestLogger as RequestLogger
+import Data.Default.Class (Default (def))
+import Data.String.Conversions (cs)
+import System.IO.Unsafe (unsafePerformIO)
+
+-- | Format a log and send it to the logger.
```
mpscholten, line 34: Can we add a code example here? :) These show up e.g. in IDE's, so it's good when most of the primaryly used functions have code examples

IHP/Log.hs
```diff
@@ -0,0 +1,77 @@
...
+warn = log Warn
+
+error :: (?context :: context, LoggingProvider context) => Text -> IO ()
+error = log Error
+
+fatal :: (?context :: context, LoggingProvider context) => Text -> IO ()
+fatal = log Fatal
+
+unknown :: (?context :: context, LoggingProvider context) => Text -> IO ()
+unknown = log Unknown
```
mpscholten, line 59: Maybe here as well, I know it's a bit redundant but this is good when using haskell language server and vscode :)

IHP/Log/Types.hs
```diff
@@ -0,0 +1,154 @@
...
+data LogLevel =
+  Debug
+  | Info
+  | Warn
+  | Error
+  | Fatal
+  | Unknown
+  deriving (Enum, Eq, Ord, Show)
+
+type FormattedTime = Text
```
mpscholten, line 74: Can we add some examples strings here? :)

IHP/Log/Types.hs
```diff
@@ -0,0 +1,154 @@
...
+type LogFormatter = FormattedTime -> LogLevel -> Text -> Text
+
+-- | Where logged messages will be delivered to. Types correspond with those in fast-logger.
+data LogDestination where
```
mpscholten, line 78: Why are we using GADTs here?

mpscholten: I was also just wondering: Why aren't we just reexporting fast loggers data types here?

IHP/Server.hs
```diff
@@ -139,10 +150,10 @@ application request respond = do
         let ?context = requestContext
         frontControllerToWAIApp RootApplication ErrorController.handleNotFound
 
-runServer :: FrameworkConfig -> Application -> IO ()
-runServer FrameworkConfig { environment = Env.Development, appPort } = Warp.runSettings $
+runServer :: (?applicationContext :: ApplicationContext) => FrameworkConfig -> Application -> IO ()
+runServer config@FrameworkConfig { environment = Env.Development, appPort } = Warp.runSettings $
                 Warp.defaultSettings
-                    |> Warp.setBeforeMainLoop (ByteString.putStrLn "Server started")
+                    |> Warp.setBeforeMainLoop (let ?context = config in Log.info "Server started")
```
mpscholten, line 156: Here we should still stick to the `ByteString.putStrLn`. The dev server [uses this here](https://github.com/digitallyinduced/ihp/blob/master/exe/IHP/IDE/DevServer.hs#L275) to detect that the app server is just about to be started. A custom log formatter or a wrong log level could then break the dev server / live reload here.

IHP/Server.hs
```diff
@@ -38,26 +43,32 @@ import qualified Data.ByteString.Char8 as ByteString
...
-                                        application
+                sessionMiddleware $
+                    ihpWebsocketMiddleware $
+                        requestLoggerMiddleware $
+                                methodOverridePost $
+                                    application
+
+    run `finally` do
+        let ?context = frameworkConfig |> get #logger
+        Log.info "exiting server..."
```
mpscholten, line 69: Is this log message really needed?

mpscholten (commented):
Great job :) Commented a couple of points, but in general this looks like we can merge it soon

IHP/Fetch.hs
zacwood9: I found that all queries constructed with the query builder and called with fetch were being logged twice: once in the fetch functions, and once in `sqlQuery`. Since all the queries built are eventually dispatched to `sqlExec` or `sqlQuery` , which both log the query, the logs here aren't needed. I might be missing something but I'm pretty sure this was a bug before

IHP/Log/Types.hs
zacwood9: GADTs are unnecessary -- switched back to regular data type. Thought it might be necessary for some reason when I started but didn't turn out that way!

I chose to wrap the fast logger types for a couple reasons:
- allow for using alternate logging back-ends in the future by decoupling the user facing IHP code from the dependency
- allow for IHP to document the logger types in its own documentation 
- allow for wrapped configuration types, for example the FileLogSpec and TimedFileLogSpec records are a little unwieldy (and use snake case for the field names???)

That being said the code I wrote here doesn't take much advantage of these benefits: although the types are wrapped they still use fast-logger types to configure them so they aren't decoupled in any meaningful way, and the documentation I have isn't any better than the fast-logger docs.

I'd be happy to simply use the fast-logger types if you'd prefer, or I can also decouple in a more meaningful way and work on extra documentation to get more of a benefit from a custom IHP type. Let me know :)

zacwood9:
Ready for review again :) thanks for the great feedback, I'm not used to writing library code so my bad documentation habits need some fixing 😄

IHP/FrameworkConfig.hs
```diff
@@ -123,6 +134,14 @@ findOption = do
...
 {-# INLINE findOption #-}
 
+findOptionMaybe :: forall option. Typeable option => State.StateT TMap.TMap IO (Maybe option)
+findOptionMaybe = do
+    options <- State.get
+    options
+        |> TMap.lookup @option
+        |> pure
+{-# INLINE findOptionMaybe #-}
+
```
mpscholten, line 144: This one can be removed now, right?

```suggestion
```

ihp.nix
```diff
@@ -125,7 +127,7 @@ mkDerivation {
 
   # For faster builds when hacking on IHP:
   # Uncommenting will build without optimizations
-  # configureFlags = [ "--flag FastBuild" ];
+  configureFlags = [ "--flag FastBuild" ];
```
mpscholten, line 130: Left over :) 
```suggestion
  # configureFlags = [ "--flag FastBuild" ];
```

ihp.nix
```diff
@@ -125,7 +127,7 @@ mkDerivation {
...
   # Uncommenting will not generate documentation
-  # doHaddock = false;
+  doHaddock = false;
```
mpscholten, line 132:
```suggestion
  # doHaddock = false;
```

zacwood9:
Fixed overlooked files and changed how files are logged -- instead of three different file constructors, condensed it into one File constructor and added RotateSettings as an argument.  I think this API is much nicer :)

mpscholten (approved):
Looks alright now :) Thanks a lot for this good work 👍

PR #723, opened 2021-01-28
Flexible Whitespace In Enum Definition

s0kil:
Keep getting parser errors for SQL formatted as:

```sql
CREATE TYPE Stuff AS ENUM (
    'One',
    'Two',
    'Three',
    'Four',
    'Five',
    'Six',
    'Seven',
    'Eight',
    'Nine',
    'Ten'
);
```

Currently IHP only allowed:
```sql
CREATE TYPE Stuff AS ENUM (
    'One',
    'Two',
    'Three',
    'Four',
    'Five',
    'Six',
    'Seven',
    'Eight',
    'Nine',
    'Ten');
```

### DIFF
--- a/IDE/IHP/IDE/SchemaDesigner/Parser.hs
+++ b/IDE/IHP/IDE/SchemaDesigner/Parser.hs
@@ -111,7 +111,8 @@ createEnumType = do
     name <- identifier
     lexeme "AS"
     lexeme "ENUM"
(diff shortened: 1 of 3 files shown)


mpscholten:
Looks good, can we add a test case for both cases to https://github.com/digitallyinduced/ihp/blob/master/Test/IDE/SchemaDesigner/ParserSpec.hs ?

s0kil:
Also added the `./run-tests` script making it more straightforward

mpscholten (approved):

PR #731, opened 2021-01-30
A minimal DuckDuckGo search bar

kodeFant:
A proposed minimal solution for getting documentation search quickly up and going. 

Takes the user to DuckDuckGo from the search form and then back to IHP again, so maybe not the smoothest solution compared to something like algolia, but works well and gives pretty good search results until we get something better :)


mpscholten (approved):
Perfect, thanks :)

PR #735, opened 2021-02-01
Fix default logger not being registered correctly

zacwood9:
I was not correctly lifting the logger into the ConfigBuilder's State monad, causing an `IO Logger` to be registered instead of a Logger. This would cause a crash if the user did not specify a logger.


mpscholten (approved):
👍

PR #739, opened 2021-02-03
Updated default production log level to "info"

### LINKED ISSUE #738: Request Logger not logging requests anymore
It seems that the request logger is not outputting the request log when running in production anymore. I guess this is related to that default log level being `warn`, and the request logger seems to log to `info`.

cc: @zacwood9

zacwood9:
Closes #738


mpscholten (approved):
Perfect, thanks :)

PR #749, opened 2021-02-10
Add database query time logging

zacwood9:
Added functionality to the query functions in `IHP/ModelSupport.hs` that measure the time each query takes to execute, and includes this in the logs for each query. Example log for a page in Debug mode:

```
Query (28.136ms): "SELECT COUNT(*) FROM (SELECT bands.* FROM bands) AS _count_values" []
Query (2.628ms): "SELECT COUNT(*) FROM (SELECT performances.* FROM performances) AS _count_values" []
GET /
  Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
  Status: 200 OK 0.032103s
```

Also updated the code generation in SchemaCompiler to use the `sqlQuery` function from ModelSupport instead of calling to the Postgres library directly. This way all queries will be logged, not just user created queries.

### DIFF
--- a/IHP/ModelSupport.hs
+++ b/IHP/ModelSupport.hs
@@ -22,6 +22,7 @@ import Data.String.Conversions (cs ,ConvertibleStrings)
 import Data.Time.Clock
 import Data.Time.LocalTime
 import Data.Time.Calendar
+import Data.Time.Format
 import Unsafe.Coerce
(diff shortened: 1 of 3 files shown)


mpscholten (approved):
Thanks :) good addition

PR #750, opened 2021-02-10
Introduce Custom ihp:load Event

s0kil:


fegu:
A IHPteardown-event is just as important to ensure cleaning up JS-resources. This is key to support JS-frameworks without having to actually reload the page entire since morphdom and turbolinks just change content within the same page.

s0kil:
I've added the `ihp:unload` event that will fire on `beforeunload`, and before morphdom patches the page.

mpscholten (approved):

PR #753, opened 2021-02-12
Typed Auto Route

zacwood9:
Currently in IHP AutoRoute is only able to parse one type at a time. These changes add the ability for `AutoRoute` to parse query parameters with the following types:

- Text
- Maybe Text
- Int
- Maybe Int
- Id (for any model type)
- [Text]
- [Int]

Realistically any type that conforms to `Read` could be added to this list.

`typedAutoRoute` is now the default router, with the ability to fallback to the old AutoRoute if necessary for some reason.

Example:

```
data TestController = TestAction { textArg :: !Text, mbTextArg :: !(Maybe Text), listArg :: [Int], idArg :: Id Band } deriving (Eq, Show, Data)
```

```
    action TestAction { .. } = do
        putStrLn textArg
        putStrLn $ show mbTextArg
        putStrLn $ show listArg
        putStrLn $ show idArg
        ...
```

```
GET /Test?textArg=Hello&mbTextArg=Sailor&idArg=ccfcbafc-3660-4c17-b9b1-5299bfe388c5&listArg=1,2,3
```

Output:
```
Hello
Just "Sailor"
[1,2,3]
ccfcbafc-3660-4c17-b9b1-5299bfe388c5
```

See Test/RouterSupportSpec.hs  for a real example.


mpscholten:
Good job 👍 did a first test and so far everything still seems to be working :)

IHP/RouterSupport.hs
```diff
@@ -100,8 +99,145 @@ class HasPath controller => CanRoute controller where
...
+-- does not work. Instead use,
+--
+-- @data MyController = MyAction { textArg :: Text, intArg :: Int }@
+querySortedByFields :: Query -> Constr -> Query
+querySortedByFields query constructor = constrFields constructor
+        |> map cs
+        |> map (\field -> (field, findInQuery field query))
+    where
+        findInQuery _ [] = Nothing
+        findInQuery field ( (k,v) : rest ) = if field == k then v else findInQuery field rest
```
mpscholten, line 185: I think you could use `Data.List.lookup` here

IHP/RouterSupport.hs
```diff
@@ -100,8 +99,145 @@ class HasPath controller => CanRoute controller where
...
+--
+-- this function will receive a representation of the @MyAction@ constructor as well as some query string
+-- @[("textArg", "some text"), ("intArg", "123")]@.
+--
+-- By iterating through the query and attempting to match the type of each constructor argument
+-- with some transformation of the query string, we attempt to call @MyAction@.
+applyConstr :: Data controller => Constr -> Query -> Either TypedAutoRouteError controller
+applyConstr constructor query = let
+
+    -- |Given some query item (key, optional value), try to parse into the current expected type
```
mpscholten, line 207:
```suggestion
    -- | Given some query item (key, optional value), try to parse into the current expected type
```

IHP/RouterSupport.hs
```diff
@@ -100,8 +99,145 @@ class HasPath controller => CanRoute controller where
...
+
+   in case State.runStateT (fromConstrM nextField constructor) (querySortedByFields query constructor) of
+        Right (x, []) -> pure x
+        Right (_) -> Left TooFewArguments
+        Left e -> Left e  -- runtime type error
+
+
 class Data controller => AutoRoute controller where
-    {-# INLINE autoRoute #-}
+    -- | DEPRECATED.
```
mpscholten, line 240: Why is this deprecated? Cannot we just replace the existing `autoRoute` with the new `typedAutoRoute`?

zacwood9: Yeah we can replace it, that works too. Wasn't sure how IHP handles breaking changes. I guess it's a good time since it's still beta!

mpscholten: What breaking changes you have in mind? While looking over the code I got the feeling that for 90% of people nothing will break.

zacwood9: Removing the `parseArgument` functions since they're no longer used would cause compiles to break in all the places where `parseArgument = parseTextArgument` is set in the AutoRoute instances. Other than that everything should work the same

mpscholten: 👍  I think this is worth breaking, it's a price worth paying for progress :)

mpscholten:
We should also update the guide:
- https://ihp.digitallyinduced.com/Guide/routing.html#advanced-custom-parseargument
- https://ihp.digitallyinduced.com/Guide/routing.html#parameter-types

mpscholten (approved):
Really well done :) Thanks a lot 🚀

Issue #726, opened 2021-01-28
libpq: failed (FATAL: sorry, too many clients already )


s0kil:
This is the error message received during development and using autoRefresh.

mpscholten:
Looks like it's related to the internal issue https://linear.app/digitally-induced/issue/CLOUD-36/fix-libpq-failed-fatal-sorry-too-many-clients-already There's also a few steps to reproduce the issue in that ticket.

s0kil:
It seems that the `DevServer` does not clean up connections, since looking at the system resources monitor there are close to 80/90 Postgres processes that are idle, and another 10 or so that are unique. The default connection limit in Postgres is 100, so that is when the error message shows up.

This can be reproduced by making a change, waiting half a second and making a change again to trigger a rebuild, if done 80 or so times, the GC won't have enough time to close all the leftover connections before they reach 80/90 and hit the limit.

Note, it seems the GC has a timeout of about 2 minutes before it cleans up a leftover/idle connection.

mpscholten:
Any updates on this one?

harryaskham:
I'm also facing this. I suspect that @s0kil is correct here as I'm a compulsive saver, meaning I trigger >1 rebuilds in a given second quite frequently.

PR #758, opened 2021-02-15
Stop Database Connections When Application Stopped.

s0kil:
Fixes https://github.com/digitallyinduced/ihp/issues/726

### DIFF
--- a/IHP/AutoRefresh.hs
+++ b/IHP/AutoRefresh.hs
@@ -59,7 +59,7 @@ autoRefresh runAction = do
             -- flash messages, the current user, ...
             --
             -- This frozen context is used as a "template" inside renderView to make a new controller context
-            -- with the exact same content we had when rendering the initial page, whenever we do a serverside rerendering
+            -- with the exact same content we had when rendering the initial page, whenever we do a server-side re-rendering
             frozenControllerContext <- freeze ?context
 
             let renderView = \requestContext -> do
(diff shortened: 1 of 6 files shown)


IHP/AutoRefresh.hs
```diff
@@ -153,16 +152,15 @@ registerNotificationTrigger touchedTablesVar autoRefreshServer = do
...
+                sessions <- (get #sessions) <$> readIORef autoRefreshServer
+                sessions
                     |> filter (\session -> table `Set.member` (get #tables session))
                     |> map (\session -> get #event session)
                     |> mapM (\event -> MVar.tryPutMVar event ())
-            pure ()
-
-
+                pure ())
+    modifyIORef autoRefreshServer (\s -> s { subscriptions = subscriptions })
```
mpscholten, line 162: Shouldn't it append those instead of overriding? Otherwise existing subscriptions will get lost.

```suggestion
    modifyIORef autoRefreshServer (\s -> s { subscriptions = get #subscriptions s <> subscriptions })
```

IHP/Server.hs
```diff
@@ -62,7 +62,9 @@ run configBuilder = do
                                 methodOverridePost $
                                     application
 
-    run `finally` (frameworkConfig |> get #logger |> get #cleanup)
+    run `finally` do
+        frameworkConfig |> get #logger |> get #cleanup
+        readIORef autoRefreshServer >>= (\autoRefreshServer -> autoRefreshServer |> get #subscriptions |> mapM_ Async.uninterruptibleCancel)
```
mpscholten, line 67: Can we extract this into a `stopAutoRefreshServer` function? Otherwise we're kind of leaking implementation details of the auto refresh server in the `IHP.Server` module

s0kil:
Looks good, feels better (;

mpscholten (approved):

PR #762, opened 2021-02-15
Show Indexes In Table View

s0kil:


s0kil:
For now will simply display index in table view. Will open another issue for complete index management.

mpscholten (approved):
Good job, thanks :)

Issue #766, opened 2021-02-17
Segmentation fault (core dumped) When Typed AutoRoute Tries To Parse UUID Value As Text Type


s0kil:

s0kil:
Related, when there is a segfault the Postgres processes are not closed.

mpscholten:
Just happend to me when editing an enum

s0kil:
Turns out this error is related to the new Typed AutoRoute introduced in https://github.com/digitallyinduced/ihp/pull/753 and https://github.com/digitallyinduced/ihp/pull/772

Who would have guessed???

s0kil:
The router crashes the DevServer when trying to parse `EditRow?tableName=events&targetPrimaryKey=ea9cd792-107f-49ff-92a1-f610f7a31f31`, more specifically, when the type of the `targetPrimaryKey` is `Text`, the router crashes parsing `ea9cd792-107f-49ff-92a1-f610f7a31f31`.

s0kil:
Looks like I will be visiting a therapist soon. (:

PR #784, opened 2021-02-25
Fix Issue 766

s0kil:
This test should pass once the bug in the router is fixed.

Fixes https://github.com/digitallyinduced/ihp/issues/766

### DIFF
--- a/IHP/IDE/Data/Controller.hs
+++ b/IHP/IDE/Data/Controller.hs
@@ -19,6 +19,7 @@ import qualified Data.Text as T
 import qualified Data.ByteString.Builder
 
 import qualified Data.ByteString.Char8 as BS
+import qualified Data.UUID as UUID
 
 instance Controller DataController where
     action ShowDatabaseAction = do
@@ -51,7 +52,7 @@ instance Controller DataController where
         connection <- connectToAppDb
         tableNames <- fetchTableNames connection
         primaryKeyFields <- tablePrimaryKeyFields connection tableName
-        let primaryKeyValues = T.splitOn "---" primaryKey
+        let primaryKeyValues = T.splitOn "---" $ UUID.toText primaryKey
         let query = "DELETE FROM " <> tableName <> " WHERE " <> intercalate " AND " ((<> " = ?") <$> primaryKeyFields)
         PG.execute connection (PG.Query . cs $! query) primaryKeyValues
         PG.close connection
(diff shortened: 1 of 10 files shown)


s0kil:
So the issue is that the router tries to first parse as UUID, and then Int, Text, etc.
Since we want a UUID value as Text, this creates a conflict. Instead of messing around with the router, will update the [DevServer to use UUID's instead of Text](https://github.com/digitallyinduced/ihp/blob/master/IHP/IDE/ToolServer/Types.hs#L72)

s0kil:
Maybe it's possible to add an runtime error in the Typed AutoRouter instead of segfaulting?

s0kil:
Tested, Should be good to go!

s0kil:
@zacwood9 Maybe you could weight in on this?

zacwood9:
That's a nasty bug. In general the auto router is all done in an `Either` type that tries to catch all the possible errors that take place, but things get weird with UUIDs.

```haskell
\case
                Just queryValue -> queryValue
                    |> fromASCIIBytes
                    |> \case
                        Just uuid -> uuid |> unsafeCoerce |> Right
                        Nothing -> Left NotMatched
                Nothing -> Left NotMatched,
```

This is different than all the other type checks, which explicitly try to cast to Int, Text, etc because the only use case I had considered was UUIDs wrapped in a polymorphic `Id` newtype, like `Id User`. There's no way I'm aware of to match on those types with `Data.Data`, so instead I just used the same way old AutoRoute handled it, which `unsafeCoerce`s the UUID into the `Id a`.

I don't remember off the top of my head but there's probably some way to safely coerce, which we could then check the result of and return a `Left NotMatched` if the type coercion failed.

zacwood9:
Alternatively we could just try the UUID check last. Then the `Text` type match will trigger and the world will be saved. This is probably better anyway since from some light profiling I did `fromASCIIBytes` is a pretty expensive function, and now will be called for every route

s0kil:
> Alternatively we could just try the UUID check last.

Similar solution that I came up with, so will give it a go.

s0kil:
We could leave the `Data/Controller` types as `UUID`, instead of reverting to `Text`.

mpscholten:
One more tricky case: The primary could also be a non UUID column (e.g. an integer or a string).

Here's some real world example:
```sql
CREATE TABLE continents (
    id CHARACTER(2) PRIMARY KEY NOT NULL,
    name CHARACTER VARYING(510) DEFAULT NULL
);
CREATE TABLE countries (
    id CHARACTER(2) PRIMARY KEY NOT NULL,
    name CHARACTER VARYING(128) NOT NULL,
    full_name CHARACTER VARYING(256),
    iso3 CHARACTER(3),
    iso_number INT,
    continent CHARACTER(2) NOT NULL,
    stripe_tax_rate_id TEXT DEFAULT NULL
);
```

So we might make this unsupported by switching it from text to UUID

mpscholten (approved):

PR #791, opened 2021-02-27
Ignore Global Packages When Running Haddock

s0kil:
Maybe we need a single file with all GHC options. Or maybe GHC supports a variable we could set, something like `export GHC_OPTIONS=''` and all instance of GHC will use it.


mpscholten:
There's `make print-ghc-options` to get a list of all options. This is also used by e.g. `.hie-bios` (for supporting haskell language server) https://github.com/digitallyinduced/ihp/blob/master/lib/IHP/.hie-bios

I think we didn't use it here as getting the output string into the right format (wrapped in `--optghc="`) is complicated and so I just did it manually

Issue #727, opened 2021-01-30
Auto-create index for when adding FK constraint


mpscholten:
The schema designer should automatically add `CREATE INDEX` statements when creating foreign key constraints. E.g. when adding a column `project_id` to some table, the schema designer will automatically wire up the foreign key constraints. It should also put an index on that `project_id` column.

PR #792, opened 2021-02-28
Create Index When Table1 Column References Table2 Column

s0kil:
Implements And Closes https://github.com/digitallyinduced/ihp/issues/727


s0kil:
Anything else I'm missing? otherwise, good to go.

IHP/IDE/SchemaDesigner/Controller/Columns.hs
```diff
@@ -54,6 +54,11 @@ instance Controller ColumnsController where
             let referenceTable = param "referenceTable"
             let onDelete = NoAction
             updateSchema (addForeignKeyConstraint tableName columnName constraintName referenceTable onDelete)
+
+            let indexName = tableName <> "_index"
```
mpscholten, line 58: Creating `users.project_id` will make a index `users_index`. I think this is not unique enough (e.g. when you have two fks, both would be named `users_index`). It would make sense to name it like `users_project_id_index`

mpscholten: Also just saw that `updateSchema` is called twice in the action, that triggers two file changes to `Schema.sql`. This triggers two app rebuilds. We should rather combine the two patch functions using `.` and then call `updateSchema` only once.

mpscholten: I guess while we're on this we could also add some tests for this. I guess for that we might need to extract the domain logic a little bit out of the actions.

mpscholten:
Any updates on this one?

mpscholten (approved):

Issue #770, opened 2021-02-19
Debounce File Changes


s0kil:
When making many file changes, merging a branch with git, the DevServer explodes with many file changes, therefore we should debouce the file changes.

PR #793, opened 2021-02-28
Debounce File Changes

s0kil:
Implements And Closes https://github.com/digitallyinduced/ihp/issues/770

### DIFF
--- a/exe/IHP/IDE/DevServer.hs
+++ b/exe/IHP/IDE/DevServer.hs
@@ -41,10 +41,7 @@ main = do
     when isDebugMode (putStrLn ("IHP Version: " <> Version.ihpVersion))
     setProcessLimits
 
-    let fileWatcherDebounceTime = Clock.secondsToNominalDiffTime 0.1 -- 100ms
-    let fileWatcherConfig = FS.defaultConfig { FS.confDebounce = FS.Debounce fileWatcherDebounceTime }
-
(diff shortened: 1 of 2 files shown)


exe/IHP/IDE/DevServer.hs
```diff
@@ -175,9 +179,10 @@ stop AppState { .. } = do
     stopFileWatcher fileWatcherState
     stopToolServer toolServerState
 
-startFilewatcher :: (?context :: Context) => IO ()
-startFilewatcher = do
-        thread <- async $ FS.withManager $ \manager -> do
+startFileWatcher :: (?context :: Context) => IO ()
+startFileWatcher = do
+        let watchConfig = ?context |> get #fileWatcherConfig
```
mpscholten, line 184: Didn't know that the file watcher library supports this out of the box :) really nice.

Any reasons you didn't move the

```haskell
    let fileWatcherDebounceTime = Clock.secondsToNominalDiffTime 0.1 -- 100ms
    let fileWatcherConfig = FS.defaultConfig { FS.confDebounce = FS.Debounce fileWatcherDebounceTime }
```

directly into the `startFileWatcher`?

This way we can avoid having the `, fileWatcherConfig :: WatchConfig` in the app state, which makes this a bit simpler.

mpscholten (approved):
good job, thanks :)

PR #797, opened 2021-03-01
Use correct pragma (OPTIONS_GHC instead of GHC_OPTIONS)!

mt-caret:
This is pretty embarrassing, but due to a typo in https://github.com/digitallyinduced/ihp/pull/781, my previous PR didn't really work as advertised... This should fix it [by using the correct pragma](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/exts/pragmas.html#options-ghc-pragma).


mpscholten:
Good catch 👍

PR #798, opened 2021-03-01
Remove breaking quotation marks

aramiscd:
The code/markup under "Trying out the login" was broken.

`<a href="{NewSessionAction}">Login</a>` results in a verbatim route `./{NewSessionAction}`.

`<a href={NewSessionAction}>Login</a>` results in the intended route `/NewSession`.


mpscholten (approved):
Good catch :)

PR #802, opened 2021-03-07
remove bad commas from OPTIONS_GHC pragma

mt-caret:
I was debugging IHP for some other reason, when I realized my earlier "fix" (https://github.com/digitallyinduced/ihp/pull/797) didn't fully fix the issue. I've made sure this time that this actually gets the warnings removed...


mpscholten (approved):
👍 Thanks :)

PR #803, opened 2021-03-07
Add  Hoogle(/Haddock) support

mt-caret:
One pain point I've had developing with IHP is that I would often need to google various functions to look up their type signature. Having a Hoogle instance that's built along with IHP will make it much easier for discovering functions and type signatures (especially so since IHP provides a custom Prelude).

I can imagine enabling this by default and putting a link in the dev server UI, but one drawback to this would be increased build time. I imagine cachix alleviates this to some extent, but I don't think there's no getting around `make -B .envrc` taking a bit longer when adding custom Haskell packages. What do the devs think?

### DIFF
--- a/NixSupport/default.nix
+++ b/NixSupport/default.nix
@@ -1,9 +1,13 @@
-{ compiler ? "ghc8103", ihp, haskellDeps ? (p: []), otherDeps ? (p: []), projectPath ? ./. }:
+{ compiler ? "ghc8103", ihp, haskellDeps ? (p: []), otherDeps ? (p: []), projectPath ? ./., withHoogle ? false }:
 
 let
     pkgs = import "${toString projectPath}/Config/nix/nixpkgs-config.nix" { ihp = ihp; };
     ghc = pkgs.haskell.packages.${compiler};
-    allHaskellPackages = ghc.ghcWithPackages (p: builtins.concatLists [ [p.haskell-language-server] (haskellDeps p) ] );
+    allHaskellPackages =
+      (if withHoogle
+      then ghc.ghcWithHoogle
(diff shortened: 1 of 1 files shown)


mpscholten:
When `make -B .envrc` is taking a bit longer this is fine to me as this adds a lot of value when having offline docu :) It's just a one-time operation anyways

mpscholten:
Just tried it out locally, can you point me to where to find / open the generated haddock documentation? :)

mt-caret:
@mpscholten
> Just tried it out locally, can you point me to where to find / open the generated haddock documentation? :)
Sorry about that, if you run `hoogle server --local -p 8080` you'll get a server running at localhost:8080, and from there you can jump to haddock docs.

mt-caret:
> When `make -B .envrc` is taking a bit longer this is fine to me as this adds a lot of value when having offline docu :) It's just a one-time operation anyways

Awesome! In that case, I think we should do the enabling by default part in another PR (potentially also adding a link form the dev server UI to Hoogle/Haddock docs).

mpscholten (approved):
Thanks :)

mpscholten:
> we should do the enabling by default part in another PR (potentially also adding a link form the dev server UI to Hoogle/Haddock docs).

Agree 👍  Especially with the link to the local docs this is nice (Also solves the issue that e.g. the docs at https://ihp.digitallyinduced.com/api-docs/index.html are usually the latest master docs). Can you open an issue for that, so we can keep track of the idea? :)

mt-caret:
@mpscholten Done: https://github.com/digitallyinduced/ihp/issues/804

PR #811, opened 2021-03-11
LoginSupport Helper View

Montmorency:
Previously didn't have LoginSupport View helper equivalent for `currentAdminOrNothing` and `currentAdmin`. This is just boilerplate that would otherwise be added in the application level. Only requires import `CurrentAdminRecord` type family and exporting the helper functions.

Non-breaking change. 

Tested by creating small IHP application with Admin Authentication and expected View rendering on Admin Sessions.


mpscholten (approved):
Perfect, thanks :)

PR #813, opened 2021-03-11
Add support for Integer ID types

zacwood9:
Currently AutoRoute can't handle non-UUID ID types. This introduces the ability for custom ID parsing through Type Application, and changes AutoRoute to be a wrapper around a new function

```haskell
 autoRouteWithIdType :: (?context :: RequestContext, Data idType) => (ByteString -> Maybe idType) -> Parser controller
```

This allows for passing in a function which can convert to another ID type, such as the included `parseIntegerId`

```haskell
parseIntegerId :: (Data idType) => ByteString -> Maybe idType
parseIntegerId queryVal = let
    rawValue :: Maybe Integer = readMay (cs queryVal :: String)
    in
       rawValue >>= Just . unsafeCoerce
```

Users need to manually tell AutoRoute which custom Id type in needs to parse in the instance declaration:

```haskell
instance AutoRoute TestController where
    autoRoute = autoRouteWithIdType (parseIntegerId @(Id ModelType))
```

If no custom types are parsed then this is not needed. This is not a breaking change.

### DIFF
--- a/Guide/routing.markdown
+++ b/Guide/routing.markdown
@@ -136,6 +136,16 @@ pathTo (MyController []) ==> "/Default"
 pathTo (MyController [1,2,3]) ==> "/Default?listParam=1,2,3"
 ```
 
+## For Integer ID types
+
+AutoRoute needs some help if your model does not use UUID as the id type and uses an integer based type instead. To get this to work, add the following to the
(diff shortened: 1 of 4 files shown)


mpscholten (approved):
Thanks 👍

PR #817, opened 2021-03-13
Uncomment background workers

### LINKED ISSUE #816: Job does not run automatically
I think the following parts are affected.

```hs
    -- let run = withBackgroundWorkers frameworkConfig do
    let run = runServer frameworkConfig $
            staticMiddleware $
                sessionMiddleware $
                        requestLoggerMiddleware $
                                methodOverridePost $
                                    application
```
https://github.com/digitallyinduced/ihp/blob/master/IHP/Server.hs#L57-L63

The IHP documentation states:.
> IHP watches the job table in the database for any new records and automatically run the job asynchronously when a new job is added. So to run a job, simply create a new record:

zacwood9:
Fixes #816 

some development code made its way through to master -- I comment out background workers when running in ghci since when running the server in ghci with background workers you're unable to CTRL-C to stop execution for some reason

### DIFF
--- a/Guide/jobs.markdown
+++ b/Guide/jobs.markdown
@@ -38,7 +38,7 @@ instance Job EmailCustomersJob where
     perform EmailCustomersJob { .. } = do
       customers <- query @Customer |> fetch
       forEach customers sendToCustomer
-      where 
+      where
         sendToCustomer customer = sendMail (MarketingMail customer)
 ```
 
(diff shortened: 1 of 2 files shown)


mpscholten:
👍 

Maybe we should also add a warning to the docs that this is currently broken on v0.9?

mpscholten (approved):

Issue #783, opened 2021-02-24
Creating An Enum And Table With The Same Name Throws Error In DevServer


s0kil:
Error:

```
NoData.MonoTraversable.headEx: empty
CallStack (from HasCallStack):
  error, called at src/Data/MonoTraversable.hs:397:28 in mono-traversable-1.0.15.1-AwDeddwIKE6FHgqF6f6LB9:Data.MonoTraversable
 match in record selector values
```

Related: Error when viewing an enum/table with the same name at `ShowEnum?enumName=Stuff` or `ShowTable?tableName=Stuff`
The error is:
```
No match in record selector values
No match in record selector unsafeGetCreateTable
```

Alexander-Block:
I would like to work on this issue if that is okay (I find the ideas behind IHP very interesting and would like to contribute to it and learn more about its inner workings by helping to resolve bugs). 

It seems to me that there are two separate problems behind this issue. The easier one is that it should not be possible to create tables or enum types with a name that is already taken by another table or enum type in the Schema Designer. I think that this can be solved by adding validation logic that is similar to the one I added to ensure uniqueness of enum values with PR #809 . I will prepare a PR for this as soon as possible. 

The errors of the form `No match in record selector <x>`, however, have a more general source. The type `Statement` is defined in `IHP.IDE.SchemaDesigner.Types` as a sum type of records:
```haskell
data Statement
    =
    -- | CREATE TABLE name ( columns );
      StatementCreateTable { unsafeGetCreateTable :: CreateTable }
    -- | CREATE TYPE name AS ENUM ( values );
    | CreateEnumType { name :: Text, values :: [Text] }
    -- | CREATE EXTENSION IF NOT EXISTS "name";
    | CreateExtension { name :: Text, ifNotExists :: Bool }
    -- | ALTER TABLE tableName ADD CONSTRAINT constraintName constraint;
    | AddConstraint { tableName :: Text, constraintName :: Text, constraint :: Constraint }
    | UnknownStatement { raw :: Text }
    | Comment { content :: Text }
    -- | CREATE INDEX indexName ON tableName (columnName);
    | CreateIndex { indexName :: Text, tableName :: Text, columnNames :: [Text] }
    deriving (Eq, Show)
```
A common pattern in the code for the Schema Designer seems to be the following:
```haskell
action UpdateColumnAction = do 
  ...
  let table = findStatementByName tableName statements
  let columns = maybe [] (get #columns . unsafeGetCreateTable) table
[...]

PR #820, opened 2021-03-16
Unique enums and tables

Alexander-Block:
This implements the requirement from Issue #783 that enums and tables with duplicate names should be rejected in EnumsController and TablesController. 
To simplify this change I standardized the validation logic between EnumsController, TablesController and EnumValuesController by defining their validators via `validateNameInSchema`.


IHP/IDE/SchemaDesigner/Controller/Validation.hs
```diff
@@ -0,0 +1,22 @@
+module IHP.IDE.SchemaDesigner.Controller.Validation where
+
+import IHP.ControllerPrelude
+import IHP.IDE.SchemaDesigner.View.Layout (isIllegalKeyword)
```
mpscholten, line 4: Wouldn't it make sense to move the `isIllegalKeyword` from the `Layout.hs` into this file here as well? :) Then the `Layout.hs` could just import it from here.

mpscholten (approved):
Looks good, just tested everything locally and works perfectly :) Thanks!

PR #823, opened 2021-03-18
Link to current nix.trustedUsers docs

das-g:
Link to [current `nix.trustedUsers` documentation](https://search.nixos.org/options?show=nix.trustedUsers&query=nix.trustedUsers) instead of to a hard-coded revision of the source code of it.


mpscholten (approved):
Thanks :)

Issue #412, opened 2020-09-21
Controller Generator mishandles Person/People singular/plural form


pigam:
I created a `people` table:
```sql
-- Your database schema. Use the Schema Designer at http://localhost:8001/ to add some tables.
CREATE TABLE people (
    id UUID DEFAULT uuid_generate_v4() PRIMARY KEY NOT NULL,
    name TEXT NOT NULL,
    forename TEXT NOT NULL,
    date_of_birth DATE NOT NULL
);
```

I then tried to generate the `People` controller.
What I expected : 
- routes named `/people/…`
- generated code with `person` or `people` for identifier's name

What I get:
### Controller
`Web/Controller/People.hs` and `Web/Types.hs` are correct, but `PeoplesAction` should be `PeopleAction`.
```haskell
instance Controller PeopleController where
    action PeoplesAction = do
```
### Views
Generated views are much worse, and use the plural form instead of the singular one. For instance, `Web/View/People/Show.hs`:
```haskell
module Web.View.People.Show where
import Web.View.Prelude

data ShowView = ShowView { people :: People }

instance View ShowView ViewContext where
    html ShowView { .. } = [hsx|
        <nav>
            <ol class="breadcrumb">
                <li class="breadcrumb-item"><a href={PeoplesAction}>Peoples</a></li>
                <li class="breadcrumb-item active">Show People</li>
            </ol>
        </nav>
        <h1>Show People</h1>
    |]
```
This does not compile, as `People` type does not exist.

### Routes
Routes generated by 
```haskell
instance AutoRoute PeopleController
type instance ModelControllerMap WebApplication Person = PeopleController
```
are `/peoples/…`

I hope I'm clear enough in my report.

pigam:
Test after @mpscholten 's suggestion : replacing `PeoplesAction` by `PeopleAction` in `Web/Type.hs`, `Web/Controller/People.hs` and `Web/Views/People/*.hs` fixed generated routes. After this modificaction, routes are of `/People` form as expected.

pigam:
I think https://github.com/digitallyinduced/ihp/blob/master/IHP/IDE/CodeGen/ControllerGenerator.hs#L58
```haskell
        pluralName = Countable.pluralize $ get #controllerName config
```
should be
```haskell
        pluralName = get #controllerName config
```

mpscholten:
I think this could work :) 

[...]

PR #834, opened 2021-03-24
Generator Better Singular/Plural Handling

s0kil:
Added a text case and fixes https://github.com/digitallyinduced/ihp/issues/412

TODO:
- The `pluralize`, `singularize` should also work on uppercased words.


s0kil:
TODO: drop the `countable-inflections` dependency.

mpscholten (approved):

PR #836, opened 2021-03-28
Added type signature for generated index view code

rametta:
I noticed the codegen doesn't add the type sig for the index view like it does for all the other views.
So this PR adds it


mpscholten (approved):
Thanks, good catch 👍

PR #841, opened 2021-03-31
Update docs for "Removing Bootstrap"

rametta:
I think the docs forgot to mention that a few extra lines could be removed when replacing bootstrap with tailwind.
The bootstrap js files and popper dependency can also be removed


s0kil (approved):

s0kil:
Looks good, thanks!

PR #843, opened 2021-04-03
Serverside components

mpscholten:
Demo project with some components: https://github.com/digitallyinduced/ihp-ssc-playground

A simple counter can be implemented like this:
```haskell
module Web.Component.Counter where

import IHP.ViewPrelude
import IHP.ServerSideComponent.Types
import IHP.ServerSideComponent.ControllerFunctions

data Counter = Counter { value :: !Int }

data CounterController
    = IncrementCounterAction
    | SetCounterValue { newValue :: !Int }
    deriving (Eq, Show, Data, Read)

instance Component Counter CounterController where
    initialState = Counter { value = 0 }

    render Counter { value } = [hsx|
        Current: {value} <br />
        <button onclick={callServerAction IncrementCounterAction}>Plus One</button>
        <hr />
        <input type="number" value={inputValue value} onchange="callServerAction('SetCounterValue { newValue = ' + this.value + ' }')"/>
    |]

    action state IncrementCounterAction = do
        state
            |> incrementField #value
            |> pure

    action state SetCounterValue { newValue } = do
        state
            |> set #value newValue
            |> pure

instance SetField "value" Counter Int where setField value' counter = counter { value = value' }
```

Open todos:
- [x] Replace use of `read` with a better encoding
- [x] Use serverside diff
- [x] Keep track of previous state instead of previous html, `htmlDiff (renderHtml(prevState), renderHtml(nextState))`

https://user-images.githubusercontent.com/2072185/113480080-cf5c4980-9492-11eb-9d15-b0baf71577ae.mov

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,7 +32,7 @@ jobs:
           echo "\$file_contents = file_get_contents('./default.nix');" >> replace.php
           echo '$new_content = preg_replace("/\".{40}\"/", "\"$argv[1]\"", $file_contents);' >> replace.php
           echo '$new_content_with_ref = preg_replace("/refs\/tags\/v\d.\d.\d/", "master", $new_content);' >> replace.php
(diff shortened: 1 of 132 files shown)


mpscholten:
just replaced morphdom with a serverside diff approach. Currently the approach how dom nodes are connected to the generated operations set (using `path`) is not reliable enough. So this still needs to be improved.

mpscholten:
Open todos:

- Documentation
- Add some more examples to the playground

PR #848, opened 2021-04-08
Minify Production CSS

rametta:
This doesn't really do real minification like renaming variable names to one letter, but it removes all spaces/tabs/new lines, and is pretty good.
Taken from here: https://mohammadthalif.wordpress.com/2016/07/25/how-to-minify-css-using-simple-bash-commands/


mpscholten (approved):
👍 thanks

Issue #741, opened 2021-02-04
HSX incorrectly removes HTML data attributes with a False value


nghamilton:
HSX removes boolean values when parsing HTML attributes (https://ihp.digitallyinduced.com/Guide/hsx.html#boolean-attribute-values). This leads to unexpected behaviour when setting HTML data attribute values to "false": the data-attribute is not included in the tag.

This fails to include `data-example` in the `div` tag:
```
[hsx| <div data-example={False}></div> |]
```

But this works as expected:
```
[hsx| <div data-example={"false"::Text}></div> |]
```

The documentation should be updated to reflect this or the HSX parsing should operate differently for data attributes?

mpscholten:
Here we mainly try to follow the conventions of the HTML standard as described in https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes

> The presence of a boolean attribute on an element represents the true value, and the absence of the attribute represents the false value.

What is the specific usecase where this behavior is causing you trouble?

fegu:
I believe one of the first bugs I fixed in the IDE had this as the root cause so it should at least be documented.

nghamilton:
I appreciate how HSX parses to clean and conformant HTML, however the boolean-attributes reference that @mpscholten specified above does not mention data attributes, which are what I am referring to.

If you have a look at https://www.w3.org/TR/html51/dom.html#embedding-custom-non-visible-data-with-the-data-attributes there is no mention of special handling of data attributes that contain boolean values. I believe the correct interpretation of the spec is that the value in the data attribute should be parsed as a string. Here is some more indication that this is the case: https://developer.mozilla.org/en-US/docs/Web/API/HTMLOrForeignElement/dataset#setting_values

A good example of incorrect behaviour: `data-turbolinks={False}` will fail to create a data attribute that disables turbolinks for an element.

PR #852, opened 2021-04-10
HSX: boolean data-attributes now use true/false values instead of behaving like HTML boolean attributes

mpscholten:
Fixes https://github.com/digitallyinduced/ihp/issues/741

### DIFF
--- a/Guide/hsx.markdown
+++ b/Guide/hsx.markdown
@@ -70,10 +70,38 @@ Writing `False`:
 <input disabled={False} />
 ```
 
-This will not render the attribute:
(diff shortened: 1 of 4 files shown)


mpscholten:
@nghamilton can you take a look? :)

fegu:
This looks good and is useful. On a related note, the docs should be clearer on how the boolean HTML attribs work, it might surprise people that false means no render. I once fixed a bug in the IDE which had this as root cause. A natural addition to this PR as we already need to update the docs (which is missing here).

mpscholten:
Good point 👍 Just pushed a change to the HSX documentation :)

PR #863, opened 2021-04-24
Fixed turbolink redirect not working as expected after form submission. Fixes #822

### LINKED ISSUE #822: Fix browser navigation forward and backward across form submission pages
I've noticed what seems like an issue with navigation using the browser back and forward buttons and form submission.

**To Reproduce**
1. Submit a form from page A
2. The action redirects to another page B
3. Click the back button in the browser
4. Click the forward button in the browser

**Expected Behavior**

The user ends up on page B.

**Observed Behavior**

The content of page A is displayed, but the URL is B.

**Other notes**

I reproduced in a fresh ihp install with a single controller/model using the default code from the generator.

I dug into what's going on in helpers.js and noticed that if I change [line 185](https://github.com/digitallyinduced/ihp/blob/0a135706417f2550d50b105a63daf5d9b8335bc1/lib/IHP/static/helpers.js#L185) from:
`history.pushState({}, '', request.responseURL);`

to:
`history.pushState({"turbolinks": "fake"}, '', request.responseURL);`

then navigation in this scenario works as expected. But I'm not super familiar with the internals of Turbolinks, so not sure if I'm breaking something else with that change.

mpscholten:
Fixes #822

### DIFF
--- a/lib/IHP/static/helpers.js
+++ b/lib/IHP/static/helpers.js
@@ -187,7 +187,7 @@ window.submitForm = function (form, possibleClickedButton) {
             );
             transitionToNewPage(request.response);
             Turbolinks.clearCache();
-            history.pushState({}, '', request.responseURL);
+            history.pushState({ turbolinks: true }, '', request.responseURL);
             var turbolinkLoadEvent = new CustomEvent('turbolinks:load');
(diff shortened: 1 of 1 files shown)


jyoder:
Thank you!

PR #868, opened 2021-05-05
feat: use frameworkConfig context in Job

err0r500:
a function with a constraint like so : `:: (?frameworkConfig :: FrameworkConfig) => ...` can't be called in Jobs like in Controllers because the constraint in the `perform` function is  

```
perform :: (?modelContext :: ModelContext, ?context :: FrameworkConfig) => job -> IO ()
```
This PR replaces the signature with : 
```
perform :: (?modelContext :: ModelContext, ?frameworkConfig :: FrameworkConfig) => job -> IO ()
```

PS : formatting problem, I didn't find a way to respect yours, sorry about that.

### DIFF
--- a/IHP/Job/Runner.hs
+++ b/IHP/Job/Runner.hs
@@ -28,8 +28,9 @@ runJobWorkers jobWorkers = do
     let oneSecond = 1000000
 
     putStrLn ("Starting worker " <> tshow workerId)
+    let ?frameworkConfig = ?context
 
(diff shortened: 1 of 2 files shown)


s0kil:
IHP has a function [`v:withFrameworkConfig`](https://ihp.digitallyinduced.com/api-docs/IHP-FrameworkConfig.html#v:withFrameworkConfig)
Can be used as:
```haskell
import Config
...
instance Job SendTextMessageJob where
    perform sendTextMessageJob = withFrameworkConfig config \frameworkConfig -> do
       -- Perform Job
```

PR #884, opened 2021-05-21
job timeouts

jeyj0:
Adds an optional timeoutInMicroseconds.

Also automatically sets jobs that are set to "running" on startup to "failed" with a custom message, as that's more appropriate than setting them to time out as we talked about before.

### DIFF
--- a/IHP/Job/Runner.hs
+++ b/IHP/Job/Runner.hs
@@ -10,6 +10,7 @@ import IHP.Prelude
 import IHP.ControllerPrelude
 import IHP.ScriptSupport
 import IHP.Job.Types
+import Control.Monad (void)
 import qualified IHP.Job.Queue as Queue
 import qualified Control.Exception as Exception
 import qualified Database.PostgreSQL.Simple as PG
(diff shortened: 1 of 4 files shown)


IHP/Job/Runner.hs
```diff
@@ -106,6 +108,9 @@ jobWorkerFetchAndRunLoop JobWorkerArgs { .. } = do
     let ?context = frameworkConfig
     let ?modelContext = modelContext
 
+    runningJobs <- Queue.fetchRunningJobs @job
+    mapM_ (Queue.jobDidFail' job "Job was running at startup") runningJobs
```
mpscholten, line 112: What happens if two job workers are started?

jeyj0: Oops, yeah, that shouldn't go there. I had something else there first, which worked with timeouts, where it actually made sense - but now it definitely shouldn't go here, I think. Will take a closer look

jeyj0:
Ready for review again, @mpscholten

mpscholten:
I think this still has the same problem: When a second worker process is started, the existing running jobs are killed

jeyj0:
Wait - but isn't the code only run at startup, and before the workers are actually started?

mpscholten:
Yes but when I start two different processes (e.g. two servers connected to one DB) it will kill the jobs of the other process

jeyj0:
Ooooohh! Yes, that's a case I hadn't thought of. I'll take a look again...

jeyj0:
Separated into two:
1. job timeouts, which has a new PR
2. cancel at startup, which requires some more thought

PR #888, opened 2021-05-23
compile equations for MatchesOp added.

hllizi:
Apparently the compile equations for MatchesOp were missing, so I added them.


mpscholten (approved):
Good catch, thanks :)

Issue #787, opened 2021-02-26
Fix redirectBackWithFallback


s0kil:
Currently, redirects as `redirectToPath`, but the `Referrer` could be a URI instead of a path.

s0kil:
And `redirectBack`

PR #897, opened 2021-05-28
Fix redirectBackWithFallback To Handle URLs and Paths

s0kil:
Fixes https://github.com/digitallyinduced/ihp/issues/787


mpscholten:
When is the problem in #787 happening? Couldn't we make the `redirectBackWithFallback` automatically detect whether it's a path or a url?

mpscholten (approved):

PR #904, opened 2021-06-06
Added new file storage module

mpscholten:
With the file storage module we can easily manage files stored inside S3 and similiar cloud storage services


zacwood9 (approved):
This is awesome! Great docs and nice APIs, IHP at its finest :) looking forward to using this!

Might be nice to see a unified way for managing AWS credentials in the settings. This uses env vars while the SES integration takes the keys as Text. As more integrations get added it might start to be redundant. Just a thought :)

mpscholten:
Thanks! :)

Agree, I want to replace the existing Config.hs approach with env vars before the v1.0 release anyways. So this will be streamlined soon :)

Issue #776, opened 2021-02-22
Rename `make ghci` to `make console`


s0kil:
Make it easier for people not familiar with Haskell ecosystem.

PR #907, opened 2021-06-07
Rename make ghci To make console

s0kil:
Fixes https://github.com/digitallyinduced/ihp/issues/776

### DIFF
--- a/lib/IHP/Makefile.dist
+++ b/lib/IHP/Makefile.dist
@@ -206,5 +206,8 @@ hie.yaml: # Configuration for haskell-language-server
 	echo "  bios:" >> hie.yaml
 	echo "    program: build/ihp-lib/.hie-bios" >> hie.yaml
 
-console: # Start IHP console (ghci) with custom configuration (https://github.com/digitallyinduced/ihp/issues/130)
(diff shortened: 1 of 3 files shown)


lib/IHP/Makefile.dist
```diff
@@ -206,5 +206,5 @@ hie.yaml: # Configuration for haskell-language-server
 	echo "  bios:" >> hie.yaml
 	echo "    program: build/ihp-lib/.hie-bios" >> hie.yaml
 
-ghci: # Start ghci with custom configuration (https://github.com/digitallyinduced/ihp/issues/130)
```
mpscholten, line 209: We should keep `make ghci` as an alias, otherwise we break BC

lib/IHP/Makefile.dist
```diff
@@ -206,5 +206,8 @@ hie.yaml: # Configuration for haskell-language-server
...
+ghci: # Start IHP console (ghci) with custom configuration (https://github.com/digitallyinduced/ihp/issues/130)
 	ghci -package-env -
+
+# https://github.com/digitallyinduced/ihp/issues/776
```
mpscholten, line 212: This text shows in the `make help`, so let's write `Alias for 'make ghci'`

mpscholten (approved):

PR #908, opened 2021-06-07
Introduce withCustomErrorMessageIO

s0kil:
Fixes https://github.com/digitallyinduced/ihp/issues/764

Maybe there is a better way, open to ideas.

### DIFF
--- a/IHP/ValidationSupport/ValidateIsUnique.hs
+++ b/IHP/ValidationSupport/ValidateIsUnique.hs
@@ -1,4 +1,4 @@
-module IHP.ValidationSupport.ValidateIsUnique (validateIsUnique) where
+module IHP.ValidationSupport.ValidateIsUnique where
 
 import IHP.Prelude
 import Database.PostgreSQL.Simple.ToField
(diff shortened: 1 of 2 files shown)


IHP/ValidationSupport/ValidateIsUnique.hs
```diff
@@ -52,3 +52,44 @@ validateIsUnique fieldProxy model = do
...
+
+-- | Overrides the error message of a given IO validator function.
+--
+-- __Example:__ Validate that an email is unique with a custom error message
+--
+-- > action CreateUserAction = do
+-- >     let user = newRecord @User
+-- >     user
+-- >         |> fill @'["email"]
+-- >         |> withCustomErrorMessage' "Email Has Already Been Used" validateIsUnique #email
```
mpscholten, line 65: Can we call this function `withCustomErrorMessageIO`? Also could you add a note to https://ihp.digitallyinduced.com/Guide/validation.html#customizing-error-messages ?

mpscholten (approved):

PR #911, opened 2021-06-10
Case insensitive

mpscholten:
Adds `filterWhereCaseInsensitive`, `validateIsUniqueCaseInsensitive` and switches the IHP built-in login controller to use a case-insensitive email check


mpscholten:
@hllizi as you've been deep into the query builder code just recently maybe could review this one?

PR #912, opened 2021-06-10
Update file-storage.markdown

rametta:
Small typo


mpscholten (commented):
Good catch :) Thanks!

PR #917, opened 2021-06-11
Explicit type cast for UPDATE statements for array types

### LINKED ISSUE #913: `updateRecord` should support explicit casting for postgres
I ran into an issue recently where I was trying to update a field on a record that had the type `[UUID]`, so it was an array of id's being stored in postgres. But when I used `updateRecord` it failed because it was implicitly type [Text].

My original code was:
```hs
item
    |> set #upvotes newUpvotes -- newUpvotes is type [UUID] and so is #upvotes field
    |> updateRecord
```
Which would throw the error: `sqlErrorMsg = "column \"upvotes\" is of type uuid[] but expression is of type text[]` even though my expression was indeed type `uuid[]`.

I had to write manual sql query to add an explicit cast `?::UUID[]` to my update statement.
```hs
result <- sqlQuery "UPDATE public.items SET upvotes = ?::UUID[] WHERE id = ? RETURNING *" (newUpvotesUUID, itemId)
```
[Full sourcecode link](https://github.com/rametta/retros/blob/main/Web/Controller/Items.hs#L36 )

^ So that works, but I would expect IHP to do that automatically with `updateRecord` since it should know the types before executing the query. Or if not automatically, a clean way to do it without writing a raw sql query.

I assume if this change would happen, it would happen here: https://github.com/digitallyinduced/ihp/blob/299e585e0063a5105694687367fc3797d8a9c397/IHP/SchemaCompiler.hs#L422

I can attempt to do this change in a PR, just want to see if I should or not

rametta:
This code was already there for INSERT statement, so I just copied it for UPDATE statements as well :smiley: 
Closes #913

### DIFF
--- a/IHP/SchemaCompiler.hs
+++ b/IHP/SchemaCompiler.hs
@@ -350,6 +350,17 @@ compileToRowValues :: [Text] -> Text
 compileToRowValues bindingValues | length bindingValues == 1 = "Only (" <> (unsafeHead bindingValues) <> ")"
 compileToRowValues bindingValues = "(" <> intercalate ") :. (" (map (\list -> if length list == 1 then "Only (" <> (unsafeHead list) <> ")" else intercalate ", " list) (chunksOf 8 bindingValues)) <> ")"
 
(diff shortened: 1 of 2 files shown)


mpscholten (approved):
Great! Thanks :)

PR #918, opened 2021-06-11
Update contributing troubleshooting section

rametta:
Add tip about what to do if you get the error message `*** WARNING: . is writable by someone else, IGNORING! Suggested fix: execute 'chmod go-w .'`

Related gitter convo: https://gitter.im/digitallyinduced/ihp?at=60c38450c705e53c1c87baef

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -137,6 +137,10 @@ Try to run `make -B build/ihp-lib` to create the symlink.
 
 In the project directory, try `make -B .envrc`
 
+### `*** WARNING: . is writable by someone else, IGNORING! Suggested fix: execute 'chmod go-w .'`
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -137,6 +137,10 @@ Try to run `make -B build/ihp-lib` to create the symlink.
 
 In the project directory, try `make -B .envrc`
 
+### `*** WARNING: . is writable by someone else, IGNORING! Suggested fix: execute 'chmod go-w .'`
```
mpscholten, line 140: The initial error was `error: Could not find moduleGenerated.Types'`, maybe we could use that one here. The chmod error only came up after debugging :)

rametta: True that. Just updated the title and pushed

mpscholten (approved):
Thanks :)

PR #919, opened 2021-06-11
Added missing IsJoined-Constraints

hllizi:
Just a small follow-up to #889: I noticed that, on most of the *JoinedTable functions, I had forgotten to add the constraint `IsJoined model joinRegister` to the context. This reduced type safety by allowing to add filtering by values of columns that belong to tables that have not been joined.


mpscholten (approved):
👍

Issue #920, opened 2021-06-11
Mail Generator Generates In Wrong Application


s0kil:
When selecting `Admin` application, it still generates in `Web`

PR #921, opened 2021-06-11
 Use applicationName When Generating Action And Mail

s0kil:
Fix issue https://github.com/digitallyinduced/ihp/issues/920


mpscholten (approved):
Good catch 👍

PR #922, opened 2021-06-13
Labeled Query Results

hllizi:
I added a function labelResults to the QueryBuilder interface that allows for labelling the results of a query with the values of corresponding values of a joined table. This can be used, for instance, to query all comments and obtain them labeled with the ids of the posts they belong to.

```haskell
labeledComments <-
   query @Comment
      |> innerJoin @Post (#postId, #id)
      |> labelResults @Post #id
      |> fetch
```
This piece of code will provide in labeledComments a list of type [LabeledData (Id' "posts") Comment]. With filters or comprehension, the comments can thus be sorted accordings to the posts they belong to.

This approach would seem to be much more flexible than the current approach using fetchRelated. For instance, it works seamlessly with many-to-one relations:

```haskell
labeledComments <-
   query @Tag
      |> innerJoin @Tagging (#id, #tagId)
      |> innerJoinThirdTable @Post @Tagging (#id, #postId)
      |> labelResults @Post #id
      |> fetch
```
In addition, it reduces the number of SQL request, since the fetchRelated-approach (if I see this correctly) requires making a separate SQL request for each post in order to preserve the association between Post and comment (or tag).

### DIFF
--- a/IHP/QueryBuilder.hs
+++ b/IHP/QueryBuilder.hs
@@ -774,13 +774,15 @@ innerJoin (name, name') queryBuilderProvider = injectQueryBuilder $ JoinQueryBui
 --
 --
 -- __Example:__ Fetch a list of all comments, each paired with the id of the post it belongs to.
--- > postLabeledComments <- query @Comment
--- >                            |> innerJoin @Post (#postId, #id)
--- >                            |> labelResults @Post #id
--- >                            |> fetch
--- > -- SELECT posts.id, comments.* FROM comments INNER JOIN posts ON comments.postId = posts.id 
+-- labeledTags <-
+--   query @Tag
+--      |> innerJoin @Tagging (#id, #tagId)
+--      |> innerJoinThirdTable @Post @Tagging (#id, #postId)
+--      |> labelResults @Post #id
+--      |> fetch
+---- > -- SELECT posts.id, tags.* FROM comments INNER JOIN taggings ON tags.id = taggings.tagId INNER JOIN posts ON posts.id = taggings.postId 
 -- 
--- postLabeledComments is a list of type ['LabeledData' (Id' "posts") Comment] such that "LabeledData postId comment" is contained in that list if "comment" is a comment to the post with id postId.
+-- postLabeledTags is a list of type ['LabeledData' (Id' "posts") Tag] such that "LabeledData postId comment" is contained in that list if "comment" is a comment to the post with id postId.
 --
 labelResults :: forall foreignModel baseModel foreignTable baseTable name value queryBuilderProvider joinRegister.
                 (
--- a/IHP/QueryBuilder.hs
+++ b/IHP/QueryBuilder.hs
@@ -36,6 +36,7 @@ module IHP.QueryBuilder
 , filterWhereILikeJoinedTable
 , filterWhereMatchesJoinedTable
 , filterWhereIMatchesJoinedTable
+, indexResults
 , EqOrIsOperator
 , filterWhereSql
 , FilterPrimaryKey (..)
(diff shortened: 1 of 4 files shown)


IHP/QueryBuilder.hs
```diff
@@ -753,6 +770,31 @@ innerJoin (name, name') queryBuilderProvider = injectQueryBuilder $ JoinQueryBui
...
 
+-- | Index the values from a table with values of a field from a table joined by 'innerJoin' or 'innerJoinThirdTable'. Useful to get, e.g., the comments to a set of posts in such a way that the assignment of comments to posts.
+--
+--
+-- __Example:__ Fetch a list of all comments, each paired with the id of the post it belongs to.
+-- > postLabeledComments <- query @Comment
+-- >                            |> innerJoin @Post (#postId, #id)
+-- >                            |> labelResults @Post #id
+-- >                            |> fetch
+-- > -- SELECT posts.id, comments.* FROM comments INNER JOIN posts ON comments.postId = posts.id
```
mpscholten, line 781: I think this example is not good as it could easily be replaced with:

```haskell
comments <- query @Comment
    |> fetch
```

As every comment has. `postId` field, you can just use `get #postId` instead of `get #labelValue`

Test/QueryBuilderSpec.hs
```diff
@@ -192,6 +192,15 @@ tests = do
                 (toSQL theQuery) `shouldBe` ("SELECT posts.* FROM posts INNER JOIN users ON posts.created_by = users.id INNER JOIN favorite_title ON posts.title = favorite_title.title WHERE users.name != ?", [Escape "Tom"])
 
 
+        describe "labelResults" do
+            it "should provide a query with index field" do
+                let theQuery = query @Post
```
mpscholten, line 197: Same as above. This test case would be simpler when just accessing the `#createdBy` field directly. I think it generally only makes sense in case you have more than one join table.

hllizi:
Thats right of course. I replaced the comment and test with the tag-retrieval example from above.

IHP/QueryBuilder.hs
```diff
@@ -753,6 +770,33 @@ innerJoin (name, name') queryBuilderProvider = injectQueryBuilder $ JoinQueryBui
...
+--
+--
+-- __Example:__ Fetch a list of all comments, each paired with the id of the post it belongs to.
+-- labeledTags <-
+--   query @Tag
+--      |> innerJoin @Tagging (#id, #tagId)
+--      |> innerJoinThirdTable @Post @Tagging (#id, #postId)
+--      |> labelResults @Post #id
+--      |> fetch
+---- > -- SELECT posts.id, tags.* FROM comments INNER JOIN taggings ON tags.id = taggings.tagId INNER JOIN posts ON posts.id = taggings.postId
```
mpscholten, line 783: In this comment the haddock format is a bit broken. Take a look at e.g. https://github.com/digitallyinduced/ihp/blob/master/IHP/AuthSupport/Authorization.hs#L14

After the `__Example:__ Fetch a list of all comments, each paired with the id of the post it belongs to.` we need a new line otherwise haddock will have troubles parsing this.

Also the code samples need a `>` in front :)

mpscholten (commented):
We should also add docs on this one to the guide, best to put it next to the join docs

IHP/QueryBuilder.hs
hllizi: Oops, right. Should be fixed now. Also corrected the content of the comment to match the new example.

mpscholten (approved):

mpscholten:
Good job 👍

PR #925, opened 2021-06-17
ProjectGenerator: Make <nixpkgs> a param.

RosarioPulella:
Turned ProjectGenerator/default.nix into a function by adding a param "pkgs" that defaults to <nixpkgs>
If nothing is passed in this should behave the same, but now the file can be imported  using the nix builtin "import".


mpscholten (approved):
Perfect, thanks :)

PR #928, opened 2021-06-20
Show SQL errors inline instead of separate crash page

rametta:
Two small changes I feel like might be nice:
- The SQL query written in the Ace input now persists between pages and won't reset after hitting `Enter`
- SQL errors are shown inline in a bootstrap alert box (see screenshot below) instead of going to a different page

image

### DIFF
--- a/IHP/IDE/Data/Controller.hs
+++ b/IHP/IDE/Data/Controller.hs
@@ -48,7 +48,7 @@ instance Controller DataController where
         queryResult :: Either PG.SqlError [[DynamicField]] <- if isQuery queryText then
                 (PG.query_ connection query <&> Right) `catch` (pure . Left)
             else
-                PG.execute_ connection query >> pure (Right []) `catch` (pure . Left)
+                (PG.execute_ connection query >> pure (Right [])) `catch` (pure . Left)
 
         PG.close connection
         render ShowQueryView { .. }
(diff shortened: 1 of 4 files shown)


IHP/IDE/Data/Controller.hs
```diff
@@ -41,9 +41,15 @@ instance Controller DataController where
...
-        when (query == "") $ redirectTo ShowDatabaseAction
-        rows :: [[DynamicField]] <- if isQuery query then PG.query_ connection (fromString (cs query)) else PG.execute_ connection (fromString (cs query)) >> return []
+        let queryText = param @Text "query"
+        when (queryText == "") $ redirectTo ShowDatabaseAction
+        let query = fromString $ cs queryText
+
+        queryResult :: Either PG.SqlError [[DynamicField]] <- if isQuery queryText then
+                (PG.query_ connection query <&> Right) `catch` (pure . Left)
+            else
+                PG.execute_ connection query >> pure (Right []) `catch` (pure . Left)
```
rametta, line 51: I have no idea if this is the best way to catch an sql error, I thought it would be nice if it was converted to an `Either` so it's easier to show in the UI, feel free to advise something better if this is bad :slightly_smiling_face:

mpscholten, line 51: I think there's a bug here. When i enter a non-query like `insert into x;`, I still get the haskell error page. For select queries it works.

Likely the `catch` is only matching the `pure (Right [])` expression. If add `()` it could be fixed likely:

```haskell
(PG.execute_ connection query >> pure (Right [])) `catch` (pure . Left)
```

mpscholten: Looks good to me 👍

rametta: Good catch! Fixed

mpscholten (approved):
Good job! :)

PR #933, opened 2021-06-23
add function to delete uploaded files (FileStorage)

jeyj0:


IHP/FileStorage/ControllerFunctions.hs
```diff
@@ -252,9 +253,21 @@ uploadToStorage :: forall (fieldName :: Symbol) context record (tableName :: Sym
     ) => Proxy fieldName -> record -> IO record
 uploadToStorage field record = uploadToStorageWithOptions def field record
 
+-- | Permanently removes a previously stored file from storage.
+removeFile :: (?context :: context, ConfigProvider context) => StoredFile -> IO (Either MinioErr ())
+removeFile StoredFile { path, url } = do
```
mpscholten, line 258: Could we rename this to `removeFileFromStorage`? Otherwise the function will collide with `System.Directory.removeFile` which sometimes is in scope

IHP/FileStorage/ControllerFunctions.hs
```diff
@@ -252,9 +253,21 @@ uploadToStorage :: forall (fieldName :: Symbol) context record (tableName :: Sym
     ) => Proxy fieldName -> record -> IO record
 uploadToStorage field record = uploadToStorageWithOptions def field record
 
+-- | Permanently removes a previously stored file from storage.
```
mpscholten, line 256: Can you add an example here to the haddock?

mpscholten (approved):

Issue #931, opened 2021-06-22
enable CORS policy


mackdunkan:
Customizable CORS policy

`Access to XMLHttpRequest at 'http://localhost:8000/api/Bonuses?siteId=a6e96d3d-cac1-4c68-a492-ffca9b78fe0b' from origin 'http://localhost:4020' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.`

PR #935, opened 2021-06-24
Added support for specifying a CORS policy

mpscholten:
Fixes https://github.com/digitallyinduced/ihp/issues/931

### DIFF
--- a/IHP/Server.hs
+++ b/IHP/Server.hs
@@ -37,6 +37,7 @@ import qualified IHP.Job.Types as Job
 import qualified Control.Concurrent.Async as Async
 import qualified Data.List as List
 import qualified Data.ByteString.Char8 as ByteString
+import qualified Network.Wai.Middleware.Cors as Cors
 
 
 run :: (FrontController RootApplication, Job.Worker RootApplication) => ConfigBuilder -> IO ()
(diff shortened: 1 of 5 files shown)


IHP/Server.hs
```diff
@@ -52,15 +53,17 @@ run configBuilder = do
...
                 staticMiddleware $
-                    sessionMiddleware $
-                            requestLoggerMiddleware $
-                                    methodOverridePost $
-                                        application
+                    corsMiddleware $
+                        sessionMiddleware $
+                                requestLoggerMiddleware $
+                                        methodOverridePost $
+                                            application
```
jeyj0, line 66: maybe we should refactor this into piping now? It's getting continuously longer and harder to read, I think.

```haskell
let run = application
        |> methodOverridePost
        |> requestLoggerMiddleware
        |> sessionMiddleware
        |> ...
```

jeyj0 (approved):

IHP/Server.hs
mpscholten: But then this would display the request-response flow in a wrong order

jeyj0: Ah, that's an angle I hadn't thought about yet. I always imagine middleware as a wrapper, and see it as layers added. But this way of thinking makes the current way of writing it clear. Refactored my thought-process, not the code: great! :D

Issue #286, opened 2020-08-01
point and bytea, and tsvector


mpscholten:
Add support for the above types

PR #938, opened 2021-06-24
Postgres tsvector Support

s0kil:
Closes https://github.com/digitallyinduced/ihp/issues/286
Supersedes https://github.com/digitallyinduced/ihp/pull/344

### DIFF
--- a/IHP/Postgres/TSVector.hs
+++ b/IHP/Postgres/TSVector.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE TemplateHaskell #-}
 {-|
 Module: IHP.Postgres.TSVector
 Description: Adds support for the Postgres tsvector type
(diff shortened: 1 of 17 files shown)


IHP/Postgres/TSVector.hs
```diff
@@ -0,0 +1,13 @@
...
+Copyright: (c) digitally induced GmbH, 2021
+-}
+module IHP.Postgres.TSVector where
+
+import BasicPrelude
+
+-- | Represents a Postgres tsvector
+--
+-- See https://www.postgresql.org/docs/current/datatype-textsearch.html
+type TSVector = Text
```
mpscholten, line 13: This needs to be a newtype

mpscholten: I don't think a string representation is the right choice here.

This is a valid tsvector: `a:1A fat:2B,4C cat:5D`

The `:` suffix has special meaning here. A good structure would be:

```haskell
data TSVector = TSVector [Lexeme]
data Lexeme = Lexeme { token :: Text, position :: Int, weight :: Text }
```

s0kil:
@mpscholten It is good to go.

IHP/Postgres/TSVector.hs
```diff
@@ -0,0 +1,65 @@
...
+
+                position <- double
+                -- The Default Weight Is `D` So Postgres Does Not Include It In The Result
+                weight <- option 'D' $ choice [char 'A', char 'B', char 'C', char 'D']
+                pure $ LexemeRanking { position = truncate position, weight }
+
+            pure $ Lexeme { token = cs token, ranking }
+
+instance ToField TSVector where
+    toField = undefined
```
mpscholten, line 65: When we have a FromField implementation maybe we could also have a ToFiedl?

mpscholten (commented):
Could you also add some tests?

s0kil:
I've added the tests.

mpscholten (approved):

PR #939, opened 2021-06-25
Use query parameters to transmit form values where the form is submitted via GET

mpscholten:


CSchank:
Ouuu actually, I think I want to use this in the pagination filtering (which I'm about ready to submit as a WIP PR). Right now it's kind of a confusing hybrid where the text box submits as a POST request but then the page numbers along the bottom have to manually put it in as a GET when the user wants to change pages. If instead it was always in the URL from the start that would make more sense.

mpscholten:
Good point, the initial problem reported via gitter was basically a similiar problem (search query input field)

CSchank:
Great, now that this is merged I will update my PR against the new version.

philschonholzer:
Thanks @mpscholten! I tested this change as you described, by placing this `helper.js` in the static folder. Now my search query behaves just as I want it to.

Issue #898, opened 2021-06-01
Convenience Functions for Pagination


joshua-obritsch:
I wanted to make a suggestion for a couple of convenience functions for pagination. The first one I made is `fetchPageCount` which fetches the total number of pages for a query. It currently looks like this:

```haskell
fetchPageCount                                                                                                                  
    :: forall (table :: Symbol)                                                                                                 
    .  ( ?modelContext :: ModelContext                                                                                          
       , KnownSymbol table                                                                                                      
       )                                                                                                                        
    => QueryBuilder table                                                                                                       
    -> IO Int                                                                                                                   
fetchPageCount !queryBuilder = do                                                                                               
    let !(theQuery', theParameters) = toSQL' (buildQuery queryBuilder)                                                          
    let theQuery = "SELECT COUNT(*) FROM (" <> theQuery' <> ") AS _count_values"                                                
    trackTableRead (symbolToByteString @table)                                                                                  
    [Only count] <- sqlQuery (Query $! cs theQuery) theParameters :: IO [Only Int]                                              
    pure $ ceiling $ fromIntegral count / fromIntegral pageSize
```
This is useful for displaying how many pages are remaining, e.g. 2/5. Ideally, you would want to pass `pageSize` in as an argument. I'm just using a constant for now. Here is how I'm using it in one of my controllers:

```haskell
pageCount <- query @Bundle |> fetchPageCount
```

The second function I made was `fetchPage` and looks like this:

```haskell
fetchPage                                                                                                                       
    :: forall model (table :: Symbol)                                                                                           
    .  ( ?modelContext :: ModelContext                                                                                          
       , FromRow model                                                                                                          
       , KnownSymbol (GetTableName model)                                                                                       
       , KnownSymbol table                                                                                                      
       )                                                                                                                        
    => Int                                                                                                                      
    -> QueryBuilder table                                                                                                       
    -> IO [model]                                                                                                               
fetchPage page !queryBuilder = do                                                                                               
    let !(theQuery, theParameters) = queryBuilder                                                                               
            |> buildQuery                                                                                                       
            |> setJust #offsetClause ("OFFSET " <> offset)                                                                      
            |> setJust #limitClause ("LIMIT " <> limit)                                                                         
            |> toSQL'                                                                                                           
    trackTableRead (tableNameByteString @model)                                                                                 
    sqlQuery (Query $ cs theQuery) theParameters                                                                                
  where                                                                                                                         
[...]

PR #940, opened 2021-06-25
Add pagination to index pages

CSchank:
This is a PR per the discussion in #898. See that document for a bit more explanation of the design of this PR.

This PR adds pagination helper functions for both views and actions, in a way that is intended to be unobtrusive to a bigger overall pipeline in the action (e.g., using `fetchRelated`). It also modifies the index pages and actions to include the code needed for pagination by default. 😊 Here is an example of how it looks on the page:

image

I'm not certain I like the << and >> vs. "Previous" and "Next". Except that it makes the buttons the same size, which I think is nicer.

It also includes the ability to filter results by a given query:

image
(sidenote: Users with roles?! Ouuu shiny!! Future PR maybe?)

Currently I still consider this a bit of a WIP and would like some feedback. The code could probably be cleaned up, and things like the defaults might need to change. Defaults are currently:

- 100 items per page (maybe too high?)
- A maximum "window size" (the radius of page links shown in the first screenshot) of 5.
- Not to include the filter box in the index view by default, since it's hard to know if their data type has a `Text` thing that can be filtered. I wonder if a non-working version should be included in the view, but I don't think this seems in line with the "batteries included" philosophy of IHP, so maybe we just need a guide of how to get that working (the documentation already shows examples).

Any other feedback would be appreciated. I am realizing I also didn't run the code through any kind of formatter; I didn't see one mentioned in the Contributing docs.

Also, should I write tests?


IHP/Pagination/ViewFunctions.hs
```diff
@@ -0,0 +1,136 @@
...
+                    <a class="page-link" href={pageUrl $ currentPage + 1} aria-label="Previous">
+                        <span aria-hidden="true">&raquo;</span>
+                        <span class="sr-only">Next</span>
+                    </a>
+                </li>
+            </ul>
+        </nav>
+    |]
+        where
+            nextClass = "page-item" ++ (if hasNextPage pagination then "" else " disabled") :: Text
```
mpscholten, line 44: IHP has a `classes` helper for this :)

```suggestion
            nextClass = classes ["page-item", (hasNextPage pagination, "disabled")]
```

mpscholten (commented):
Looks good so far 👍 

Maybe you could also add some tests :)

mpscholten:
Any updates @CSchank ?

CSchank:
@mpscholten sorry, our research work has taken me away from the project using IHP for a little bit. I'll try to get it rolling very soon.

IHP/Pagination/ViewFunctions.hs
CSchank: Forgive my potential ignorance, is the `(Bool,a)` syntax for including an element in a list correct? Is it provided by a language extension?

IHP/Pagination/ControllerFunctions.hs
```diff
@@ -0,0 +1,176 @@
...
+-- | Customize the ``window size'' in an exsiting 'Options', which 
+--   is the radius of the pages shown on the navbar. Can be used as 
+--   part of 'paginateOptions':
+--
+-- > (userQ, pagination) <- query @User 
+-- >        |> orderBy #email
+-- >        |> paginateOptions 
+-- >            (defaultOptions |> setWindowSize 3)
+setWindowSize :: Int -> Options -> Options
+setWindowSize n options =
```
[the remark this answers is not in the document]
CSchank: This is a good idea, and I should probably rename the field if people are using it directly.

IHP/Pagination/ControllerFunctions.hs
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+-- | Customize the ``window size'' in an exsiting 'Options', which 
+--   is the radius of the pages shown on the navbar. Can be used as 
+--   part of 'paginateOptions':
+--
+-- > (userQ, pagination) <- query @User 
+-- >        |> orderBy #email
+-- >        |> paginateOptions 
+-- >            (defaultOptions |> setWindowSize 3)
```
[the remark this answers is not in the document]
CSchank: Good catch, thanks!

IHP/Pagination/ControllerFunctions.hs
```diff
@@ -0,0 +1,176 @@
...
+--
+-- Example:
+--
+-- > action UsersAction = do
+-- >    (userQ, pagination) <- query @User 
+-- >        |> orderBy #email
+-- >        |> paginateOptions (itemsPerPage 5)
+-- >    user <- userQ |> fetch
+-- >    render IndexView { .. }
+paginateOptions :: (?context::ControllerContext, ?modelContext :: ModelContext, ?theAction :: controller, KnownSymbol q) =>
```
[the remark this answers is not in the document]
CSchank: That sounds good! Thanks.

IHP/Pagination/ControllerFunctions.hs
```diff
@@ -0,0 +1,176 @@
+{-|
+Module: IHP.PageTitle.ControllerFunctions
```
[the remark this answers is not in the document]
CSchank: Thanks!! You caught me copy+pasting ;)

IHP/IDE/CodeGen/ViewGenerator.hs
```diff
@@ -171,6 +171,7 @@ buildPlan' schema config =
                 <> "                </thead>\n"
                 <> "                <tbody>{forEach " <> pluralVariableName <> " render" <> singularName <> "}</tbody>\n"
                 <> "            </table>\n"
+                <> "            { renderPagination pagination }\n"
```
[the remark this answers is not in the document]
CSchank: Sounds good, thanks. This is a habit I have developed from using languages like CalcCheck (https://calccheck.mcmaster.ca) where spaces are needed to denote different symbols, since they have full Unicode support for all symbols in the language (proof by emoji is my personal favourite 😉)

IHP/Pagination/ViewFunctions.hs
mpscholten: [This is provided by the `IHP.View.Classes` module](https://ihp.digitallyinduced.com/api-docs/IHP-View-Classes.html) :) The implementation is relatively short:

```haskell
classes :: [(Text, Bool)] -> Text
classes !classNameBoolPairs =
    classNameBoolPairs
    |> filter snd
    |> map fst
    |> unwords
```

It's initially inspired by react's classnames helper, which is now replaced by https://www.npmjs.com/package/classnames

CSchank: Oh, okay, I see. But the list above still confuses me because it appears to have elements of different types. Perhaps just a typo? Would it be something like `[("page-item", True), ("disabled", hasNextPage pagination)]`?

mpscholten: Got it :) The `"page-item"` is a trick. We have a `IsString` instance for this in `IHP.View.Classes`.

The `["page-item", ("disabled", hasNextPage pagination)]` is rewritten to `[fromString "page-item", (fromString "disabled", hasNextPage pagination)]` because of the GHC `OverloadedStrings` extension.

Then we have a special `IsString` instance in `IHP.View.Classes`:

```haskell
instance IsString (Text, Bool) where
    fromString string = (cs string, True)
```

Now if you substitute the `fromString "page-item"` with the `(cs "page-item", True)` you have matching types :)

TLDR: `"page-item"` can be the short form of `("page-item", True)` if the `IsString (Text, Bool)` type instance  is used.

CSchank: Ahh yes `IsString` and `OverloadedStrings`!!! Of course!!!!! I should have known. Very clever 😃 You guys really make clever use of GHC. 😊 (I know this one isn't particularly fancy, but there are lots of other clever ways you use extensions too!)

Very cool, thanks.

CSchank: Sidenote, I sometimes find it a bit hard to know which extensions are enabled since you hide that from the source file itself. Which I agree with for most (non-Haskell) users since you don't want to overwhelm them with details. But it makes it hard to track down these sorts of things and what actually enables them.

mpscholten: :) agree, we could add a list of all enabled language extensions to [`Helpful Tips & Special Haskell Syntax`](https://ihp.digitallyinduced.com/Guide/helpful-tips.html). Do you think this would improve the situation?

CSchank: Frankly, I had not found that particular page! We could do that, yes. But I like how that page gives you the syntax and then tells you what enables it (at least for some of them). Either way, it's not like it really affected my experience at the beginning. I just decided to take certain things for granted as I was starting, as if it was a brand new language. But as I got more curious I wanted to know how things actually worked under the hood.

IHP/Pagination/ControllerFunctions.hs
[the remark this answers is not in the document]
CSchank: For this, I need an instance like 

```haskell
instance SetField "windowSize" Options Int where
    setField value options =
        options { windowSize = value }
```

in the `Types.hs`. Is there any way to derive this automatically? Seems silly to have to write it for each field.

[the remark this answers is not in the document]
mpscholten: Currently not :/ I think this is coming to GHC in future versions, our `SetField` is kind of a polyfill until this is natively implemented

CSchank:
@mpscholten I have implemented your fixes and also a way for the user to select how many items to view. Last thing left is to go tests.

image

mpscholten:
Looks really already, good job! :)

mpscholten:
I think we can already merge this now :) Tests can be added in a seperate PR

PR #942, opened 2021-06-27
Add path to redirect to after user is created

GarrisonJ:
I'm not sure what's best to put here. But the action as it is will cause a runtime error.


mpscholten (approved):
Thanks 👍

PR #951, opened 2021-07-03
replace fetch by fetchOneOrNothing in IHP/Guide/recipes.markdown

zmijunkie:
fetchOneOrNothing is the better choice here


mpscholten (approved):
Thanks :)

Issue #950, opened 2021-07-01
Undocumented usecase on collectionFetchRelated?


kodeFant:
Hey. I had a go at `collectionFetchRelated` where the examples shows how to fetch comments from many `posts` that could have many `comments` and misunderstood the docs a bit.

https://ihp.digitallyinduced.com/Guide/relationships.html#multiple-records

Only thing was that I misread and tried to do it the oposite way, you could say I used it for fetching many `comment` that  each could have one `postId` because I want to list this datatype out where I could see the title of the post for each entry.

I first tried out something equivalent to this, mapping it out identically to what what done in the docs and it failed because of an ambiguity problem which led me to confusion. I went to this section because I wanted to prefetch this data for multiple comments.

```haskell
getCommentCollections :: (?modelContext::ModelContext) => IO [Include "posts" Comment]
getCommentCollections =  query @Comment
                                |> fetch 
                                >>= collectionFetchRelated #posts
```

In retrospect I clearly see that this does not make sense after getting some help from this [Slack thread](https://ihpframework.slack.com/archives/C01DQE0F4F8/p1625150078184300) with help from @s0kil . The relationship goes the other way.

But it appears it also works for my usecase, where I query many comments, and want its one parent post, so I would propose that showcasing this usecase could be useful, and maybe could avoid future confusion? I think mapping out both cases could be making it clearer.

By doing this, I am able to pre-fetch the post data.

```haskell
getCommentCollections :: (?modelContext::ModelContext) => IO [Include "postId" Comment]
getCommentCollections =  query @Comment
                                |> fetch 
                                >>= collectionFetchRelated #postId
```

Then in the listing I could of course map it down.
```
{forEach commentCollections renderComment}
```
And fetch the data from the parent post like this.
```
{get #postId transaction |> get #title}
```
And the post title will appear, just like how the section about the `Belongs To` relationship works, but with multiple posts.

mpscholten:
I pushed a PR that improves the docs there. Can you take a look? :)

kodeFant:
This looks great 👍

PR #952, opened 2021-07-03
Added docs on fetching all Comments with their Posts (Many-to-one)

mpscholten:
Fixes https://github.com/digitallyinduced/ihp/issues/950


kodeFant:
Very good, I think the usage is presented very clearly :1st_place_medal:

Issue #961, opened 2021-07-06
`CREATE UNIQUE INDEX` not supported


hendi:
`CREATE UNIQUE INDEX unique_source_id ON listings (source, source_id) WHERE source IS NOT NULL AND source_id IS NOT NULL;`

fails with

```
Application/Schema.sql:105:8:
    |
105 | CREATE UNIQUE INDEX unique_source_id ON listings (source, source_id) WHERE source IS NOT NULL AND source_id IS NOT NULL;
    |        ^^^^^^^^^
unexpected "UNIQUE IN"
expecting "EXTENSION", "FUNCTION", "INDEX", "OR", "TABLE", "TRIGGER", or "TYPE"
```

PR #962, opened 2021-07-06
Support CREATE UNIQUE INDEX

s0kil:
Fixes https://github.com/digitallyinduced/ihp/issues/961
https://www.postgresql.org/docs/current/indexes-unique.html

### DIFF
--- a/IHP/IDE/SchemaDesigner/View/Layout.hs
+++ b/IHP/IDE/SchemaDesigner/View/Layout.hs
@@ -176,13 +176,15 @@ renderColumnIndexes tableName statements = forEach (findTableIndexes statements 
         renderIndex index = [hsx|
             <tr class="index">
                 <td class="index-name">{get #indexName index}</td>
+                {unique}
                 <td class="index-expressions">{expressions}</td>
             </tr>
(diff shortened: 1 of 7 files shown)


IHP/IDE/SchemaDesigner/View/Layout.hs
```diff
@@ -176,13 +176,15 @@ renderColumnIndexes tableName statements = forEach (findTableIndexes statements
         renderIndex index = [hsx|
             <tr class="index">
                 <td class="index-name">{get #indexName index}</td>
+                {unique}
                 <td class="index-expressions">{expressions}</td>
             </tr>
         |]
             where
+                unique = if get #unique index then [hsx|<td class="index-expressions">UNIQUE</td>|] else mempty
```
mpscholten, line 184: Using when is shorter here :)

```suggestion
                unique = when (get #unique index) [hsx|<td class="index-expressions">UNIQUE</td>|]
```

mpscholten (commented):
Looks good 👍  one small thing

mpscholten (approved):

PR #970, opened 2021-07-12
Show icons in Schema Editor, fix #955

### LINKED ISSUE #955: Show icons to discern enums, tables etc.
Should probably use the familiar icons from pgadmin
image

hendi:
This fixes #955 and adds some icons from the pgadmin4 project to the schema editor:

image

I originally added some code to order the DDL statements, e.g. first tables, then indices, then enums etc. But I removed that change since I usually have my DDL statements in an order that makes sense to me (regarding ordering, also have a look at #584). That's also why I changed the code to include comments as well. Thoughs?

### DIFF
--- a/IHP/IDE/SchemaDesigner/View/Layout.hs
+++ b/IHP/IDE/SchemaDesigner/View/Layout.hs
@@ -3,7 +3,7 @@ module IHP.IDE.SchemaDesigner.View.Layout (schemaDesignerLayout, findStatementBy
 import IHP.ViewPrelude
 import IHP.IDE.SchemaDesigner.Types
 import IHP.IDE.ToolServer.Types
-import IHP.IDE.ToolServer.Layout
+import IHP.IDE.ToolServer.Layout hiding (tableIcon)
 import IHP.IDE.SchemaDesigner.Compiler (compileIdentifier, compilePostgresType, compileExpression)
 import qualified IHP.IDE.SchemaDesigner.Parser as Parser
 import qualified Text.Megaparsec as Megaparsec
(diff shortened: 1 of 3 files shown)


IHP/IDE/SchemaDesigner/View/Layout.hs
```diff
@@ -287,17 +287,67 @@ renderObjectSelector statements activeObjectName = [hsx|
         |]
             where
                 contextMenuId = "context-menu-" <> tshow id
-        renderObject Comment {} id = mempty
```
mpscholten, line 290: Why have these empty statements been removed? The idea was that we hide all the statements that the user cannot interact with from the GUI

hendi: My train of thought was: icons look nice -> oh, we only show tables and enums -> let's show more (all) to see more pretty icons

mpscholten: I agree, but in big project's this really wastes a huge amount of screen space (I've tested it)

hendi: We could sort the entries by type (e.g. tables, types, indexes), add a sub-header and by default collapse everything except tables. That's also how pgadmin and dbeaver work.

(FWIW, I'm also fine with simply hiding everything except tables, types and indexes)

mpscholten: This is all too complicated for things that cannot be used from the Schema Designer. Let's keep the behaviour from master and hide everything which cannot be interacted with

mpscholten (approved):
Thanks :) Good job

Issue #979, opened 2021-07-18
Means to display time only


joshua-obritsch:
Currently, there exist only three functions in `IHP.View.TimeAgo`, namely `timeAgo`, `dateTime` and `date`, none of which display only the time. It would be nice to add a function that displays only the time.

PR #980, opened 2021-07-18
Implement time function in TimeAgo

joshua-obritsch:
I wanted to add a function that displays only the time. This addresses issue #979.

Here's an example of how it looks:

2021-07-18-170019_5760x2160_scrot

I'm not sure about the naming of the function `time`. This is probably too generic, right?


mpscholten:
Thanks 👍 Happy with the name `time`, hasn't caused a conflict for me. In case this will cause problems in the future, we can still rename the function to something else later on

mpscholten (approved):

PR #981, opened 2021-07-19
Fix width and height of enumIcon

joshua-obritsch:
I noticed a bug in the SchemaDesigner with the latest commit. The width and height of the enumIcon are not set.

Here is a picture of the bug:

bug

Here is a picture of the fix:

fix


mpscholten:
This is mostly a cache issue. In the CSS it has been changed here: https://github.com/digitallyinduced/ihp/pull/970/files#diff-e270180674b34f801920ece24f7a0c0a5be4e6e5d4e5ff3484aa6fccf21aa2f3R42

Maybe we could then remove the CSS rule and just apply the `24px` inline here?

mpscholten (approved):
Thanks!

PR #985, opened 2021-07-21
Update authentication guide

GarrisonJ:
This didn't compile for me without the parameter `user`. Also, `#isConfirmed` with a capital C didn't compile for me.

### DIFF
--- a/Guide/authentication.markdown
+++ b/Guide/authentication.markdown
@@ -228,7 +228,7 @@ To block login (requires `isConfirmed`boolean field in `Users` table):
 ```haskell
 instance Sessions.SessionsControllerConfig User where
     beforeLogin user = do
-        unless (get #isconfirmed user) do
+        unless (get #isConfirmed user) do
             setErrorMessage "Please click the confirmation link we sent to your email before you can use IHP Cloud"
             redirectTo NewSessionAction
 ```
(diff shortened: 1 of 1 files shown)


mpscholten:
Regarding the `#isconfirmed`: Can you share the `CREATE TABLE` statement of your users table?

GarrisonJ:
Sure!

```sql
CREATE TABLE users (
    email TEXT NOT NULL,
    password_hash TEXT NOT NULL,
    locked_at TIMESTAMP WITH TIME ZONE DEFAULT NULL,
    failed_login_attempts INT DEFAULT 0 NOT NULL,
    timezone TEXT NOT NULL,
    username TEXT NOT NULL UNIQUE,
    isConfirmed BOOLEAN DEFAULT false NOT NULL
);
```

mpscholten:
The column name should `is_confirmed` instead of `isConfirmed` :) The reason is that postgres only has lowercased column names, so `isConfirmed` is basically `isconfirmed` for postgres.

GarrisonJ:
Oh, that makes sense. Thanks!

mpscholten:
Could you update the PR to only include the missing user parameter? Then we can merge it :)

mpscholten (approved):
Perfect, thanks :)

PR #987, opened 2021-07-21
link slack in welcome view

jeyj0:


s0kil (dismissed):

mpscholten:
This view is not actually used anymore as it's now generated as part of the project generation. See https://github.com/digitallyinduced/ihp/blob/master/IHP/IDE/CodeGen/ApplicationGenerator.hs#L161 This was done to make it easier to play with live reload initially when setting up a project.

What tests did you run before doing a PR @jeyj0 ? What tests were done when approving this PR @s0kil ?

s0kil:
@mpscholten It was a mistake on my end.

mpscholten (approved):

PR #991, opened 2021-07-21
Small fix in the JS example in views doc

newtonbeck:
This MR introduces a small fix in the views documentation. The example code was mixing `async/await` and `then` method chaining. 

The MR changes the example to use only the `async/await` approach for readability improvement.


s0kil (approved):

s0kil:
Looks good, thanks!

Issue #998, opened 2021-07-23
error screen: links style makes them hard to read


CSchank:
I am guessing these are meant to have newlines or at least spaces between them. Might be a regression.

![image](https://user-images.githubusercontent.com/12722104/126812264-ea55ab02-8973-49f4-9356-1b1c436f8919.png)

mpscholten:
I guess this only happens when you make a change that causes the compiling screen to be shown (e.g. a type or syntax error). Then the live reload will refresh the page, get's the error message and morphdoms to the new page. But it will not morphdom the `<head>` (only the body), so the css is not loaded as expected.

mpscholten:
E.g. here's how this looks for me when I first start the app successfully and then insert a syntax error:

![image](https://user-images.githubusercontent.com/2072185/126860919-69fa4342-619d-44d7-9ff1-654d581c735b.png)

When I now do full page refresh, this is how it looks:

![image](https://user-images.githubusercontent.com/2072185/126860928-bbb996e9-f5ba-4a61-a133-1f64bf03f437.png)

CSchank:
Yes, that's how the page looks if I do a full refresh. So, this is a morphdom problem? Is this related to my question on Gitter about how it interacts with other JS things, like Elm? It definitely seems a bit glitchy in certain edge cases.

mpscholten:
Fixed 👍

PR #1000, opened 2021-07-24
Livereload morphdom Patch Head

s0kil:
Fix issue https://github.com/digitallyinduced/ihp/issues/998

### DIFF
--- a/lib/IHP/static/livereload.js
+++ b/lib/IHP/static/livereload.js
@@ -27,6 +27,10 @@ function refresh() {
         .then(html => {
             var parser = new DOMParser();
             var dom = parser.parseFromString(html, 'text/html');
+
+            // https://github.com/digitallyinduced/ihp/issues/998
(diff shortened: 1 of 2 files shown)


lib/IHP/static/livereload.js
```diff
@@ -27,6 +27,10 @@ function refresh() {
         .then(html => {
             var parser = new DOMParser();
             var dom = parser.parseFromString(html, 'text/html');
+
+            // https://github.com/digitallyinduced/ihp/issues/998
+            morphdom(document.head, dom.head)
+
```
CSchank2, line 33: I wonder if this will help us with our problem of Elm not interacting well with the livereload.

CSchank, line 33: I wonder if this will help us with our problem of Elm not interacting well with the livereload.

mpscholten (approved):
Works :) Good job!

Issue #978, opened 2021-07-18
Issue when editing nullable columns in development tooling


joshua-obritsch:
I've noticed an issue when using the data section in the development tooling at localhost:8001.

When editing nullable columns, the field is prefilled with an empty string and throws the following error on submitting the form without manually setting all fields to NULL:

```
SqlError {sqlState = "22007", sqlExecStatus = FatalError, sqlErrorMsg = "invalid input syntax for type timestamp with time zone: \"\"", sqlErrorDetail = "", sqlErrorHint = ""}
An exception was raised while running the action UpdateRowAction
```

PR #1003, opened 2021-07-25
EditRowView Render NULL In SQL Mode

s0kil:
Fixes https://github.com/digitallyinduced/ihp/issues/978


IHP/IDE/Data/View/EditRow.hs
```diff
@@ -171,6 +171,38 @@ instance View EditRowView where
...
+                                            />
+                                        <label class="form-check-label" for={get #columnName def <> "-sqlbox"}> Parse as SQL</label>
+                                    </a>
+                                    <input
+                                        type="hidden"
+                                        name={get #columnName def <> "_"}
+                                        value={inputValue False}
+                                        />
+                                </div>
+                            </div>|]
```
mpscholten, line 205: I think if it only changes the `value` attribute, we can archive this with a lot less duplicate code. Right now it's like 90% the same as the code below. This will make it harder to fix bugs in the future as we then have to change it at multiple places.

s0kil: It not only changes the value attribute, but also the dropdown-items.

PR #1004, opened 2021-07-25
Add multiple Attribute To FormField

s0kil:
Closes https://github.com/digitallyinduced/ihp/issues/893

### DIFF
--- a/Guide/form.markdown
+++ b/Guide/form.markdown
@@ -269,6 +269,31 @@ This will render like:
 </div>
 ```
 
+### Multiple
+
+You can mark an input with multiple, to ensure it will aceept multiple values, like this:
+
+```haskell
+{(textField #email) { multiple = True } }
+```
+
+This will render like:
+
+```html
+<div class="form-group" id="form-group-post_email">
+    <label for="post_email">Email</label>
+
+    <input
+        type="email"
+        name="email"
+        id="post_email"
+        class="form-control"
+        multiple="multiple"
+    />
+</div>
+```
+
+Note: `multiple` attribute is only valid for email input, file input, and select
 
 ### Custom Submit Button Text
 
--- a/IHP/View/CSSFramework.hs
+++ b/IHP/View/CSSFramework.hs
@@ -78,7 +78,7 @@ instance Default CSSFramework where
                     validationResult = unless (get #disableValidationResult formField) ((get #styledValidationResult cssFramework) cssFramework formField)
 
                     renderCheckboxFormField :: FormField -> Blaze.Html
-                    renderCheckboxFormField formField@(FormField {fieldType, fieldName, fieldLabel, fieldValue, validatorResult, fieldClass, disableLabel, disableValidationResult, fieldInput, labelClass, required, autofocus }) = do
+                    renderCheckboxFormField formField@(FormField {fieldType, fieldName, fieldLabel, fieldValue, validatorResult, fieldClass, disableLabel, disableValidationResult, fieldInput, labelClass, required, autofocus, multiple }) = do
                         formGroup do
                             (H.div ! A.class_ "form-check") do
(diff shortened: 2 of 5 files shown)


Guide/form.markdown
```diff
@@ -269,6 +269,31 @@ This will render like:
...
+        type="email"
+        name="email"
+        id="post_email"
+        class="form-control"
+        multiple="multiple"
+    />
+</div>
+```
+
+Note: `multiple` attribute is only valid for email input, file input, and select
```
mpscholten, line 296: Can we also add a note here on how to consume a multi input from the action? I guess it's with `paramList`, but have not checked it.

s0kil: If someone creates a column such as `email TEXT[] NOT NULL`, meaning an list of emails. Then they probably expect to be able to do this:
```haskell
renderForm :: Person -> Html
renderForm person = formFor person [hsx|
    {(selectField #email emails) { multiple = True }}
    {submitButton}
|]
    where
        emails :: [Text]
        emails = ["test1@example.com", "test2@example.com", "test3@example.com"]
```

however, there is a compiler error:
```
Couldn't match type `SelectValue Text' with `[Text]'
        arising from a use of `selectField'
```

s0kil: It seems the `selectField` function does not account for multiple selections, or a list of emails.

amitaibu: @mpscholten any pointer on how to fix this?

amitaibu: Seems the above error is unrelated to the PR. I've tried with https://ihp.digitallyinduced.com/Guide/form.html#select-inputs-with-integers and seem to get a similar error

```haskell
renderForm :: Foo -> Html
renderForm foo = formFor foo [hsx|
    {(selectField #ints ints)}
    {submitButton}

|]
    where
        ints :: [Int]
        ints = [1..10]

instance CanSelect Int where
    type SelectValue Int = Int
    selectValue quantity = quantity
    selectLabel quantity = tshow quantity
```

image

mpscholten: This is not as easy to fix as it sounds.

Basically we'd need to change the data structure for the select field to support multiple values. Right now most form fields use this big [`FormField`](https://github.com/digitallyinduced/ihp/blob/master/IHP/View/Types.hs#L49) data structure.

We would like need to make a specialized `SelectField` data structure that is then returned by the `selectField` function. The `SelectField` then could have support for a list of values.

Not sure if that's worth the trouble for now.

amitaibu: Ok, so the current workaround, I guess, would be to use checkboxes as described in https://ihp.digitallyinduced.com/Guide/controller.html#multiple-params-with-same-name-checkboxes

mpscholten: Yep that works 👍

amitaibu: @mpscholten it might be too complex for my current skills, but maybe I could give it a try, for educational fun :smiley: 

Couple of questions I have:

1. Is it correct that the only diff should be in [fieldValue](https://github.com/digitallyinduced/ihp/blob/951141e99db554f59c1a884773006a12f4316ac0/IHP/View/Types.hs#L49). It should probably be `[Text]` instead of `Text`?
2. If so, would we copy the entire FormField, and change only that property, or should we move it to some instance?

If it's more elaborate than that, maybe someone could write down how the types would look, and I'll try to follow from there.

mpscholten: > Is it correct that the only diff should be in fieldValue. It should probably be [Text] instead of Text? 

yep that's basically it :) We need to be careful not to break BC with existing forms, that might actually be a bit tricky (e.g. that bug with the duplicate `label` field name we had recently on master would likely also be triggered here)

Likely the best way is to just get started and then we'll iterate our way there

SwarnaSSD (approved):

Issue #778, opened 2021-02-22
Missing bash dependenci


tino415:
I tried to [follow tutorial](https://ihp.digitallyinduced.com/Guide/index.html) in [nix docker container](https://hub.docker.com/r/nixos/nix/) and I get
```
88e5fa50902f:/app/blog# ./start
env: can't execute 'bash': No such file or directory
```
I build container using this Dockerfile
```Dockerfile
FROM nixos/nix:latest

RUN nix-channel --update 
RUN nix-env -f https://ihp.digitallyinduced.com/ihp-new.tar.gz -i ihp-new
```
I resolved it by running `nix-env -i bash` so probably missing dependenci? (I'm new to nix)

mpscholten:
Thanks for reporting :) I think we should add the bash dependency to `ihp-new`

PR #1005, opened 2021-07-25
Add Bash Dependency To ihp-new

s0kil:
Closes https://github.com/digitallyinduced/ihp/issues/778

### DIFF
--- a/ProjectGenerator/default.nix
+++ b/ProjectGenerator/default.nix
@@ -9,5 +9,5 @@ stdenv.mkDerivation rec {
     cp bin/ihp-new $out;
     makeWrapper $out/ihp-new $out/bin/ihp-new --prefix PATH ":" "${git}/bin";
   '';
-  buildInputs = [ git makeWrapper ];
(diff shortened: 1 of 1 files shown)


ProjectGenerator/default.nix
```diff
@@ -9,5 +9,5 @@ stdenv.mkDerivation rec {
     cp bin/ihp-new $out;
     makeWrapper $out/ihp-new $out/bin/ihp-new --prefix PATH ":" "${git}/bin";
   '';
-  buildInputs = [ git makeWrapper ];
-}
\ No newline at end of file
+  buildInputs = [ git makeWrapper bash_5 ];
```
mpscholten, line 12: I think this will not solve the issue as it's only defined as a build input here, but the bug is caused at runtime. For git we also use `makeWrapper` to make our compile time git available at runtime.

s0kil: So `bash` will also need to be wrapped with `makeWrapper`?

mpscholten: Yes

Issue #1010, opened 2021-07-28
[pagination] error clicking a number when there is a `maxItems`


CSchank:
I think this was a regression from removing the pagination JavaScript and going to IHP-only. It's caused by maxItems being a string instead of a number in the URL query parameter. I'll try to fix this and submit a PR.

mpscholten:
This has been fixed on master via https://github.com/digitallyinduced/ihp/commit/843398fdf5df8c3befb14ca166cee2d842a9b2ee :)

CSchank:
Okay. Great, I wondered if it had been fixed because I couldn't reproduce when I used the latest commit in Master. And yet I couldn't see anything newer than e8144c95830ebe4c3ccfded8451647d8db4909c6 in that file's commit history. Was this fixed in the function itself. Anyways, thanks!

mpscholten:
> Was this fixed in the function itself?

The problem was that `paramOrNothing @Text "maxItems"` returned a `Text`. Later that value is applied to `tshow`, and `tshow "some string"` always adds quotes.  By changing it to `paramOrNothing @Int` the call to `tshow` will call it on an `Int` which will not add quotes then.

CSchank:
Ah, yes. So the `paramOrNothing` function was fixed. Great!

PR #1011, opened 2021-07-28
small spelling fix

CSchank:
Very small fix for a spelling error I noticed when debugging #1010.


mpscholten (approved):
👍  good catch, thanks :)

PR #1019, opened 2021-08-03
Fix NUMERIC support (map to Data.Scientific)

### LINKED ISSUE #1017: Getting Incompatible error when using Numeric column
```
Incompatible {errSQLType = "numeric", errSQLTableOid = Just (Oid 17158), errSQLField = "cost",
 errHaskellType = "Float", errMessage = "types incompatible"}
```

Apparently the generated model type is Float, which is obviously incorrect. "Numeric" is a fixed-point number, so I would expect it to map to `Data.Fixed.Fixed`. As it is, it just doesn't work, apparently.

lierdakil:
Fixes #1017 

I've added no tests, but there were no tests for this to begin with.

### DIFF
--- a/IHP/Controller/Param.hs
+++ b/IHP/Controller/Param.hs
@@ -300,6 +300,16 @@ instance ParamReader Double where
                     Right integer -> Right (fromIntegral integer)
     readParameterJSON _ = Left "ParamReader Double: Expected Double"
 
+instance ParamReader Scientific.Scientific where
+    {-# INLINABLE readParameter #-}
+    readParameter byteString =
+        case Attoparsec.parseOnly (Attoparsec.scientific <* Attoparsec.endOfInput) byteString of
+            Right value -> Right value
+            Left error -> Left ("ParamReader Scientific: " <> cs error)
+
+    readParameterJSON (Aeson.Number number) = Right number
+    readParameterJSON _ = Left "ParamReader Scientific: Expected Scientific"
+
 instance ParamReader Float where
     {-# INLINABLE readParameter #-}
     readParameter byteString =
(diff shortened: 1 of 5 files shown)


IHP/Controller/Param.hs
```diff
@@ -300,6 +300,16 @@ instance ParamReader Double where
                     Right integer -> Right (fromIntegral integer)
     readParameterJSON _ = Left "ParamReader Double: Expected Double"
 
+instance ParamReader Scientific.Scientific where
```
mpscholten, line 303: Tests for ParamReader instances can be added here https://github.com/digitallyinduced/ihp/blob/master/Test/Controller/ParamSpec.hs :) Could you add a case there? :)

IHP/ModelSupport.hs
```diff
@@ -154,6 +155,9 @@ instance InputValue value => InputValue [value] where
 instance InputValue Aeson.Value where
     inputValue json = json |> Aeson.encode |> cs
 
+instance InputValue Scientific where
```
mpscholten, line 158: Tests for InputValue can be added here https://github.com/digitallyinduced/ihp/blob/master/Test/ModelSupportSpec.hs#L16 :)

lierdakil:
I've added some tests (mostly copy-pasted from similar tests for `Double`).

Test/Controller/ParamSpec.hs
```diff
@@ -141,32 +142,53 @@ tests = do
 
                 it "should accept floating point input" do
                     (readParameter @Double "1.2") `shouldBe` (Right 1.2)
-                    (readParameter @Float "1.2345679") `shouldBe` (Right 1.2345679)
-                
+                    (readParameter @Double "1.2345679") `shouldBe` (Right 1.2345679)
```
lierdakil, line 145: While at it, I've fixed an apparent copy-paste error in `ParamReader Double` tests. Which reasonably speaking doesn't affect anything, but hey.

mpscholten: Good catch :)

mpscholten (approved):
Thanks! :)

PR #1025, opened 2021-08-07
Add step to remove `<th></th>`

dharmatech:
It doesn't cause errors to have it in there, but it seems like a good idea to have beginners practice this.


mpscholten (approved):
Thanks 👍

Issue #1034, opened 2021-08-10
joined tables with `queryUnion` produces SQL error


CSchank:
I have the following queries:

```haskell
queryAllPublishedToGroups :: _ => _ -- something like [Id Group] -> JoinQueryBuilderWrapper joinRegister table 
queryAllPublishedToGroups groupIds =
    query @ModulePublish
        |> innerJoin @ModuleGroupPublish (#id, #modulePublishId)
        |> filterWhereInJoinedTable @ModuleGroupPublish (#groupId, groupIds)

queryAllPublishedToUser :: _ => _ -- something like JoinQueryBuilderWrapper joinRegister table 
queryAllPublishedToUser =
    query @ModulePublish
        |> innerJoin @ModuleUserPublish (#id, #modulePublishId)
        |> filterWhereJoinedTable @ModuleUserPublish (#userId, currentUserId)

queryAllPublishedToPublic :: _ => _ -- something like JoinQueryBuilderWrapper joinRegister table 
queryAllPublishedToPublic =
    query @ModulePublish
        |> filterWhere (#publishExtent, PublishPublic)
```

In my action they're combined together like:

```haskell
modulePublishes <- 
           queryAllPublishedToGroups userGroupIds
        |> queryUnion queryAllPublishedToUser
        |> queryUnion queryAllPublishedToPublic
        |> filterWhere (#publishType, AbilityToImport)
        |> fetch
        >>= collectionFetchRelated #snapshotId
```

And then I run this action, I get:

```haskell
SqlError {sqlState = "42P01", sqlExecStatus = FatalError, sqlErrorMsg = "missing FROM-clause entry for table \"module_user_publishes\"", sqlErrorDetail = "", sqlErrorHint = ""}
```

CSchank:
@hllizi I believe you worked on joins, so I'm tagging you here as well. :)

hllizi:
Hi, I just submitted a pull request, but it's not yet a solution that will make your code work, it will just change the error message. Union query builders do not yet support complex queries. The mistake was not to treat join queries as complex.

CSchank:
Okay, well that's better than nothing for now. I take it it's going to take significant work to make things work properly.

hllizi:
[...]

PR #1038, opened 2021-08-12
refuse to build union query if one of the queries contains joins

hllizi:
'Fix' for git@github.com:digitallyinduced/ihp.git #1034: join queries need to be regarded as complex queries.


mpscholten (approved):
Thanks :) (tests are only failing because i broke master 🔧 )

Issue #990, opened 2021-07-21
`paginate` function does not work with joined queries


GarrisonJ:
When I try to use the `paginate` function with a joined query I get a type error that `JoinQueryBuilderWrapper` is not a `QueryBuilder`.

joshua-obritsch:
I've also noticed this. Would be nice to be able to use joined queries.

CSchank:
I need to learn a bit more about the underlying type system to support this, but I feel like it should be doable. Perhaps @mpscholten could guide me a bit or tell me if there are similar functions which can handle different types of `QueryBuilder`-type things.

GarrisonJ:
Yeah, I don't understand either. 

But I do see this: https://github.com/digitallyinduced/ihp/blob/master/IHP/QueryBuilder.hs#L183 You can get the query builder out of a `JoinQueryBuilderWrapper` with the `getQueryBuilder` function.

GarrisonJ:
Maybe this function https://github.com/digitallyinduced/ihp/blob/acddf9d4c535309ce5d02da945bfc556cad9f457/IHP/Pagination/ControllerFunctions.hs#L52 should take an instance of `HasQueryBuilder` instead of a `QueryBuilder` itself.

CSchank:
Yes, I'll look into it. The second option is likely the most promising path forward.

mpscholten:
This has been fixed by @CSchank on master :) You can follow https://ihp.digitallyinduced.com/Guide/updating.html to update to latest master :)

CSchank:
@GarrisonJ Thanks for your idea, it worked perfectly! I didn't go general enough originally. Now, with this PR, it can work with anything that has `HasQueryBuilder`, which makes perfect sense since the `paginate` function only relies on `fetchCount` which has implementations for `HasQueryBuilder` data types, and I also fixed `filterList` to do the same, since it relies on `filterWhereIn`.

PR #1040, opened 2021-08-12
generalize pagination to work with all types of `HasQueryBuilder`

CSchank:
This addresses #990. When I wrote this originally I was not nearly as familiar with IHP's codebase and as such I didn't go general enough. This PR fixes the `paginate` and `filterList` functions so they work with anything which is in the `HasQueryBuilder` typeclass. This includes joined queries!! :) This is how it should have been written from the start.

This should not break any existing code using `paginate` and/or `filterList`, but should allow for a broader range of use for the pagination and more complex queries.

### DIFF
--- a/IHP/Pagination/ControllerFunctions.hs
+++ b/IHP/Pagination/ControllerFunctions.hs
@@ -18,7 +18,7 @@ import IHP.Pagination.Types
     ( Options(..), Pagination(..) )
 
 import IHP.QueryBuilder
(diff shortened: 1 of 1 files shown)


mpscholten (approved):
Perfect, thanks 💯  :)

CSchank:
You are very welcome, I ended up needing this in my own project as well which motivated me to finally do it!

PR #1041, opened 2021-08-12
print sql query if errors

CSchank:
This is a simple one: when you get an error like in #1034 you don't get to see the SQL query that produced it. This PR prints it out so you get to know what query caused the error. It's kind of ugly tbh, so maybe you have an idea of how to make it prettier. E.g:

```
IHP> IHP> *** Exception: Error in query: 
"INSERT INTO user_group_roles (user_group_role, group_id, user_id, id) VALUES (?, ?, ?, ?) RETURNING *"
libpq: failed (could not connect to server: No such file or directory
        Is the server running locally and accepting
        connections on Unix domain socket "<..>/build/db/.s.PGSQL.5432"?
)
  error, called at ./IHP/ModelSupport.hs:333:21 in ihp-0.13.0-EAKYYT2yzmdKJVICEpohgJ:IHP.ModelSupport
```


IHP/ModelSupport.hs
```diff
@@ -326,11 +326,16 @@ measureTimeIfLogging queryAction theQuery theParameters = do
...
-            result <- queryAction
-            end <- getCurrentTime
-            let theTime = end `diffUTCTime` start
-            logQuery theQuery theParameters theTime
-            pure result
+            result <- try $ queryAction
+
+            case result of
+                Left (e :: SomeException) -> do
+                    error $ "\nError in query: " <> show theQuery <> "\n" <> show e
```
mpscholten, line 333: This will break some special error handling logic here: https://github.com/digitallyinduced/ihp/blob/master/IHP/ErrorController.hs#L172

We should add a `QueryException { query :: ByteString, pgException :: SomeException }` and then we could update the ErrorController to deal with that exception

mpscholten: Just pushed this new exception onto this branch 👍

Issue #1030, opened 2021-08-08
Keep message through multiple redirects


hllizi:
Messages set with `setSuccessMessage` and its ilk do not survive multiple redirects. A simple toy example:

Step1 controller:
```
    action TestStep1Action = do
        setSuccessMessage "Step 1"
        redirectTo TestStep2Action
```

Step2 controller:
```
    action TestStep2Action = do
        message <- getSuccessMessage
        redirectTo TestStep3Action
```
Step3 controller:
```
    action TestStep3Action = do
        render TestView
```
The (by me) desired behaviour would be to have TestView displayed with the flash message "Step 1". But the message is missing. Rendering TestView in Step2 shows the message in the usual way.

Is this a bug or are there reasons for this behaviour that I don't yet see? If the latter should be the case, is there a sensible workaround to get the message passed on?

mpscholten:
It's caused by [the call to `initFlashMessage`](https://github.com/digitallyinduced/ihp/blob/master/IHP/ControllerSupport.hs#L97) that will put the flash messsage context into the controller context and clears it.

In the past this wasn't solvable otherwise. Nowadays we could move the call to `renderHtml` near the `beforeRender` lifecycle event: https://github.com/digitallyinduced/ihp/blob/master/IHP/Controller/Render.hs#L62

Then the flash message would only be cleared when a view is rendered.

PR #1046, opened 2021-08-16
Pass flash messages on

hllizi:
OK, I mindlessly moved the call to initFlashMessages to the place mentioned in #1030 and it indeed works. Could be merged if found desirable.


mpscholten (approved):
Perfect, thanks :)

PR #1050, opened 2021-08-20
Added allEnumValues helper docs to Form documentation

Montmorency:
All `allEnumValues` very useful function but I've forgotten about it a couple times and thought it would be useful to have an extra reference to it right next to the existing Form documentation on `CanSelect` and selectfields. At the moment it is only referred to in `recipes.markdown`.


mpscholten (approved):
Great addition! Thanks :)

Issue #1042, opened 2021-08-14
Data editor displays textfields in incorrect order


joshua-obritsch:
When using version 0.12.0, I was able to create a table and modify the order of columns in `Schema.sql` without any issue.

However, after upgrading to version 0.13.1, if I change the order of columns in `Schema.sql`, the changes are not reflected in the data editor. Thus, the textfields are out of order and submitting the form typically results in a mismatched types error.

How to reproduce:
* Create a table called `test` in the schema editor.
* Add column `first_field` with type `TEXT` in the schema editor.
* Add column `second_field` with type `BOOLEAN` in the schema editor.
* Swap the positions of `first_field` and `second_field` in `Schema.sql`.
* Try to add a new row in the data editor.

joshua-obritsch:
After removing and recreating some of the tables in the schema editor, the order of the textfields is still wrong. It seems to be ordering the textfields by type. For example, `timestamp` comes before `uuid` which comes before `text` etc. yet this ordering is still not reflected in `Schema.sql`.

mpscholten:
In the steps to reproduce, at what point are you calling `Update DB`?

joshua-obritsch:
After adding the second column. I just tried this again with 0.14.0 and after adding `first_field` and `second_field` in that order in the `test` table, I immediately see that the order of the columns is reversed in the data editor without even touching `Schema.sql`.

joshua-obritsch:
I just tested this a few more times, and the order in the data editor seems to be random. Sometimes `first_field` appears first, and sometimes second. The order seems to be determined though when I create the columns in the schema editor.

mpscholten:
One thing that could be confusing here: The data editor is not using the Schema.sql. Instead it queries the actual table inside the PG database to see what columns are available.

So unless you run `Update DB` the data editor might still use the old order.

joshua-obritsch:
I've been running `Update DB`.

So, I went through the commit history and found that the problem is related to this commit: https://github.com/digitallyinduced/ihp/commit/acddf9d4c535309ce5d02da945bfc556cad9f457

Before this commit, everything works fine. After this commit, the problem starts. I'll take a look at the source code and see if I can narrow it down some more.

joshua-obritsch:
So, I've managed to figure out the following:

[...]

PR #1061, opened 2021-08-28
Fix bug regarding order of fields in data editor

joshua-obritsch:
This addresses issue https://github.com/digitallyinduced/ihp/issues/1042


mpscholten (approved):
Thanks :) Good job 👍

PR #1064, opened 2021-08-29
Add an example of the BlazeHTML equivalent of HSX

dharmatech:


Guide/hsx.markdown
```diff
@@ -10,6 +10,52 @@ HSX can be written pretty much like normal HTML. You can write an HSX expression
 
 Because the HSX is parsed, you will get a syntax error when you type in invalid HTML.
 
+## Example: HSX and the equivalent BlazeHtml
```
mpscholten, line 13: Can we move this new section after the basic syntax explanations or at the end of the file? :) I think the basic syntax is more important than the blaze details, so it should come first

dharmatech: I agree with you. :-) This is indeed somewhat an "under the hood" explanation for the curious. I don't feel strongly about having it in here. I was curious about how this worked. If you're OK with having it towards the end of the page, cool.

dharmatech: OK, I've moved it. 👍

mpscholten (approved):
Thanks :) Really appreciate your additions!

PR #1066, opened 2021-08-30
potential mod for create form actions to helpers.js 

Montmorency:
This PR would make the behaviour for a failed Form request the same with and without Turbolinks. The pattern is that if a   `Left branch` failed form submission leads to a `render SomeView {..}` the form creation action and controller URL will match. In this case we likely do not want the url to update to the target for the form creation because on a user manual browser refresh this will trigger a `GET` request to the Create action which leads to a `400` error (those endpoints require `POST` by IHP convention). This PR catches the url equality and doesn't update the local url so a refresh triggers the page reload.

Tested [here](https://github.com/Montmorency/ihp-create-action).

### DIFF
--- a/lib/IHP/static/helpers.js
+++ b/lib/IHP/static/helpers.js
@@ -187,7 +187,9 @@ window.submitForm = function (form, possibleClickedButton) {
             );
             transitionToNewPage(request.response);
             Turbolinks.clearCache();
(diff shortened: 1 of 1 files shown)


mpscholten (approved):
👍  I was always thinking that we already had fixed this bug in the past. With the two branches here it totally makes sense :) Thanks!

PR #1073, opened 2021-09-01
remove some unneeded/wrong rounded corners

hendi:
Please merge this ASAP to get rid of atrocities like this:

1-before

2-before

which this splendid PR brings to glorious pixel perfection:

1-after

2-after


mpscholten (approved):
Finally fixed haha :)

PR #1074, opened 2021-09-01
add `matchesRegex` validator

hendi:
use like 

```haskell
buildPost post = post
    |> fill @'["year"]
    |> validateField #year (matchesRegex "^[0-9]{4}$")
```

(bad example regex, please don't use in production ;))


s0kil:
The error message could be improved, it's too technical for end user.

hendi:
It's rather impossible to provide easy to understand explanations for regular expressions, in general. For less technical / more helpful error messages there's `withCustomErrorMessage`.

mpscholten (approved):
Perfect, thanks :)

PR #1078, opened 2021-09-03
Typo on ContentType in CanSelect form docs.

Montmorency:
Accidentally pluralized data type. Could be confusing for someone following docs.


mpscholten (approved):
Good catch 👍

PR #1079, opened 2021-09-03
Codegen app.js

Montmorency:
Suggestion that if `app.js` is shipped with ihp-boilerplate might make sense to have it in the CodeGen as well. E.g. a developer who opens a test project and starts playing with customizing `app.js`  but changes aren't getting picked up in starter project. Also attaching new cache buster to this automatically makes sure production app.js changes are picked up straight away.


mpscholten (approved):
That's a good one 👍 thanks

PR #1088, opened 2021-09-14
add setIfJust which ignores updates on Nothing

hendi:
This adds a function `setIfJust` which is similar to `set` but which takes a `Mabye value` and updates the record if the value is `Just`, and doesn't change it if it's `Nothing`.

### DIFF
--- a/IHP/HaskellSupport.hs
+++ b/IHP/HaskellSupport.hs
@@ -12,6 +12,7 @@ module IHP.HaskellSupport
 , get
 , set
 , setJust
+, setIfJust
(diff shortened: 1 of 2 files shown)


s0kil:
Does it make more sense to call it `setMaybe`?

zacwood9:
Agreed with setMaybe since it aligns with fromMaybe

hendi:
Thanks, makes sense! Changed.

mpscholten (approved):
Nice, good addition 👍

PR #1090, opened 2021-09-23
Update OAuth documentation

GarrisonJ:
I'm not sure, but I think the actual intention was for this path to be GoogleConnectCallbackAction


mpscholten (approved):
Ah yes good catch 👍 :)

PR #1101, opened 2021-10-02
Add note about multiple nested resources

amitaibu:
can have multiple resources


mpscholten (approved):
Good addition thanks :)

mpscholten:
btw: there's also `Include'` which allows you to write `Include' ["tags", "comments"] Post` instead of `Include "tags" (Include "comments" Post)`

amitaibu:
Oh, then maybe I'll change doc to reflect that, as it's easier

PR #1104, opened 2021-10-05
link api docs in guide

jeyj0:
Every mention of an IHP function that I could find is now linked to the corresponding documentation. I changed the formatting where it made sense. Functions in code blocks are not linked, as that doesn't work due to markdown limitations. (It would be pretty cool though...)

### DIFF
--- a/Guide/assets.markdown
+++ b/Guide/assets.markdown
@@ -15,7 +15,7 @@ To avoid this problem, web applications typically append a hash to the url of yo
 <script src="/app.js?v=19319eb"></script>
 ```
 
-IHP provides an `assetPath` view helper to automatically add these hashes:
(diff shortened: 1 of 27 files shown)


mpscholten:
Thanks 👍 

```bash
cd Guide
make watch
```

fails for me with this error:

```
   |
82 | When the `callServerAction('IncrementCounterAction')` is called, it will trigger the [`action state IncrementCounterAction = do`]() haskell block to be called on the server.
   |                                                                                                                                   ^
unexpected ')'
expecting '<', URI, or white space
make: *** [Makefile:53: server-side-components.html] Error 1
```

jeyj0:
No clue how I did not notice that, as I had `make watch` running continuously... Fixed

mpscholten (approved):

PR #1126, opened 2021-10-10
Fix Testing docs

amitaibu:


amitaibu:
`:l Test/Main` is actually for IHP lib. Not for a project

mpscholten:
Yes 👍 

> :l Test/Main is actually for IHP lib. Not for a project

Should be the same for the actual project itself. Or then we should update the testing docs to make clear that tests should be stored in the `Test/` directory of the project (which needs to be created on first use)

mpscholten:
> So should I copy the Test/Main.hs from IHP lib?

Yes that sounds good to me 👍

amitaibu:
Getting closer.. I'm trying to add it [here](https://github.com/amitaibu/ihp-bid/pull/5/files) as a playground, but now have 

```
Test/Controller/BidSpec.hs:20:19: error:
    Variable not in scope: makeConfig :: IO ConfigBuilder
   |
20 | spec = beforeAll (makeConfig >>= mockContext WebApplication) do
```

However I couldn't fund `makeConfig` - where is it coming from?

amitaibu:
Note to self, add comment about 

```
ghci
GHCi, version 8.10.3: https://www.haskell.org/ghc/  :? for help
*** WARNING: . is writable by someone else, IGNORING!
Suggested fix: execute 'chmod go-w .'
```

amitaibu:
PR is actually outdated with your changes - I was just working on an updated one :)

Issue #1099, opened 2021-10-02
`nonEmpty` validation doesn't work on Int values


amitaibu:
I have a non nullable Int field, however I cannot add `nonEmpty` field on it as there's the error `Could not deduce (IsEmpty Int) arising from a use of `nonEmpty'`

mpscholten:
If the int field is empty, the int field parser will fail to parse the empty string and produce an error like this:

![image](https://user-images.githubusercontent.com/2072185/135745049-0d197663-dfb5-4919-a746-9e9a093aa7d2.png)


So i think no `nonEmpty` is needed in this case. Or do i miss something?

amitaibu:
I have this an in exiting Repo, where the error of validation of the form is that of a sub-record.
Here's the code - https://github.com/amitaibu/ihp-bid/blob/a02ee2860811baf35b591bebf4c44d1e6274be99/Web/Controller/Items.hs#L71

## Steps to reproduce

1. Checkout https://github.com/amitaibu/ihp-bid
2. Visit http://localhost:8000/NewItem
2. Add title
3. Submit form

## Current behavior


![Selection_999(266)](https://user-images.githubusercontent.com/125707/135845322-faa6280d-3cf5-470f-847c-a4bc7267a96d.jpg)


![image](https://user-images.githubusercontent.com/125707/135845390-f5118a91-1f63-484c-bf4c-0152eaeca104.png)

amitaibu:
I was able to find the line that is causing the `'min' is invalid CallStack...` and that's `getValidationFailure #min bidStep`

Still not sure why, I'll try to dig.

PR #1130, opened 2021-10-11
Add paramListOrNothing function

amitaibu:
Trying to address https://github.com/digitallyinduced/ihp/issues/1099#issuecomment-939147941

I have followed https://github.com/digitallyinduced/ihp/blob/5f9f65ae272a9a3d1c8c89c0ce5cfe2391ad057c/CONTRIBUTING.md but facing an error upon compilation, so I could use some help here :sweat_smile: 

image

### DIFF
--- a/IHP/Controller/Param.hs
+++ b/IHP/Controller/Param.hs
@@ -138,15 +138,15 @@ paramList name =
 -- []
 --
 --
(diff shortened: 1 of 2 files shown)


mpscholten:
Did you run the `compile` function?

amitaibu:
I have `:set -XPackageImports` on ghci but now there are more errors. Am I'm doing something wrong here? :smile_cat: 

Selection_999(322)

PR #1174, opened 2021-10-26
Breadcrumbs functionality

amitaibu:
https://github.com/digitallyinduced/ihp/discussions/1166

Moving Breadcrumbs to use CSS framework

```haskell
instance View ShowView where
    html ShowView { .. } = [hsx|
        {breadcrumbs}

        <h1>{get #title post}</h1>
        <p>{get #createdAt post |> timeAgo}</p>
    |]
        where breadcrumbs = renderBreadcrumbs
                [ BreadcrumbsItem
                    { label = [hsx|Posts|]
                    , url = Just $ pathTo PostsAction
                    , isActive = False
                    }
                , BreadcrumbsItem
                    { label = [hsx|Show Post|]
                    , url = Nothing
                    , isActive = True
                    }
                ]
```

Selection_999(376)


amitaibu:
Ready for review :smile_cat:

IHP/Breadcrumb/ViewFunctions.hs
```diff
@@ -0,0 +1,43 @@
...
+        where
+            breadcrumbsView = BreadcrumbsView
+                { cssFramework = theCSSFramework
+                , breadcrumbItems = breadcrumbItemsRendered
+                }
+
+            breadcrumbItemsRendered =  [hsx|{forEach breadcrumbItems (styledBreadcrumbItem theCSSFramework theCSSFramework breadcrumbItems)}|]
+
+
+breadcrumbWithLink :: (Show controller, AutoRoute controller) => Html -> controller -> BreadcrumbItem
```
mpscholten, line 31: Could you add haddock comments to all functions in this module? This will show up in tooltips when using e.g. VSCode with Haskell Language Server. Therefore it's very worthwhile to add this :)

IHP/View/CSSFramework.hs
```diff
@@ -239,6 +243,29 @@ instance Default CSSFramework where
...
+                    </ol>
+                </nav>
+            |]
+
+
+            styledBreadcrumbItem :: CSSFramework -> [ BreadcrumbItem ]-> BreadcrumbItem -> Blaze.Html
+            styledBreadcrumbItem _ breadcrumbItems breadcrumbItem@BreadcrumbItem {label, url} =
+                let
+                    lastItemLabel = breadcrumbItems |> last |> maybe mempty (get #label)
+                    isLastItem = renderHtml lastItemLabel == renderHtml label
```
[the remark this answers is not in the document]
mpscholten: This feels a bit inefficient. Can't we pass a boolean `isLast` argument to `styledBreadcrumbItem` and then pass it from the `forEach` that is calling `styledBreadcrumbItem`?

IHP/Breadcrumb/ViewFunctions.hs
mpscholten: Maybe it also makes sense to add a short paragraph with examples to https://ihp.digitallyinduced.com/Guide/view.html#common-view-tasks ?

IHP/IDE/CodeGen/ViewGenerator.hs
```diff
@@ -65,127 +66,157 @@ buildPlan' schema config =
...
+                    html ShowView { .. } = [hsx|
+                        {breadcrumb}
+                        <h1>Show ${singularName}</h1>
+                        <p>{${singularVariableName}}</p>
+
+                    ${qqClose}
+                        where
+                            breadcrumb = renderBreadcrumb
+                                            [ breadcrumbWithLink "${pluralName}" ${indexAction}
+                                            , breadcrumbWithoutLink [hsx|Show ${singularName}${qqClose}
```
mpscholten, line 116: To make this consistent we should use `""` here instead of `[hsx||]` on all calls :)

```suggestion
                                            , breadcrumbWithoutLink "Show ${singularName}"
```

IHP/Breadcrumb/ViewFunctions.hs
mpscholten, line 31: What do you think about naming these function `breadcrumbLink` and `breadcrumbText`? IMO these look a little bit shorter to me and it feels nicer as they line up perfectly because they have the same string length :)

image


image

IHP/View/Types.hs
```diff
@@ -99,6 +101,12 @@ data InputType
     | FileInput
 
 
+data BreadcrumbsView =
+    BreadcrumbsView
+    { cssFramework :: !CSSFramework
```
amitaibu, line 106: This `cssFramework` is never used, so I think can we remove it?

mpscholten: Yes 👍 

What do you think about renaming this data structure to just `Breadcrumb`?

amitaibu: Your call, but I think it's better like that so it's consistent with `PaginationView`

amitaibu: Actually in this case we don't need this at all, as we already pass `[BreadcrumbItem]`

amitaibu: Sorry, we do need it - to pass the `BreadcrumbItem`s as HTML

amitaibu:
Thanks for the review - I've addressed new comments (only haven't changed [this](https://github.com/digitallyinduced/ihp/pull/1174#discussion_r739658475) - let me know if you'd want me still to rename it)

mpscholten:
I've created a new project with a `Posts` controller.

image

The generated breadcrumb in my showview is not having `active` class here. I think it was working the last time i tested it :) 

The expected look would be this (added the `active` class manually with the chrome inspector):

image

IHP/Breadcrumb/ViewFunctions.hs
```diff
@@ -0,0 +1,45 @@
...
+import IHP.ViewSupport (theCSSFramework)
+import IHP.ControllerPrelude
+
+renderBreadcrumb :: (?context :: ControllerContext) => [BreadcrumbItem] -> Html
+renderBreadcrumb breadcrumbItems = styledBreadcrumb theCSSFramework theCSSFramework breadcrumbItems breadcrumbsView
+        where
+            breadcrumbsView = BreadcrumbsView
+                { breadcrumbItems = forEachWithIndex breadcrumbItems (\(index, breadcrumbItem) ->
+                    let
+                        isLast = index == length breadcrumbItems
```
mpscholten, line 26: I think the issue is here :) Like this should be:

```haskell
isLast = index == (length breadcrumbItems - 1)
```

mpscholten: Maybe it makes sense to also add a test for this rendering function? That would have caught the issue earlier

amitaibu: Good catch, thanks.

Here's the test I'd like to write

```haskell
it "should render the wrapping breadcrumb and last item as active" do
    let breadcrumbs = [breadcrumbText "First item", breadcrumbText "Last item"]

    -- This below doesn't work
    let frameworkConfig = getFrameworkConfig ?context

    renderBreadcrumb breadcrumbs `shouldRenderTo` "..."
```

How can we mock a `?context` in the test?

amitaibu: Ok, I found:

`let ?context = createControllerContextWithParams [("page", "1")]`

and

```haskell
context <- newControllerContext
let ?context = context
```

amitaibu: @mpscholten I'll appreciate your help here :smiley_cat: I believe we'd need a more generic `let ?context = mockControllerContext` so that we can hand it over in turn into `renderBreadcrumb`

mpscholten: The `createControllerContextWithParams` is defined like this:

```haskell
createControllerContextWithParams params =
        let
            requestBody = FormBody { params, files = [] }
            request = Wai.defaultRequest
            requestContext = RequestContext { request, respond = error "respond", requestBody, vault = error "vault", frameworkConfig = error "frameworkConfig" }
        in FrozenControllerContext { requestContext, customFields = TypeMap.empty }
```

I guess in our case the problem is that `theCSSFramework` accesses the `frameworkConfig` field of the context to retrieve the current configured css framework.

With the framework config it could look like this:

```haskell
createControllerContextWithCSSFramework cssFramework =
        let
            frameworkConfig = FrameworkConfig.buildFrameworkConfig do
                    option cssFramework
            requestBody = FormBody { params, files = [] }
            request = Wai.defaultRequest
            requestContext = RequestContext { request, respond = error "respond", requestBody, vault = error "vault", frameworkConfig = frameworkConfig }
        in FrozenControllerContext { requestContext, customFields = TypeMap.empty }
```

amitaibu:
`Active` fixed and test added, ready for review

mpscholten (approved):
Good job! :)

PR #1183, opened 2021-10-29
Explain the reason for passing a second `CSSFramework`

amitaibu:
Edited from https://github.com/digitallyinduced/ihp/pull/1157#discussion_r732235585

### DIFF
--- a/IHP/View/Types.hs
+++ b/IHP/View/Types.hs
@@ -116,8 +116,56 @@ data PaginationView =
     }
 
 -- | Render functions to render with Bootstrap, Tailwind CSS etc.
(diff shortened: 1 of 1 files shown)


mpscholten:
The `cssFramework` inside the `FormContext` is used by `textField` and other functions to access the `cssFramework` to render things. [Specifically it's used here](https://github.com/digitallyinduced/ihp/blob/6d1725620a53e0d202354d113d6cd6c9d3b2abf9/IHP/View/Form.hs#L741) to render the form field to HTML. The `cssFramework` variable is basically just passed around the form functions to be available inside that render function.

mpscholten (approved):
Thanks :)

PR #1190, opened 2021-11-02
Fixed breadcrumb label field causing 'Record update is amiguous and r…

mpscholten:
…equires a type signature' error

### DIFF
--- a/IHP/Breadcrumb/Types.hs
+++ b/IHP/Breadcrumb/Types.hs
@@ -7,13 +7,13 @@ import ClassyPrelude
 
 data BreadcrumbItem =
     BreadcrumbItem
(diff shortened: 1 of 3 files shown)


jeyj0 (approved):

amitaibu:
I haven't noticed this error - what has caused it?

mpscholten:
The `SubmitButton` type also has a `label` field https://github.com/digitallyinduced/ihp/blob/master/IHP/View/Types.hs#L67 :) Now when we have code like this:

```haskell
{submitButton {label="Weiter"}}
```

This will cause the following error:
image (28)

GHC cannot figure out that we're dealing with a `SubmitButton` here, because the `BreadcrumbItem` also has a `label` field :) Therefore the fix is to rename the breadcrumb label field to something else. Then it's unique again and GHC can figure it out again.

PR #1202, opened 2021-11-07
Don't show pagination if not needed

amitaibu:
If we have less than two pages, we shouldn't show the Pagination at-all

### DIFF
--- a/IHP/View/CSSFramework.hs
+++ b/IHP/View/CSSFramework.hs
@@ -170,28 +170,29 @@ instance Default CSSFramework where
             styledSubmitButtonClass = ""
 
             styledPagination :: CSSFramework -> PaginationView -> Blaze.Html
-            styledPagination _ paginationView =
-                [hsx|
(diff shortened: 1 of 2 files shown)


IHP/View/CSSFramework.hs
```diff
@@ -170,28 +170,29 @@ instance Default CSSFramework where
             styledSubmitButtonClass = ""
 
             styledPagination :: CSSFramework -> PaginationView -> Blaze.Html
-            styledPagination _ paginationView =
-                [hsx|
+            styledPagination _ paginationView@PaginationView {pagination, linkPrevious, linkNext, pageDotDotItems, itemsPerPageSelector} =
+                when (showPagination pagination)
```
mpscholten, line 174: IMO it would be better to put the view logic here into [`renderPagination`](https://github.com/digitallyinduced/ihp/blob/master/IHP/Pagination/ViewFunctions.hs#L30). The CSSFramework should only deal with very basic html generation / styling.

amitaibu: Good point :+1:

amitaibu:
Closed in favor of https://github.com/digitallyinduced/ihp/pull/1207

PR #1207, opened 2021-11-08
Don't show pagination if not needed

amitaibu:


Test/View/CSSFrameworkSpec.hs
```diff
@@ -219,6 +220,20 @@ tests = do
                     let render = Blaze.renderMarkup $ styledPagination cssFramework cssFramework paginationView
                     Text.isInfixOf "<div class=\"d-flex justify-content-md-center\">" (cs render) `shouldBe` True
 
+                it "should not render the pagination if there aren't enough elements" do
```
amitaibu, line 223: I'll add another test to confirm Pagination shows when there are enough elements

amitaibu: done

mpscholten (approved):
Perfect, thanks :)

PR #1213, opened 2021-11-12
Fix job dashboard retry

hendi:
In a last [commit](https://github.com/digitallyinduced/ihp/commit/272565698c56dcfeb56b474e3d8e493c6a4b5394) the "retry" button on the job list view has been fixed, but not on the Job's detail page. There was also a wrong success message (talking about "job record deleted") which this PR also addresses.


mpscholten (approved):
perfect thx 👍

PR #1216, opened 2021-11-12
Add troubleshooting steps for Purescript on IHP Cloud

jyoder:
I had run into a few issues trying to get my app deployed on IHP Cloud. These are the steps that I took to get it deploying and running properly.


mpscholten (approved):
Thanks 👍

PR #1224, opened 2021-11-20
Modernize FormField's CSSframework integration

amitaibu:
- Move Blaze to hsx
- Adapt TW implementation to look better


IHP/View/CSSFramework.hs
```diff
@@ -22,6 +22,7 @@ import IHP.Breadcrumb.Types
 import IHP.Pagination.Helpers
 import IHP.Pagination.Types
 import IHP.View.Types (PaginationView(linkPrevious, pagination))
+import Control.Lens (element)
```
mpscholten, line 25: This looks like a typo / wrong import :)

IHP/View/CSSFramework.hs
```diff
@@ -93,52 +93,104 @@ instance Default CSSFramework where
...
+                                            {theInput}
+                                            {validationResult}
+                                            {helpText}
+                                        </div>
+                                    |]
+                                else [hsx|
+                                        {theInput}
+                                        <label
+                                            class={classes [("form-check-label", labelClass == ""), (labelClass, labelClass /= "")]}
+                                            for={fieldInputId}
```
amitaibu, line 126: On `master` the `for` is missing. Was this for a reason?

IHP/View/CSSFramework.hs
```diff
@@ -93,52 +93,104 @@ instance Default CSSFramework where
...
+                                            name={fieldName}
+                                            id={fieldInputId}
+                                            class={classes [inputClass, (inputInvalidClass, isJust validatorResult), (fieldClass, not (null fieldClass))]}
+                                            value={fieldValue}
+                                            disabled={disabled}
+                                            required={required}
+                                            autofocus={autofocus}
+                                        >
+                                            <option selected={not isValueSelected} disabled={True}>{placeholder}</option>
+                                            {forEach (options fieldType) (getOption)}
```
amitaibu, line 179: There are now spaces that appear on the test, but I don't know where they are coming from.

image

amitaibu:
@mpscholten in order to override the HTML structure/add classes on TW we would currently need to override `styledFormField`. Maybe instead of calling those internal functions, we should convert them to styled...

https://github.com/digitallyinduced/ihp/blob/341a58473143c5ec56c5fdeac68e447521e0a90a/IHP/View/CSSFramework.hs#L71

```diff
- SelectInput {} -> renderSelectField formField
+ SelectInput {} -> styledSelectField formField 
```

So for TW we would override `styledSelectField`

IHP/View/CSSFramework.hs
```diff
@@ -93,52 +93,104 @@ instance Default CSSFramework where
...
+                                                name={fieldName}
+                                                class={classes ["form-check-input", (inputInvalidClass, isJust validatorResult), (fieldClass, not (null fieldClass))]}
+                                                id={fieldInputId}
+                                                checked={fieldValue == "yes"}
+                                                required={required}
+                                                disabled={disabled}
+                                                autofocus={autofocus}
+                                            />
+
+                                            <input type="hidden" name={fieldName} value={inputValue False} />
```
amitaibu, line 111: Why do we need the hidden input?

mpscholten: If the checkbox is checked off, the browser will not send the parameter as part of the form. This will then make it impossible to set a field to `False` using a checkbox.

Now when the checkbox is ignored because it's checked off, the hidden field with `value={inputValue False}` is used to set the parameter to `False`

IHP/View/CSSFramework.hs
mpscholten: Likely a bug :)

IHP/View/CSSFramework.hs
mpscholten: These whitespaces come from HSX expression like `<some-tag> {someVariableExpr}`. In this case HSX cannot safely remove the full whitespace between these two nodes as  we don't know whats inside `someVariableExpr` at compile-time.

You could rewrite the code like this:

```haskell
<option selected={not isValueSelected} disabled={True}>{placeholder}</option>{forEach (options fieldType) (getOption)}
```

Doesn't look nice, but get rids of the space there then :)

mpscholten:
> @mpscholten in order to override the HTML structure/add classes on TW we would currently need to override `styledFormField`. Maybe instead of calling those internal functions, we should convert them to styled...
> 
> https://github.com/digitallyinduced/ihp/blob/341a58473143c5ec56c5fdeac68e447521e0a90a/IHP/View/CSSFramework.hs#L58-L72
> 
> ```diff
> - SelectInput {} -> renderSelectField formField
> + SelectInput {} -> styledSelectField formField 
> ```
> 
> So for TW we would override `styledSelectField`

Sounds good to me 👍

IHP/View/CSSFramework.hs
amitaibu: Those spaces are actually between the closing `</select>` and the closing `</div>`. They are the empty spaces as a result of 

```
{validationResult}
{helpText}
```

I could solve it by adding two spaces to the expected result, but not sure it's the right way.

amitaibu:
The last commit brings more consistency to the `styled-*` functions, making sure they always get `CSSFrameWork` as the first argument, and some code cleanup.

IHP/View/CSSFramework.hs
```diff
@@ -35,6 +35,10 @@ instance Default CSSFramework where
                         ErrorFlashMessage message -> [hsx|<div>{message}</div>|]
                 , styledFlashMessages
                 , styledFormField
+                , styledTextFormField
+                , styledCheckboxFormField
+                , styledSelectFormField
```
amitaibu, line 40: @mpscholten I'd like to check the `selectField` . Is there an example somewhere on how to populate an Int field with custom input. I'm looking for something like `selectField #number [(1, "one"), (2, "two")]` ?

IHP/View/CSSFramework.hs
```diff
@@ -57,93 +61,145 @@ instance Default CSSFramework where
...
+
                     -- | Wraps the input inside a @<div class="form-group">...</div>@ (unless @disableGroup = True@)
                     formGroup :: Blaze.Html -> Blaze.Html
                     formGroup renderInner = case formField of
                         FormField { disableGroup = True } -> renderInner
-                        FormField { fieldInputId } -> [hsx|<div class={formGroupClass} id={"form-group-" <> fieldInputId}>{renderInner}</div>|]
+                        FormField { fieldInputId } -> styledFormGroup cssFramework fieldInputId renderInner
 
-                    inputClass :: (Text, Bool)
-                    inputClass = ((get #styledInputClass cssFramework) formField, True)
```
amitaibu, line 81: I believe `(get #styledInputClass cssFramework)` is redudent here, we can call `styledInputClass` directly. Anway, moved it into the `styled-*` functions.

IHP/View/CSSFramework.hs
mpscholten: Try to collapse this to `{validationResult}{helpText}`. Does that remove the spaces?

IHP/View/CSSFramework.hs
mpscholten: If you follow https://ihp.digitallyinduced.com/Guide/form.html#select-inputs you should be able to get it working with this instance:

```haskell
instance CanSelect (Int, Text) where
    type SelectValue (Int, Text) =  Int
    selectValue (int, text) = int
    selectLabel (int, text) = text
```

IHP/View/CSSFramework.hs
amitaibu: Removes one space, but one more to go :)

image

IHP/View/CSSFramework.hs
amitaibu: Thanks. I'll create a follow up PR to add it to the docs

IHP/View/CSSFramework.hs
amitaibu: No, I was wrong. When calling it directly it skips the late-binding. So I think this should also be fixed in the Pagination/ Breadcrumbs, when one `styled` is calling the other

amitaibu:
Update: Overhauled code is now looking the same as the original one

image

mpscholten:
Nice, good job so far! :)

amitaibu:
Fields with TW

image

IHP/View/CSSFramework.hs
amitaibu: @mpscholten should I add spaces to the expected result, or do you have a better idea?

amitaibu:
@mpscholten Last thing for this PR is - https://github.com/digitallyinduced/ihp/pull/1224#discussion_r753843651

IHP/View/CSSFramework.hs
mpscholten: In that case let's just add the space to the test case 👍

IHP/View/CSSFramework.hs
```diff
@@ -433,7 +433,7 @@ tailwind = def
                     placeholder={placeholder}
                     id={fieldInputId}
                     class={classes [inputClass, (inputInvalidClass, isJust validatorResult), (fieldClass, not (null fieldClass))]}
-                    value={fieldValue}
+                    value={maybeValue}
```
amitaibu, line 436: @mpscholten I tried the Maybe value added in https://github.com/digitallyinduced/ihp/commit/8c5b9cbe5a360542b9fc95ea4032401b17440203

But it still seems to output the `value=""` attribute. Anything obvious I'm missing? :smile_cat:

mpscholten (approved):

mpscholten:
Anything to still to do or can I merge this?

amitaibu:
:+1:

Issue #1233, opened 2021-11-22
history.pushState is triggered on the same URL


amitaibu:
A bit related to https://github.com/digitallyinduced/ihp/issues/1232 , when a user will try to click several times on Submit of a form with errors, the `/CreatePost` is being pushed to the navigation history. So clicking on back once, might not be enough to return to form.

https://github.com/digitallyinduced/ihp/blob/cb7e4a2fddbb209f63f14ded75b7a5a7b3b87d49/lib/IHP/static/helpers.js#L195-L203

I believe this is a regression since I think the way `formAction` is calcualted has changed.

`this.responseURL == http://localhost:8000/CreatePost` but `formAction = /CreatePost`

amitaibu:
Closing as seems related to https://github.com/digitallyinduced/ihp/issues/1232 . PR is coming

Issue #1232, opened 2021-11-22
Reloading form with errors results with an error


amitaibu:
1. got to `/NewPost`
2. Don't fill in the title, and get validations errors
3. Since we're redirected to `/CreatePost`, when reloading page user now sees an error:

![image (18)](https://user-images.githubusercontent.com/125707/142892242-95936ec4-e0b4-4840-90d9-14752a99b1b1.png)

PR #1234, opened 2021-11-22
Prevent changing URL on form with error

amitaibu:
fixes https://github.com/digitallyinduced/ihp/issues/1232
fixes https://github.com/digitallyinduced/ihp/issues/1233


amitaibu:
Ready for review

mpscholten (approved):
Thanks :) Took me a minute to understand why this works, but that totally makes sense 👍

PR #1250, opened 2021-12-01
Ask before using DigitallyInduced cachix cache

smichel17:
Currently, `ihp-new` runs `cachix use digitallyinduced` more-or-less unconditionally.

Since that's similar in some ways to adding a 3rd party repository under a traditional apt/dnf model\*, better to ask (and notify) the user first, especially since it isn't *strictly* necessary. \**Eventually, reproducible builds should make it much safer than adding a 3rd party repo, but we're not there yet, as far as I understand.*

I'm happy to tweak the wording, which is based on the description at https://ihp.digitallyinduced.com/Guide/package-management.html#binary-cache and trying to mimic the style of the other places in the script.

### DIFF
--- a/ProjectGenerator/bin/ihp-new
+++ b/ProjectGenerator/bin/ihp-new
@@ -234,6 +234,22 @@ if ! [ -x "$(command -v make)" ]; then
 	esac
 fi
 
+if ! grep -s -q "digitallyinduced.cachix.org-1:y+wQvrnxQ+PdEsCt91rmvv39qRCYzEgGQaldK26hCKE=" ~/.config/nix/nix.conf; then
+	echo -e "\e[1mCachix doesn't seem to be configured\e[0m"
+	echo -e "IHP provides a cachix cache with binaries for all IHP packages and commonly used dependencies."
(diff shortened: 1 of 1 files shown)


ProjectGenerator/bin/ihp-new
```diff
@@ -234,6 +216,40 @@ if ! [ -x "$(command -v make)" ]; then
...
+	case "$choice" in
+	y | Y) nix-env -iA ${NIXPKGS_ATTR}.cachix ;;
+	*)
+		echo -e "\e[31mPlease install cachix manually and then re-run this program.\e[0m"
+		exit 1
+		;;
+	esac
+fi
+
+if ! grep -s -q "digitallyinduced.cachix.org-1:y+wQvrnxQ+PdEsCt91rmvv39qRCYzEgGQaldK26hCKE=" ~/.config/nix/nix.conf; then
```
smichel17, line 237: This is kind of a fragile way to check whether the cache is enabled, but I couldn't figure out anything better.

mpscholten (approved):
Thanks 👍 Good addition

PR #1255, opened 2021-12-05
Refactored AutoRefresh to use the new PGListener interface

mpscholten:
The new PGListener API is less resource hungry: Previous AutoRefresh was taking one database connection per watched table from the database pool. With PGListener it's only ever using one database connection from the database pool.


IHP/AutoRefresh.hs
```diff
@@ -213,7 +220,26 @@ gcSessions autoRefreshServer = do
...
-    readIORef autoRefreshServer >>= (\autoRefreshServer -> autoRefreshServer |> get #subscriptions |> mapM_ uninterruptibleCancel)
+-- | Returns the event name of the event that the pg notify trigger dispatches
+channelName :: ByteString -> ByteString
+channelName tableName = "did_change_" <> tableName
+
+-- | Returns the sql code to set up a database trigger
+notificationTrigger :: ByteString -> PG.Query
+notificationTrigger tableName = PG.Query $ ""
+        <> "BEGIN;\n"
+        <> "CREATE OR REPLACE FUNCTION " <> functionName <> "() RETURNS TRIGGER AS $$"
```
amitaibu, line 231: Maybe use `[trimmed||]` instead of string concatenation? Could make it a bit more readable

mpscholten: Good point 👍  will apply this

PR #1257, opened 2021-12-07
Validate input arguments passed to functions of the public DataSync api

mpscholten:
This will ease debugging when something goes wrong

### DIFF
--- a/lib/IHP/DataSync/ihp-datasync.js
+++ b/lib/IHP/DataSync/ihp-datasync.js
@@ -216,6 +216,13 @@ function initIHPBackend({ host }) {
 }
 
 export async function createRecord(table, record) {
+    if (typeof table !== "string") {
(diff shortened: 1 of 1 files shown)


lib/IHP/DataSync/ihp-datasync.js
```diff
@@ -216,6 +216,13 @@ function initIHPBackend({ host }) {
 }
 
 export async function createRecord(table, record) {
+    if (typeof table !== "string") {
+        throw new Error(`Table name needs to be a string, you passed ${JSON.stringify(table)} in a call to createRecord(${JSON.stringify(table)}, ${JSON.stringify(record, null, 4)})`);
```
jeyj0, line 220: `you passed ${...} as a first argument in a call to ...` would give more information. The variable name is not necessarily known to the user of the function.

mpscholten: I think it's mostly clear to the user. This error only happens if there's a typo or something is undefined :)

PR #1274, opened 2021-12-14
Haskell Language Server: Build with -eventlog flag

mpscholten:
For use in https://github.com/haskell/haskell-language-server/issues/2340


zacwood9:
Closing since the performance issue was fixed with #1323

Issue #1275, opened 2021-12-14
Improve auto generated migration when changing the default value of a column


mpscholten:
Currently this will generate a `DROP COLUMN; ADD COLUMN` sequence instead of a `ALTER COLUMN`.

oclbdk:
Would it make sense to also handle `DROP DEFAULT`?

mpscholten:
Yes, would be useful 👍

PR #1277, opened 2021-12-16
Handle default value changes in migrations

oclbdk:
My attempt at addressing #1275.

I tested in ghci:

```
:l Test/IDE/CodeGeneration/MigrationGenerator.hs
hspec tests
```

### DIFF
--- a/IHP/IDE/SchemaDesigner/Parser.hs
+++ b/IHP/IDE/SchemaDesigner/Parser.hs
@@ -546,6 +546,7 @@ alterType = do
 
 -- | ALTER TABLE users ALTER COLUMN email DROP NOT NULL;
 --  ALTER TABLE users ALTER COLUMN email SET NOT NULL;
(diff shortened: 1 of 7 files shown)


IHP/IDE/SchemaDesigner/Parser.hs
```diff
@@ -546,6 +546,7 @@ alterType = do
 
 -- | ALTER TABLE users ALTER COLUMN email DROP NOT NULL;
 --  ALTER TABLE users ALTER COLUMN email SET NOT NULL;
+--  ALTER TABLE users ALTER COLUMN email SET DEFAULT 'value';
```
mpscholten, line 549: Could you add a test case for this query in the Parser and Compiler tests? While the new branches are called from the migration tests, it would still be useful to have a test in the parser and compiler as well 👍

mpscholten (commented):
Looks really good 👍 great job!

mpscholten (approved):
Perfect, thanks :)

PR #1278, opened 2021-12-19
Handle dropping default values in migrations

oclbdk:
Follow-up to #1275


oclbdk:
I also added a test case to check when a default value is added, just in case!

mpscholten (approved):
Perfect, thanks! :)

Issue #1198, opened 2021-11-06
IHP Job Dashboard should use the normal IHP Pagination


mpscholten:
Right now the IHP Job Dashboard [implements it's own pagination functionality](https://github.com/digitallyinduced/ihp/blob/master/IHP/Job/Dashboard/Utils.hs#L8). The pagination module was added to IHP after the job dashboard.

[To simplify things it would be nice if the Job Dashboard's pagination is replaced with the standard IHP pagination functionality.](https://github.com/digitallyinduced/ihp/blob/master/IHP/Job/Dashboard/View.hs#L114)

oclbdk:
I'd like to confirm whether an API-breaking change is necessary and get thoughts on a proposal for jobs that implement `TableViewable`.

I numbered my questions with "**Q#**" to make them easier to find.

**Q1.** How do the `TableViewable` jobs get rendered? `makeDashboardSectionFromTableViewable` doesn't seem to be called anywhere.

# Existing Functions

* The `getPage` pagination function for `TableViewable`: https://github.com/digitallyinduced/ihp/blob/master/IHP/Job/Dashboard/Types.hs#L52
* The `paginateWithOptions` IHP pagination function: https://github.com/digitallyinduced/ihp/blob/master/IHP/Pagination/ControllerFunctions.hs#L90

# Discrepancy

The legacy `getPage` interface is expected to return results with an arbitrary implementation (not necessarily a query builder).

The `paginateWithOptions` function receives a query builder and returns a paginated form for the caller to `fetch`.

# Proposal

**Q2.** Would it make sense to replace `getPage` in `TableViewable` with an interface based around query builders and pagination options?

```hs
class TableViewable a where

    ...

    -- remove the following

    getPage :: (?context :: ControllerContext, ?modelContext :: ModelContext) => Int -> Int -> IO [a]

    -- add the following

    queryForListPage :: forall controller table queryBuilderProvider joinRegister .
        ( ?context :: ControllerContext
        , ?modelContext :: ModelContext
        , KnownSymbol table
        , HasQueryBuilder queryBuilderProvider joinRegister) =>
        -> queryBuilderProvider table

    paginationOptionsForListPage :: IHP.Pagination.Types.Options
    paginationOptionsForListPage = IHP.Pagination.ControllerFunctions.defaultPaginationOptions
```

## Index View

**Q3.** Should we change the `getIndex` function to also work with query builders so it matches the interface for the list page?for the list page?

[...]

PR #1294, opened 2021-12-28
Use standard `renderPagination` in list page for Jobs Dashboard

oclbdk:
Addresses #1198

# Test Plan

I tested locally by creating 65 job instances in a sample project and checking that the pagination renders as expected for various page and window sizes:

Screen Shot 2021-12-28 at 2 35 54 PM

## Cases

* No pagination is rendered when all items fit on a single page.
* Only the necessary number of pages are listed when `window size > total pages`.
* Only the specified number of items are rendered per page when selecting the "items per page" dropdown.
* Clicking on an arbitrary page number navigates to that page.
* Clicking on the "previous" and "next" arrows navigates to the associated pages.

# QueryBuilder / Limitations

There's still redundancy with `IHP.Pagination.ControllerFunctions.paginateWithOptions`.

I tried to use that function but couldn't figure out how to initialize a `QueryBuilder` with a `Text`-valued table name. This isn't possible, right? Since the static types from the application aren't accessible to the IHP library.

I'm wondering if it makes sense to duplicate the validation logic here? In particular, the logic to constrain the page size.

### DIFF
--- a/IHP/Job/Dashboard.hs
+++ b/IHP/Job/Dashboard.hs
@@ -38,6 +38,7 @@ import IHP.ControllerPrelude
 import Unsafe.Coerce
 import IHP.Job.Queue ()
 import IHP.RouterPrelude hiding (get, tshow, error, map, putStrLn, elem)
+import IHP.Pagination.Types
 import qualified Database.PostgreSQL.Simple as PG
 import qualified Database.PostgreSQL.Simple.Types as PG
 import qualified Database.PostgreSQL.Simple.FromField as PG
@@ -176,11 +177,14 @@ instance JobsDashboard '[] where
 
     listJob = error "listJob: Requested job type not in JobsDashboard Type"
     listJob' _ = do
-        let page = fromMaybe 1 $ param "page"
-            table = param "tableName"
-        numberOfPages <- numberOfPagesForTable table 25
-        jobs <- queryBaseJobsFromTablePaginated table (page - 1) 25
-        render $ HtmlView $ renderBaseJobTablePaginated table jobs page numberOfPages
+        let table = param "tableName"
+            options = defaultPaginationOptions
+            page = paramOrDefault 1 "page"
+            pageSize = paramOrDefault (maxItems options) "maxItems"
+        totalItems <- totalRecordsForTable table
+        jobs <- queryBaseJobsFromTablePaginated table (page - 1) pageSize
+        let pagination = Pagination { currentPage = page, totalItems, pageSize, window = windowSize options }
+        render $ HtmlView $ renderBaseJobTablePaginated table jobs pagination
         render $ EmptyView
 
     viewJob = error "viewJob: Requested job type not in JobsDashboard Type"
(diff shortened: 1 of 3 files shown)


IHP/Job/Dashboard.hs
```diff
@@ -176,11 +177,14 @@ instance JobsDashboard '[] where
...
-        render $ HtmlView $ renderBaseJobTablePaginated table jobs page numberOfPages
+        let table = param "tableName"
+            options = defaultPaginationOptions
+            page = paramOrDefault 1 "page"
+            pageSize = paramOrDefault (maxItems options) "maxItems"
+        totalItems <- totalRecordsForTable table
+        jobs <- queryBaseJobsFromTablePaginated table (page - 1) pageSize
+        let pagination = Pagination { currentPage = page, totalItems, pageSize, window = windowSize options }
+        render $ HtmlView $ renderBaseJobTablePaginated table jobs pagination
         render $ EmptyView
```
mpscholten, line 188: The `render $ EmptyView` is unreachable. Can it be removed?

```suggestion
```

IHP/Job/Dashboard/Utils.hs
```diff
@@ -7,10 +7,14 @@ import qualified Database.PostgreSQL.Simple.Types as PG
...
 numberOfPagesForTable table pageSize = do
-    (PG.Only totalRecords : _) <- sqlQuery
-        (PG.Query $ cs $ "SELECT COUNT(*) FROM " <> table)
-        ()
+    totalRecords <- totalRecordsForTable table
     pure $ case totalRecords `quotRem` pageSize of
         (pages, 0) -> pages
         (pages, _) -> pages + 1
 
+totalRecordsForTable :: (?modelContext::ModelContext) => Text -> IO Int
```
mpscholten, line 15:
```suggestion
totalRecordsForTable :: (?modelContext :: ModelContext) => Text -> IO Int
```

IHP/Job/Dashboard/Utils.hs
```diff
@@ -7,10 +7,14 @@ import qualified Database.PostgreSQL.Simple.Types as PG
...
+totalRecordsForTable table = do
```
mpscholten, line 16: Next to `sqlQuery` we also have [`sqlQueryScalar`](https://ihp.digitallyinduced.com/api-docs/IHP-ModelSupport.html#v:sqlQueryScalar) which only expects a single scalar result value (e.g. a single Integer like here). Using this function we could write this a bit shorter:

```haskell
totalRecordsForTable table = sqlQueryScalar (PG.Query $ cs $ "SELECT COUNT(*) FROM " <> table) ()
```

mpscholten:
Good job on the PR again :) Thanks! 👍 

> I tried to use that function but couldn't figure out how to initialize a QueryBuilder with a Text-valued table name. This isn't possible, right? Since the static types from the application aren't accessible to the IHP library.

Yes exactly. The QueryBuilder uses a lot of type magic to ensure that only valid columns etc. are used. This makes it very hard to use in these dynamic cases.

> I'm wondering if it makes sense to duplicate the validation logic here? In particular, the logic to constrain the page size.

The dashboard is designed for internal use only. So I'm fine if we just leave this out.

zacwood9 (approved):
Looks great! Thanks

mpscholten (approved):

PR #1297, opened 2021-12-29
Support types besides strings in DataSync's filterWhere

mpscholten:


lib/IHP/DataSync/ihp-querybuilder.js
```diff
@@ -82,4 +82,18 @@ function query(table) {
...
+        return { tag: 'TextValue', contents: value };
+    } else if (typeof value === "number") {
+        return { tag: 'IntValue', contents: value };
+    } else if (typeof value === "boolean") {
+        return { tag: 'BoolValue', contents: value };
+    } else if (value === null || value === undefined) {
+        return { tag: 'NullValue' };
+    }
+
+    throw new Error('Could no transform JS value to DynamicValue');
```
amitaibu, line 96: Maybe indicate which values are valid?

```suggestion
    throw new Error('Could no transform JS value to DynamicValue (Text, Int, Bool)');
```

amitaibu: btw, do we support Float?

mpscholten: Good point 👍 will add this.

Floats are not yet implemented

mpscholten: Created an issue for Floats here https://github.com/digitallyinduced/ihp/issues/1298

PR #1302, opened 2022-01-03
Add prefixes for alpine.js support

kodeFant:
These seems to be all the prefixes needed to fully support alpine.js according to the [documentation](https://alpinejs.dev/start-here). All the other special syntax is inside strings.

Example 
```html
<div x-data="{ count: 0 }">
    <button x-on:click="count++">Increment</button>
 
    <span x-text="count"></span>
</div>
```

### DIFF
--- a/IHP/HSX/Parser.hs
+++ b/IHP/HSX/Parser.hs
@@ -178,8 +178,12 @@ hsxAttributeName = do
         isValidAttributeName name =
             "data-" `Text.isPrefixOf` name
             || "aria-" `Text.isPrefixOf` name
+            -- Prefix for HTMX support
             || "hx-" `Text.isPrefixOf` name
-            || "hx-" `Text.isPrefixOf` name
+            -- Prefixes for Alpine.js support
+            || "x-" `Text.isPrefixOf` name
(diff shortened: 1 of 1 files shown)


kodeFant:
As discussed on Slack regarding what are strictly legal characters for attributes according to the HTML spec, I tried to find whether `@` (U+0040) or `:` (U+003A) are legal characters for attribute name: 
https://html.spec.whatwg.org/multipage/syntax.html#attributes-2

As far as I can understand of it, `@` (U+0040) is not strictly legal in an attribute, so that can be taken into consideration if that's a dealbreaker for supporting it.

mpscholten:
[Could you also add a test case?](https://github.com/digitallyinduced/ihp/blob/master/Test/HSX/ParserSpec.hs)

kodeFant:
Will do!

kodeFant:
I just sat down and attempted to write test cases for stuff like `<div x-data="{ count: 0 }">` and realized that not only the illegal characters are the issue, but also the syntax with brackets {}. 

This conflicts so much with HSX's haskell interpolation, so I came to the conclusion I don't want this badly enough :)

PR #1305, opened 2022-01-06
Fixed data editor errors when trying to edit the schema_migrations table

mpscholten:
@Montmorency this should fix the issue below. Can you review?
image (38)


Montmorency:
Looks like I still get a syntax error when trying to select `schema_migrations` table in the IDE `syntax error at or near \"OFFSET"`".

PR #1310, opened 2022-01-09
Tailwind 3 support

kodeFant:
When creating a new IHP project and installing Tailwind, Tailwind 3 will be installed.

There is a slight difference in the configuration file, where the `contents` and `options` are now moved up in the hierarchy and the tailwind cli throws an error if using the old structure.

Other than that, there aren't any incompatability issues with Tailwind 3 as I am following this guide with this modification in two fresh projects now :+1: 

CC @amitaibu

### DIFF
--- a/Guide/tailwindcss.markdown
+++ b/Guide/tailwindcss.markdown
@@ -66,12 +66,10 @@ module.exports = {
     content: [
         "../Web/View/**/*.hs",
     ],
-    options: {
-        safelist: [
-        // Add here custom class names. Since we're using TW's jit (Just-In-
-        // Time), `safelist` must be full class names, and not regex.
-        ],
-    },
+    safelist: [
+        // Add custom class names.
+        // https://tailwindcss.com/docs/content-configuration#safelisting-classes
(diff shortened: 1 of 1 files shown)


Guide/tailwindcss.markdown
```diff
@@ -63,17 +63,15 @@ module.exports = {
         extend: {
         },
     },
-    purge: {
-        content: [
-            "../Web/View/**/*.hs",
-        ],
+    content: [
+        "../Web/View/**/*.hs",
+    ],
```
amitaibu, line 68: Thanks. Let's change also the `safelist` - which needs to be outside of options. Also the comment is now outdated. It can be something like - https://github.com/Gizra/drupal-starter/blob/585c7d020b8adc27e5fd0e686df8c3241c89d094/web/themes/custom/server_theme/tailwind.config.js#L21-L24

kodeFant: Done!

amitaibu (approved):
Great, thanks

PR #1331, opened 2022-01-24
Add to Guide on Using ihp-datasync with pure JS

s0kil:


Guide/realtime-spas.markdown
```diff
@@ -686,8 +686,59 @@ DELETE FROM todos WHERE id = "66cc037e-5729-435c-b507-a17492fe44f4"
 ```
 
 
+## Using DataSync library with pure JavaScript
```
mpscholten, line 689: I think `pure` could be confusing here, maybe `without a bundler` is more precise?

```suggestion
## Using DataSync library without a bundler
```

mpscholten (commented):
really nice addition 👍  didn't expect this to be that easy :)

mpscholten (approved):

PR #1334, opened 2022-01-24
Add `deleteRecordByIds`

amitaibu:
I've also renamed `model` to `record`, to keep consistency. However, I'm getting this error now:

Selection_999(623)

### DIFF
--- a/Guide/database.markdown
+++ b/Guide/database.markdown
@@ -567,6 +567,8 @@ do
     deleteRecord user
 ```
 
+Or pass only the ID with `[`deleteRecordById`](https://ihp.digitallyinduced.com/api-docs/IHP-ModelSupport.html#v:deleteRecordById).
+
 This will execute:
 
 ```sql
@@ -583,10 +585,14 @@ do
     deleteRecords users
 ```
 
+Or pass only the ID with `[`deleteRecordByIds`](https://ihp.digitallyinduced.com/api-docs/IHP-ModelSupport.html#v:deleteRecordByIds).
+
 This will execute:
 
 ```sql
 DELETE FROM users WHERE id IN (...)
+
+
 ```
 
 ### Deleting all records
--- a/IHP/ModelSupport.hs
+++ b/IHP/ModelSupport.hs
@@ -605,12 +605,6 @@ deleteRecordById id = do
 deleteRecords :: forall record id. (?modelContext :: ModelContext, Show id, Table record, HasField "id" record id, ToField id) => [record] -> IO ()
 deleteRecords records =
     deleteRecordByIds @record (ids records)
-
-    -- do
-    -- let theQuery = "DELETE FROM " <> tableName @record <> " WHERE id IN ?"
-    -- let theParameters = PG.Only (PG.In (ids records))
-    -- sqlExec (PG.Query . cs $! theQuery) theParameters
-    -- pure ()
 {-# INLINABLE deleteRecords #-}
 
 -- | Like 'deleteRecordById' but for a list of Ids.
@@ -619,7 +613,7 @@ deleteRecords records =
 -- >>> delete projectIds
 -- DELETE FROM projects WHERE id IN ('..')
 --
-deleteRecordByIds :: forall record id. (?modelContext :: ModelContext, Show id, Table record, HasField "id" record id, ToField id) => [id] -> IO ()
+deleteRecordByIds :: forall record id. (?modelContext :: ModelContext, Show id, Table record, ToField id) => [id] -> IO ()
 deleteRecordByIds ids = do
     let theQuery = "DELETE FROM " <> tableName @record <> " WHERE id IN ?"
     let theParameters = (PG.Only (PG.In ids))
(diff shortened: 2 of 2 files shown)


IHP/ModelSupport.hs
```diff
@@ -608,12 +603,29 @@ deleteRecordById id = do
...
+    -- pure ()
+{-# INLINABLE deleteRecords #-}
+
+-- | Like 'deleteRecordById' but for a list of Ids.
+--
+-- >>> let projectIds :: [ Id Project ] = ...
+-- >>> delete projectIds
+-- DELETE FROM projects WHERE id IN ('..')
+--
+deleteRecordByIds :: forall record id. (?modelContext :: ModelContext, Show id, Table record, HasField "id" record id, ToField id) => [id] -> IO ()
```
mpscholten, line 622: The mentioned type error is caused by the unused type variable `record` here :) As there's only the ids and no record passed to this function, GHC cannot figure out the type here.

Removing all mentions of `record` in the type signature will fix the issue:
```suggestion
deleteRecordByIds :: forall id. (?modelContext :: ModelContext, Show id, ToField id) => [id] -> IO ()
```

amitaibu: It has `tableName @record` in the function, so I think it's still needed

mpscholten: Good point 👍  missed that

IHP/ModelSupport.hs
```diff
@@ -578,14 +578,9 @@ logQuery query parameters time = do
...
-deleteRecord model = do
-    let condition = primaryKeyCondition model
-    let whereConditions = condition |> map (\(field, _) -> field <> " = ?") |> intercalate " AND "
-    let theQuery = "DELETE FROM " <> tableName @model <> " WHERE " <> whereConditions
-    let theParameters = map snd condition
-    sqlExec (PG.Query . cs $! theQuery) theParameters
-    pure ()
+deleteRecord :: forall record id. (?modelContext :: ModelContext, Show id, Table record, HasField "id" record id, ToField id) => record -> IO ()
+deleteRecord record =
+    deleteRecordById (get #id record)
```
mpscholten, line 583: Can you try passing the `record` type argument explicit here? This might help with the error

```
deleteRecordById @record (get #id record)
```

mpscholten:
We should also add the new function to the Guide here https://ihp.digitallyinduced.com/Guide/database.html#delete :)

IHP/ModelSupport.hs
```diff
@@ -594,10 +589,10 @@ deleteRecord model = do
 -- >>> delete projectId
 -- DELETE FROM projects WHERE id = '..'
 --
-deleteRecordById :: forall model id. (?modelContext :: ModelContext, Show id, Table model, HasField "id" model id, ToField id) => id -> IO ()
+deleteRecordById :: forall record id. (?modelContext :: ModelContext, Show id, Table record, ToField id) => id -> IO ()
 deleteRecordById id = do
```
amitaibu, line 593: This now compiles. However, I tried to cut down a bit more on code and have 

```haskell
deleteRecordById :: forall record id. (?modelContext :: ModelContext, Show id, Table record, ToField id) => id -> IO ()
deleteRecordById id = deleteRecordByIds [ids]
```

But maybe that's not possible, since we don't use the `record` - but can't remove it?

image

Guide/database.markdown
```diff
@@ -567,6 +567,8 @@ do
     deleteRecord user
 ```
 
+Or pass only the ID with `[`deleteRecordById`](https://ihp.digitallyinduced.com/api-docs/IHP-ModelSupport.html#v:deleteRecordById).
```
mpscholten, line 570: I'd actually put this under it's own headline `Deleting a single record by id`. Then it's easy to find when using the navigation at the top of the page :)

mpscholten (approved):

IHP/ModelSupport.hs
amitaibu: @mpscholten do you think the above is possible?

amitaibu:
Sorry, merged too quickly. We now get errors when using:

1) `deleteRecordById`

image


2) `deleteRecordByIds` 

Selection_999(628)

amitaibu:
Seems `deleteRecordById` didn't work also before this PR. Maybe there's some type hinting we're expected to add when calling this func, but I couldn't find how.

mpscholten:
Just pushed a fix via https://github.com/digitallyinduced/ihp/commit/b0fadac9ae06b7d7fb9c312f6f8e3770bc887b51

amitaibu:
Thanks! How did you get that fix? Is that from your knowledge - or is there some Haskel trick to get that?

mpscholten:
I saw that the `record` and the `id` type variables where not connected in any way. But the `record` can always be deduced when you have a `Id' "some_table"` type. So I replaced `id` with a more specific `Id' table` and then added a constraint `record ~ GetModelByTableName table` (`GetModelByTableName` is a type level function that returns e.g. `User` when called `GetModelByTableName "users"`). Then I fixed everything GHC mentioned after that and simplified a bit.

PR #1337, opened 2022-01-26
Add CustomMiddleware option for user-provided WAI middleware

zacwood9:
The PR adds a `CustomMiddleware` option to `FrameworkConfig`.  By providing a middleware here the user can run any arbitrary WAI middleware, providing a real "escape-hatch" for IHP.  It would even be possible, for example, to combine a Servant app or Yesod app with an IHP app using this option. Further examples are provided in the docs.


mpscholten (approved):
👍 thanks!

PR #1351, opened 2022-02-08
Jannis/datasync api additions: `or`, `and`, `where`, `whereNot`

jeyj0:
Allows the following api to be used:

```javascript
query('table')
    .where('a', a)
    .whereNot('b', b)
    .or(where('c', c).where('d', d))
    .and(where('e', e).or(where('f', f)))
    .where('g', g)
    .fetch()
```

`filterWhere` is still kept, in fact `where` is just an alias to that function. `whereNot` is added as an alias to the (also added) `filterWhereNot` function.

### DIFF
--- a/Guide/realtime-spas.markdown
+++ b/Guide/realtime-spas.markdown
@@ -611,12 +611,32 @@ const todos = await query('todos')
 // WHERE id = 'd94173ec-1d91-421e-8fdc-20a3161b7802'
 ```
 
-##### Chaining conditions
+##### Other Operators
+The above syntax is actually a shorthand for:
+
+```javascript
+const todos = await query('todos')
+        .where(eq('id', 'd94173ec-1d91-421e-8fdc-20a3161b7802'))
+        .fetch()
+
+// SQL:
+// SELECT * FROM todos
+// WHERE id = 'd94173ec-1d91-421e-8fdc-20a3161b7802'
+```
+
+If you want to use a different operator than `=`, you can use the following functions:
+- `notEq` for `<>` (shorthand `whereNot`)
+- `lessThan` for `<` (shorthand `whereLessThan`)
+- `lessThanOrEqual` for `<=` (shorthand `whereLessThanOrEqual`)
+- `greaterThan` for `>` (shorthand `whereGreaterThan`)
+- `greaterThanOrEqual` for `>=` (shorthand `whereGreaterThanOrEqual`)
+
+##### Multiple conditions
 
 You can chain as many conditions as you want - they will be `AND`ed in the resulting SQL query:
 
 ```javascript
-const todos = await('todos')
+const todos = await query('todos')
         .where('title', 'test')
         .where('id', 'd94173ec-1d91-421e-8fdc-20a3161b7802')
         .fetch()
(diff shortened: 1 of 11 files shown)


Guide/realtime-spas.markdown
```diff
@@ -597,6 +597,81 @@ const todos = await query('todos').fetchAndRefresh(callback);
 
 The `fetchAndRefresh` function is using a websocket to be notified about any changes to the selected data set. [It's using IHP's `DataSubscription` API.](https://github.com/digitallyinduced/ihp/blob/master/lib/IHP/static/vendor/ihp-datasync.js) For more fine grained control you can use the `DataSubscription` API directly instead of relying on `fetchAndRefresh`.
 
+#### Filtering
+
+Filtering records works similar to normal IHP. In JavaScript you have the choice between `filterWhere` and `where` - the two function work identically.
```
mpscholten, line 602: To simplify I'd decide on one version here (IMO `where` is better as it's shorter), and then not mention the alternative anywhere in the docs (maybe only mention it as a legacy version for keeping b.c.).

E.g. this sentence here

>  If you want to check for inequality instead of equality of a column value, you can use the `filterWhereNot` and `whereNot` functions. They work identically to their `filterWhere` and `where` counterparts and can be used in any situation those functions can be used in.

Could be simplified if it was just using one variant of the functions.

mpscholten:
Some thoughts on the overall API:

The `.and` and `.or` methods look a bit unclear to me. It's not always clear what the operands are. A design with clear operands could look like this:

```javascript
query('records')
     .where(
        or(
            eq("is_public", true),
            eq("user_id", getCurrentUserId()
        )
    )
```

In that case writing `.where('field', value)` is just a shortcut for `.where(eq('field', value))`.

Then we could also later allows syntax like this:

```javascript
query('records')
     .where({
        isPrivate: true,
        userId: getCurrentUserId()
    })

// equivalent of `.where("isPrivate", true).where("userId", getCurrentUserId())`
```


Basically `where` would have type like this:

```
data ConditionLike = (Column, Value) | KeyValueObject | Condition
where :: QueryBuilder -> ConditionLike -> QueryBuilder
```

mpscholten:
Also how does this deal with `<`, `>`, `<=`, `>=` operators?

Guide/realtime-spas.markdown
jeyj0: Makes sense. I'll make `where` and `whereNot` the default functions, aliasing the old `filterWhere` and `filterWhereNot` functions to those (instead of the way it currently is) and update the docs :+1:

jeyj0:
> Some thoughts on the overall API:
>
> The .and and .or methods look a bit unclear to me. It's not always clear what the operands are. A design with clear operands could look like this:

It would've been great to receive that feedback when I asked for it on Friday, so we could've talked about it before I implement the current version... I did propose your example as an option then, you supported the current version on Friday.

In my opinion, the current API reads a lot better, as you never have to "backtrack" while reading (the order of the functions is the same as how you would talk). The equivalent of your first example would be
```javascript
query('records')
     .where("is_public", true)
     .or(where("user_id", getCurrentUserId()))
```

which is both how you would say it, and how the SQL reads.

With your version, you have to first read one of the cases, then go back to read the `or` (since it could also be `and`) and then read the second case.

Both ways have advantages and disadvantages.

Your other, future API suggestion is also very similar to one I posted on Friday and is compatible with both the now-implemented API and the first one you now suggested.

> Also how does this deal with <, >, <=, >= operators?

Other operators weren't supported in the previous API, so it wasn't much of a thought yet, though custom `whereGreaterThan` etc. functions are an option and would read quite well, I think.

mpscholten:
> In my opinion, the current API reads a lot better, as you never have to "backtrack" while reading (the order of the functions is the same as how you would talk). The equivalent of your first example would be

Agree, that example is better. In fact rails actually does it the same way.

Maybe we can combine both approaches? Using the `eq`, `or` and `and` functions as the building blocks and then still having the `.or` methods?


> Other operators weren't supported in the previous API, so it wasn't much of a thought yet, though custom whereGreaterThan etc. functions are an option and would read quite well, I think.

[DataSync technically already supports these operators, so it's mainly adding it to the JS.](https://github.com/digitallyinduced/ihp/blob/master/IHP/DataSync/DynamicQueryCompiler.hs#L88) I think when we're now re-designing this query api it makes sense to also deal with those other operators.

jeyj0:
> Maybe we can combine both approaches? Using the `eq`, `or` and `and` functions as the building blocks and then still having the .or methods?

That should work.

> [DataSync technically already supports these operators, so it's mainly adding it to the JS.](https://github.com/digitallyinduced/ihp/blob/master/IHP/DataSync/DynamicQueryCompiler.hs#L88) I think when we're now re-designing this query api it makes sense to also deal with those other operators.

Agreed, would have done it next. I'll make a larger PR then.

jeyj0:
Check `lib/IHP/DataSync/ihp-querybuilder.test.js` to see all the new API options in action. The previous implementation works 100% the same way. The docs should give an overview of the options too.

Some further improvements are added to our internal ticketing. All of those will require changes on the Haskell side too (I think).

mpscholten (approved):
👍

PR #1362, opened 2022-02-12
Fixed build of ormolu on M1 computers

mpscholten:
Fix based on https://gist.github.com/Gabriel439/61be09ed5c223c17e5a9268323afa223

You can try it with this in `default.nix`:
```nix
let
    ihp = builtins.fetchGit {
        url = "https://github.com/digitallyinduced/ihp.git";
        rev = "71c4966967fe0743d5bbace1a31d32a651fd4f12";
        allRefs = true;
    };
```

### DIFF
--- a/NixSupport/make-nixpkgs-from-options.nix
+++ b/NixSupport/make-nixpkgs-from-options.nix
@@ -63,6 +63,7 @@ let
               manualOverrides
 
               (self: super: { haskell-language-server = pkgs.haskell.lib.appendConfigureFlag super.haskell-language-server "--enable-executable-dynamic"; })
+              (self: super: { ormolu = if pkgs.system == "aarch64-darwin" then pkgs.haskell.lib.overrideCabal super.ormolu (_: { enableSeparateBinOutput = false; }) else super.ormolu; })
             ];
           };
         }


zacwood9:
Having the same issue here. Any solutions? @mpscholten

mpscholten:
[The error is happening during detection of available controllers for usage in the dev tools](https://github.com/digitallyinduced/ihp/blob/master/IHP%2FIDE%2FToolServer.hs#L142). In the end it's just calling [`System.Directory.listDirectory`](https://github.com/digitallyinduced/ihp/blob/cb7e4a2fddbb209f63f14ded75b7a5a7b3b87d49/IHP/IDE/ToolServer/Helper/Controller.hs#L69). So there shouldn't be much that can go wrong.

Some ideas:
- Can you check that the `Web/` directory is there (I guess you checked already, just want to double confirm)?
- A race condition might be somehow switching the working directory to something else (but only on M1), and then the working directory isn't the project root dir anymore. We could check that by adding a print statement before the `listDirectory` call
- A bug in `System.Directory`. I have quickly skimmed through the github repo there and haven't found any reports, so it seems unlikely

zacwood9:
The project works when creating `Web` along with some other required files. Would it be possible to get this change into ihp-boilerplate or will it have to wait until the next release? I'm currently starting a lot of new projects so manually creating or copying over needed files is a lot of work :)

mpscholten:
We should make this a new release. Likely we could Tag a new release still this week

eftychis:
@mpscholten: The demo works for me with this patch. (nix 2.7.0) Is there a new release on the near horizon? (Seems like the last release was in January.)

mpscholten:
Yes we can put put the next release very soon :)

PR #1376, opened 2022-03-03
add sqlQueryScalarMaybe

jeyj0:
What do you think, @mpscholten?

I'll have to either use this, or catch the exception - which definitely isn't as nice looking, or understandable. Alternatively, I could change the SQL, but it would get overly complex as well.

### DIFF
--- a/IHP/ModelSupport.hs
+++ b/IHP/ModelSupport.hs
@@ -430,11 +430,11 @@ sqlQueryScalar theQuery theParameters = do
 --
 -- __Example:__
 --
--- > usersCount <- sqlQueryScalarMaybe "SELECT COUNT(*) FROM users"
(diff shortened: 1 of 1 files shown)


IHP/ModelSupport.hs
```diff
@@ -426,6 +426,27 @@ sqlQueryScalar theQuery theParameters = do
...
+    result <- measureTimeIfLogging
+        (withDatabaseConnection \connection -> enhanceSqlError theQuery theParameters do
+            PG.query connection theQuery theParameters
+        )
+        theQuery
+        theParameters
+    pure case result of
+        [] -> Nothing
+        [PG.Only result] -> Just result
+        _ -> error "sqlQueryScalar: Expected a scalar result value or nothing"
```
mpscholten, line 447:
```suggestion
        _ -> error "sqlQueryScalarOrNothing: Expected a scalar result value or an empty result set"
```

PR #1392, opened 2022-03-08
IHP GraphQL Standalone

mpscholten:


ihp.nix
```diff
@@ -134,8 +136,16 @@ mkDerivation {
     unagi-chan
     with-utf8
     ihp-hsx
+    jwt
   ];
   license = lib.licenses.mit;
+  preBuild = ''
```
s0kil, line 142: This stage currently fails when I check out these changes locally and perform a build:
I have worked around that by escaping nix sandboxing: `nix-shell --option build-use-sandbox false` since otherwise npm will fail due to no internet access. Is this the best option available for nix+npm, escaping the sandbox?

mpscholten: Yes known issue. The thin backend binaries are [also built without sandbox](https://github.com/digitallyinduced/thin-backend/blob/master/.github/workflows/build.yml#L32). We can fix this by using something like https://github.com/svanderburg/node2nix

PR #1393, opened 2022-03-09
parse EXCLUDE constraints

### LINKED ISSUE #1377: Add support for EXCLUDE constraints
See postgresql docs: https://www.postgresql.org/docs/9.4/sql-createtable.html#SQL-CREATETABLE-EXCLUDE

These constraints for example allow creating unique constraints that only apply to a subset of columns that are deferrable. (creating a unique index works for this case too, except for the deferrable feature)

jeyj0:
fixes #1377 

Do I have to do anything else, @mpscholten?

### DIFF
--- a/IHP/IDE/SchemaDesigner/Parser.hs
+++ b/IHP/IDE/SchemaDesigner/Parser.hs
@@ -141,7 +141,7 @@ parseTableConstraint = do
         lexeme "CONSTRAINT"
         identifier
     (Left <$> parsePrimaryKeyConstraint) <|>
-      (Right <$> (parseForeignKeyConstraint name <|> parseUniqueConstraint name <|> parseCheckConstraint name))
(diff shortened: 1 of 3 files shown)


IHP/IDE/SchemaDesigner/Parser.hs
```diff
@@ -172,6 +172,34 @@ parseCheckConstraint name = do
...
+    where
+        excludeElement = do
+            element <- identifier
+            space
+            lexeme "WITH"
+            space
+            operator <- parseOperator
+            pure ExcludeConstraintElement { element, operator }
+
+parseOperator = choice $ map lexeme
```
mpscholten, line 191: Maybe we can call this `parseExcludeOperator` to avoid this being confused with the general operator parsing logic (which handles infix, prefix, postfix etc.)?

```suggestion
parseExcludeOperator = choice $ map lexeme
```

mpscholten (approved):

PR #1399, opened 2022-03-10
exclude constraint using index_type

jeyj0:
Required for migrations to work, as pg_dump always sets the index type used for the exclude constraint.

### DIFF
--- a/IHP/IDE/SchemaDesigner/Parser.hs
+++ b/IHP/IDE/SchemaDesigner/Parser.hs
@@ -189,6 +189,7 @@ parseCheckConstraint name = do
 
 parseExcludeConstraint name = do
     lexeme "EXCLUDE"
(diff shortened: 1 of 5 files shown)


IHP/IDE/SchemaDesigner/Parser.hs
```diff
@@ -546,21 +547,23 @@ createIndex = do
...
     whereClause <- optional do
         lexeme "WHERE"
         expression
     char ';'
     pure CreateIndex { indexName, unique, tableName, expressions, whereClause, indexType }
 
+parseIndexType = do
+    lexeme "USING"
+
+    choice $ map (\(s, v) -> do symbol' s; pure v)
```
mpscholten, line 561: This is really nice 👍

mpscholten (approved):

Issue #1397, opened 2022-03-10
PostgreSQL Function Accept Parameters


s0kil:
Reference: https://www.postgresql.org/docs/current/sql-createfunction.html

mpscholten:
Can you share the sql function that didn't work?

s0kil:
The function starts as:
```sql
CREATE FUNCTION message_thread_id(message_to TEXT, message_from TEXT) RETURNS TEXT LANGUAGE plpgsql AS $$
```

It fails because of: https://github.com/digitallyinduced/ihp/blob/master/IHP/IDE/SchemaDesigner/Parser.hs#L569

mpscholten:
Ah got it thx 👍

PR #1402, opened 2022-03-11
Schema Parser And Compiler Support Function Parameters

s0kil:
Fixes https://github.com/digitallyinduced/ihp/issues/1397


IHP/IDE/SchemaDesigner/Parser.hs
```diff
@@ -569,7 +569,16 @@ createFunction = do
     orReplace <- isJust <$> optional (lexeme "OR" >> lexeme "REPLACE")
     lexeme "FUNCTION"
     functionName <- qualifiedIdentifier
-    lexeme "()"
+
+    let functionParameter = do
+        parameterName <- qualifiedIdentifier
+        space
+        parameterType <- qualifiedIdentifier
```
mpscholten, line 576: It would be more specific to use the `sqlType` parser here. That will accept more complex types as well (e.g. arrays)

```suggestion
        parameterType <- sqlType
```

IHP/IDE/SchemaDesigner/Types.hs
```diff
@@ -34,8 +34,8 @@ data Statement
     | CreateIndex { indexName :: Text, unique :: Bool, tableName :: Text, expressions :: [Expression], whereClause :: Maybe Expression, indexType :: Maybe IndexType }
     -- | DROP INDEX indexName;
     | DropIndex { indexName :: Text }
-    -- | CREATE OR REPLACE FUNCTION functionName() RETURNS TRIGGER AS $$functionBody$$ language plpgsql;
-    | CreateFunction { functionName :: Text, functionBody :: Text, orReplace :: Bool, returns :: PostgresType, language :: Text }
+    -- | CREATE OR REPLACE FUNCTION functionName(param1 TEXT, param2 INT) RETURNS TRIGGER AS $$functionBody$$ language plpgsql;
+    | CreateFunction { functionName :: Text, functionParameters :: [(Text, Text)], functionBody :: Text, orReplace :: Bool, returns :: PostgresType, language :: Text }
```
mpscholten, line 38: Postgres itself refers to these as `arguments` instead of `parameters` (e.g. `argname` or `argtype` in the docs). Could we rename this from parameters to arguments to be consistent with postgres?

mpscholten (approved):
Looks good, thanks :)

Issue #1426, opened 2022-04-03
Export all FileInfo accessors


s0kil:
Here: https://github.com/digitallyinduced/ihp/blob/master/IHP/Controller/FileUpload.hs#L17
We only export: `fileContent`, but we should export all accessors: https://hackage.haskell.org/package/wai-extra-3.1.8/docs/Network-Wai-Parse.html#t:FileInfo

PR #1427, opened 2022-04-03
Export all FileInfo accessors

s0kil:
Closes https://github.com/digitallyinduced/ihp/issues/1426


mpscholten (approved):
Thanks, good improvement 👍

Issue #1430, opened 2022-04-03
Add a Sitemap Builder And Generator


s0kil:
I imagine sitemap builder as:
```haskell
data Sitemap = Sitemap { links :: [SitemapLink] }
data SitemapLink = SitemapLink {  url :: Text, lastModified :: Maybe Text, changeFrequency :: Maybe SitemapChangeFrequency  }
data SitemapChangeFrequency = Always | Hourly | Daily | Weekly | Monthly | Yearly | Never
```

```haskell
posts <- query @Post |> fetch
mySitemaps = posts |> map (\post -> SitemapLink { url = urlTo $ ShowPostAction (get #id post),  lastModified = Just (get #updatedAt post) })
```

Need to think  through how to hook it up with IHP, so `/sitemap.xml` will be resolved automatically.

PR #1431, opened 2022-04-04
Sitemap Generator

s0kil:
Implements and closes https://github.com/digitallyinduced/ihp/issues/1430

### DIFF
--- a/ihp.nix
+++ b/ihp.nix
@@ -64,6 +64,8 @@
 , unagi-chan
 , with-utf8
 , ihp-hsx
+, xml-conduit
+, xml-hamlet
 }:
 mkDerivation {
   pname = "ihp";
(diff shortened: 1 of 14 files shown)


ihp.nix
```diff
@@ -134,6 +136,8 @@ mkDerivation {
     unagi-chan
     with-utf8
     ihp-hsx
+    xml-conduit
```
mpscholten, line 139: Not a fan of bringing in these heavy dependencies into IHP core. Is there another solution? Otherwise maybe we can make this a plugin?

s0kil: Ah, yes! I remembered that Blaze supports custom tags and attributes, no new deps and faster.

mpscholten:
Looks very good now 👍 Maybe we could also add a few tests for the xml generation functionality? Otherwise someone else might break it in the future

mpscholten (approved):

PR #1432, opened 2022-04-08
Adding a unitHsx block to ViewSupport.

Montmorency:
When folding or otherwise recursing over hsx blocks it is useful to have an empty hsx block. Sometimes you may want to nest a fold in an hsx block e.g. when creating many table rows from a `[TimeSlots]`:
```
 [hsx| <tr> {foldr ((<>) . renderTimeSlot) unitHsx timeSlots} </tr>|]
```
`unitHsx` is so called because  `[hsx|<p>An hsx block </p>|] <> unitHsx =  [hsx|<p>An hsx block </p>|]`.

### DIFF
--- a/IHP/ViewSupport.hs
+++ b/IHP/ViewSupport.hs
@@ -17,6 +17,7 @@ module IHP.ViewSupport
 , css
 , onClick
 , onLoad
+, unitHsx
 , theRequest
 , viewContext
 , addStyle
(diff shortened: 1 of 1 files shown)


s0kil:
Why not just use `mempty`?

mpscholten:
I also typically use `mempty` for this :)

s0kil:
Maybe we could have a better name for `mempty`?

mpscholten:
`mempty` is a pretty standard function in the haskell ecosystem so I'd rather not change it.

Montmorency:
Agreed. I started defining unitHsx and never realized mempty works just as well!

PR #1446, opened 2022-04-22
Updated the join documentation to highlight the (:.) type constructor.

Montmorency:
Not sure this is documented anywhere yet. A useful approach to custom joins using IHP datatypes.


mpscholten (approved):
Perfect. Thanks :)

PR #1448, opened 2022-04-26
Add isActiveAction to IHP.ViewSupport

s0kil:


IHP/ViewSupport.hs
```diff
@@ -156,6 +157,20 @@ isActiveController =
...
+-- | Returns @True@ when the given action matches the path of the currently executed action
+--
+-- __Example:__ The browser has requested @\/PostsAction@.
+--
+-- >>> isActiveAction PostsAction
+-- True
+--
+-- Returns @True@ because the current path is the same as the path to the action
+isActiveAction :: forall controllerAction. (?context::ControllerContext, HasPath controllerAction) => controllerAction -> Bool
+isActiveAction controllerAction =
```
mpscholten, line 169: How is this different from `isActivePath`?

mpscholten: ok makes sense. Maybe we should add this to the docs of the function 👍

IHP/ViewSupport.hs
```diff
@@ -94,7 +95,7 @@ currentViewId =
         moduleParts :: [Text]
         moduleParts = Text.splitOn "." moduleName
 
--- | Returns @True@ when the current request path matches the path given as argument.
+-- | Returns @True@ when the current request path matches the path given as argument. Takes into account the search query: `?name=value`
```
mpscholten, line 98: In haskell documentation you have to use `@` symbols for inline code like `@?name=value`

IHP/ViewSupport.hs
```diff
@@ -156,6 +157,21 @@ isActiveController =
...
 
+-- | Returns @True@ when the given action matches the path of the currently executed action
+--
+-- __Example:__ The browser has requested @\/PostsAction@.
+--
+-- >>> isActiveAction PostsAction
+-- True
+--
+-- Returns @True@ because the current path is the same as the path to the action, does not take into account the search query.
+-- Use `isActivePath` if you need to match path and search query.
```
mpscholten, line 168: same here

Guide/view.markdown
```diff
@@ -192,6 +192,10 @@ Use [`isActivePath`](https://ihp.digitallyinduced.com/api-docs/IHP-ViewSupport.h
 
 Use [`isActiveController`](https://ihp.digitallyinduced.com/api-docs/IHP-ViewSupport.html#v:isActiveController).
 
+### Check whether this view is called from a specific action
+
+Use [`isActiveController`](https://ihp.digitallyinduced.com/api-docs/IHP-ViewSupport.html#v:isActiveAction).
```
mpscholten, line 197:
```suggestion
Use [`isActiveAction`](https://ihp.digitallyinduced.com/api-docs/IHP-ViewSupport.html#v:isActiveAction).
```

PR #1478, opened 2022-06-26
Add slot parameter for custom elements

kodeFant:
To make slots for custom elements work, a `slot` parameter needs to be enabled.

Usage example:

```html
<my-paragraph>
  <span slot="my-text">Let's have some different text!</span>
</my-paragraph>
```

Source:
https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_templates_and_slots#adding_flexibility_with_slots


mpscholten (approved):
Thanks :)

PR #1486, opened 2022-07-26
Single Routing Entrypoint

jeyj0:
This PR introduces an optional `router` function for `FrontController`s, which allows handling the routing for that FrontController only in that function, or extending the `defaultRouter`, which is the previous behavior. Additionally, the function allows using `putContextRouter`, a new function that is used to modify the `ControllerContext` during routing, ~*before*~ ~*after*~ *before* `initContext` is called.

The default implementation for `router` is `router = defaultRouter`, which is 100% backwards-compatible.

An example of why this is useful and how it is used is this router implementation:

```haskell
instance FrontController WebApplication where
    ... -- the previous code assigning controllers is still required

    router additionalControllers = do
        mlocale <- optional do
            string "/"
            string "en"

        let router = defaultRouter additionalControllers

        case mlocale of
            Just locale -> case locale of
                "en" -> putContextRouter EN_UK router
                _ -> IHP.RouterPrelude.error "This should be unreachable"
            Nothing -> router
```

This handles all routes as before but also allows an optional `/en` to be added in front of all routes. In case it is added, the `ControllerContext` is modified to contain a locale (`EN_UK`), which allows the application to show in English (instead of the default German for this application).

~The `additionalControllers` are the `controllers` of the `FrontController`, and the other controllers added by IHP by default (for example for DataSync).~
The `additionalControllers` are the controllers added by IHP by default (for example for DataSync). The controllers of the `FrontController` itself are accessible via the typeclass' `controllers` function.

Since `controllers` returns a list of `RouteParser`s, it is possible to nest these custom routers using `mountFrontController` as before. This means it is possible to override all routing by implementing `router` for the `RootApplication`.

---

Since implementing the `router` function can change how routes are handled, it can also break using `pathTo`, `urlTo`, or implicit `pathTo` in HSX. In the future, `pathTo` should be made `ControllerContext`-aware so those methods can be compatible with the `router`. In the meantime, I suggest not adding this feature to the documentation yet, to avoid unnecessary confusion.

As an example, in the above application, none of `pathTo`, `urlTo`, or implicit `pathTo` in HSX would ever result in an `/en` being added to the url. By making `pathTo` `ControllerContext`-aware, it could read the current locale from context, and prepend `/en` to the path if the locale is `EN_UK`. As a workaround for this application, a `localizedPathTo` and `localizedUrlTo` will be added, which does this, but care has to be taken to use these functions instead of `pathTo` and `urlTo`, and implicit `pathTo` cannot be used.

---

**Edit:** updated the PR text to match actual behavior that I wrongly remembered.

### DIFF
--- a/IHP/ControllerSupport.hs
+++ b/IHP/ControllerSupport.hs
@@ -17,6 +17,8 @@ module IHP.ControllerSupport
 , ControllerContext
 , InitControllerContext (..)
 , runActionWithNewContext
+, newContextForAction
+, runActionWithContext
 , respondAndExit
 , ResponseException (..)
 , jumpToAction
@@ -87,9 +89,13 @@ runAction controller = do
 
     doRunAction `catches` [ Handler handleResponseException, Handler (\exception -> ErrorController.displayException exception controller "")]
 
-{-# INLINE runActionWithNewContext #-}
-runActionWithNewContext :: forall application controller. (Controller controller, ?applicationContext :: ApplicationContext, ?context :: RequestContext, InitControllerContext application, ?application :: application, Typeable application, Typeable controller) => controller -> IO ResponseReceived
-runActionWithNewContext controller = do
+{-# INLINE runActionWithContext #-}
+runActionWithContext :: forall application controller. (Controller controller, ?modelContext :: ModelContext, ?applicationContext :: ApplicationContext, ?requestContext :: RequestContext) => ControllerContext -> controller -> IO ResponseReceived
+runActionWithContext controllerContext controller = let ?context = controllerContext in runAction controller
+
+{-# INLINE newContextForAction #-}
+newContextForAction :: forall application controller. (Controller controller, ?applicationContext :: ApplicationContext, ?context :: RequestContext, InitControllerContext application, ?application :: application, Typeable application, Typeable controller) => controller -> IO (Either (IO ResponseReceived) ControllerContext)
+newContextForAction controller = do
     let ?modelContext = ApplicationContext.modelContext ?applicationContext
     let ?requestContext = ?context
     controllerContext <- Context.newControllerContext
@@ -99,13 +105,23 @@ runActionWithNewContext controller = do
 
     try (initContext @application) >>= \case
         Left (exception :: SomeException) -> do
-            case fromException exception of
+            pure $ Left $ case fromException exception of
                 Just (ResponseException response) ->
                     let respond = ?context |> get #requestContext |> get #respond
                     in respond response
                 Nothing -> ErrorController.displayException exception controller " while calling initContext"
+        Right _ -> pure $ Right ?context
+
+{-# INLINE runActionWithNewContext #-}
+runActionWithNewContext :: forall application controller. (Controller controller, ?applicationContext :: ApplicationContext, ?context :: RequestContext, InitControllerContext application, ?application :: application, Typeable application, Typeable controller) => controller -> IO ResponseReceived
+runActionWithNewContext controller = do
+    contextOrResponse <- newContextForAction controller
+    case contextOrResponse of
+        Left response -> response
         Right context -> do
-            runAction controller
+            let ?modelContext = ApplicationContext.modelContext ?applicationContext
+            let ?requestContext = ?context
+            runActionWithContext context controller
 
 -- | If 'IHP.LoginSupport.Helper.Controller.enableRowLevelSecurityIfLoggedIn' was called, this will copy the
 -- the prepared RowLevelSecurityContext from the controller context into the ModelContext.
--- a/IHP/ControllerSupport.hs
+++ b/IHP/ControllerSupport.hs
@@ -90,7 +90,7 @@ runAction controller = do
 
 applyContextSetter :: (TypeMap.TMap -> TypeMap.TMap) -> ControllerContext -> IO ControllerContext
 applyContextSetter setter ctx@ControllerContext { customFieldsRef } = do
-    modifyIORef customFieldsRef (applySetter setter)
+    modifyIORef' customFieldsRef (applySetter setter)
     pure $ ctx { customFieldsRef }
     where
         fromSetter :: (TypeMap.TMap -> TypeMap.TMap) -> TypeMap.TMap
--- a/IHP/ControllerSupport.hs
+++ b/IHP/ControllerSupport.hs
@@ -49,7 +49,7 @@ import qualified IHP.ErrorController as ErrorController
 import qualified Data.Typeable as Typeable
 import IHP.FrameworkConfig (FrameworkConfig (..), ConfigProvider(..))
 import qualified IHP.Controller.Context as Context
-import IHP.Controller.Context (ControllerContext)
+import IHP.Controller.Context (ControllerContext(ControllerContext), customFieldsRef)
 import Network.HTTP.Types.Header
 import qualified Data.Aeson as Aeson
 import qualified Network.Wai.Handler.WebSockets as WebSockets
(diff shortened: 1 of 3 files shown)


IHP/RouterSupport.hs
```diff
@@ -68,9 +69,81 @@ import qualified IHP.WebSocket as WS
...
+        => [RouteParser] -> RouteParser
+    router = defaultRouter
+
+    defaultRouter
+        :: (?applicationContext :: ApplicationContext, ?application :: application, ?context :: RequestContext)
+        => [RouteParser] -> RouteParser
+    defaultRouter additionalControllers = do
+        let allControllers = controllers <> additionalControllers
+        ioResponseReceived <- choice $ map (\r -> r <* endOfInput) allControllers
+        pure ioResponseReceived
```
mpscholten, line 146: It seems like this could be moved out of the type class into a standalone function

IHP/RouterSupport.hs
```diff
@@ -68,9 +69,81 @@ import qualified IHP.WebSocket as WS
 import GHC.TypeLits as T
 import IHP.Controller.Context
 import IHP.Controller.Param
+import qualified Data.TMap as TMap
+import qualified IHP.ApplicationContext as ApplicationContext
+
+putContextRouter :: forall value. (Typeable value) => value -> RouteParser -> RouteParser
```
mpscholten, line 75: Why can't we use the existing controller context also during routing?

Right now the controller context is created when the action is called. But it seems there's nothing preventing us from creating the controller context before routing.

Then we could just use `putContext` eveywhere, which would simplify the public API of IHP.

IHP/ControllerSupport.hs
```diff
@@ -99,13 +105,23 @@ runActionWithNewContext controller = do
 
     try (initContext @application) >>= \case
         Left (exception :: SomeException) -> do
-            case fromException exception of
+            pure $ Left $ case fromException exception of
```
mpscholten, line 108: Why is this change needed?

IHP/RouterSupport.hs
jeyj0: `putContext` is an `IO` function. So parsing would require constant lifts, which to me seemed less nice than having to use a special `putContextRouter` function for parsing.

mpscholten: Ok this makes sense 👍

IHP/ControllerSupport.hs
jeyj0: I think here, the diff is mostly confusing. What I've done is taken the logic from `runActionWithNewContext` and split it into a `newContextForAction` and `runActionWithContext`, then implementing the old `runActionWithNewContext` in terms of those two new functions. So the old function has the same signature and does the same thing. I just needed separate access to the two steps within `RouterSupport`, because I needed a way to create a `ControllerContext` with `initContext` and all other defaults in mind, apply changes to it (from the router), and then run an action with the updated `ControllerContext`.

While writing this, I am realizing that my explanation in the PR is wrong though: first, `initContext` is executed, then the changes by the `router` are applied, not the other way around. I'll have to think about which order makes more sense.

jeyj0:
As written above in a comment, currently, the order of creating the `ControllerContext` before the action is called is:
1. IHP defaults
2. `initContext`
3. `router` changes are applied

I think the more useful order is:
1. IHP defaults
2. `router` changes are applied
3. `initContext`

This way, the `router`'s changes are available in `initContext`, where any generic logic can be executed due to the current context state. For example, if from routing it is unclear which locale should be used for the app, `initContext` can use the `Accept-Language` header to detect it, and possibly redirect to a better route.

The current order does not have any benefit, as the `router` has far less flexibility in what it can do. For example, the above use case would require a change in all actions to work properly.

I will update the order of creating the `ControllerContext`.

IHP/ControllerSupport.hs
```diff
@@ -87,25 +89,61 @@ runAction controller = do
 
     doRunAction `catches` [ Handler handleResponseException, Handler (\exception -> ErrorController.displayException exception controller "")]
 
-{-# INLINE runActionWithNewContext #-}
-runActionWithNewContext :: forall application controller. (Controller controller, ?applicationContext :: ApplicationContext, ?context :: RequestContext, InitControllerContext application, ?application :: application, Typeable application, Typeable controller) => controller -> IO ResponseReceived
-runActionWithNewContext controller = do
+{-# INLINE runActionWithContext #-}
+runActionWithContext :: forall application controller. (Controller controller, ?modelContext :: ModelContext, ?applicationContext :: ApplicationContext, ?requestContext :: RequestContext) => ControllerContext -> controller -> IO ResponseReceived
+runActionWithContext controllerContext controller = let ?context = controllerContext in runAction controller
```
mpscholten, line 94: This function IMO is the same as `runAction`. Can't we then just remove this function entirely?

IHP/ControllerSupport.hs
```diff
@@ -87,25 +89,61 @@ runAction controller = do
...
+
+applyContextSetter :: (TypeMap.TMap -> TypeMap.TMap) -> ControllerContext -> IO ControllerContext
+applyContextSetter setter ctx@ControllerContext { customFieldsRef } = do
+    modifyIORef customFieldsRef (applySetter setter)
```
mpscholten, line 98: The strict version `modifyIORef'` might be better here. Otherwise `applyContextSetter` might not catch `error` calls that could be created in the router system until the context is actually used.

IHP/ControllerSupport.hs
mpscholten: > While writing this, I am realizing that my explanation in the PR is wrong though: first, initContext is executed, then the changes by the router are applied, not the other way around. I'll have to think about which order makes more sense.

As a user I'd naively expect the router context to come first and be accessible from the `initContext`

jeyj0: I have already changed it to this behavior below :+1:

IHP/ControllerSupport.hs
jeyj0: Yes, you are absolutely correct. Probably a holdover from some changes I made. Fixed!

IHP/ControllerSupport.hs
jeyj0: Agreed. Fixed.

mpscholten (approved):

mpscholten:
Sorry, I removed it because I thought not many people were using it. Additionally it caused a lot of complexity in the routing.

Here is an example of how this can be done using the request vault (a feature of wai, the underlying http server library IHP uses, it's like a simple key value object, see https://hackage.haskell.org/package/vault):

Let's assume we have a function to enable i18n in routing:
```haskell
instance FrontController WebApplication where
    controllers =
        [ startPage WelcomeAction
        -- ...
        ]

    router additionalControllers = do
        mlocale <- optional do
            string "/"
            string "en" -- <|> string "de"

        let router = defaultRouter additionalControllers
        case mlocale of
            Just locale -> case locale of
                "en" -> putContextRouter EN_UK router
                -- "de" -> putContextRouter DE_DE router
                _ -> IHP.RouterPrelude.error "This should be unreachable"
            Nothing -> putContextRouter defaultLocale router
        defaultRouter additionalControllers
```

This function allows urls such as `/en/...` and `/de/...` to be routed and also the language will be stored in the context via `putContextRouter DE_DE`.

The new way is to write a middleware that stores it into the request vault:

```haskell
import qualified Data.Vault.Lazy as Vault
import System.IO.Unsafe (unsafePerformIO)
import Network.Wai

localeVaultKey :: Vault.Key Locale
localeVaultKey = unsafePerformIO Vault.newKey
{-# NOINLINE localeVaultKey #-}

localeMiddleware :: Middleware
localeMiddleware app request respond =
    let
        locale = localeFromPathInfo request.pathInfo
        vault' = Vault.insert localeVaultKey locale request.vault
        request' = request { vault = vault' }
    in
        app request' respond

localeFromPathInfo :: [Text] -> Locale
localeFromPathInfo pathInfo =
    case pathInfo of
        ("en":_) -> EN_UK
        otherwise -> DE_DE
```

Once enabled, the locale can be access from the vault instead of the context directly:
```
-- OLD:
-- currentLocale :: _ => Locale
-- currentLocale = fromMaybe defaultLocale $ maybeFromFrozenContext @Locale
-- NEW:
currentLocale :: (?context :: ControllerContext) => Locale
currentLocale =
    Vault.lookup localeVaultKey request.vault
    |> fromMaybe (error "no locale set")
```

The middleware also needs to be activated in the `Config.hs`:

```

config :: ConfigBuilder
config = do
    option (CustomMiddleware localeMiddleware)
```

Now `putContextRouter` is not needed anymore.

stephenbenedict:
@mpscholten Wow! Thank you so much for the comprehensive response. I will try that.

stephenbenedict:
Hm, I must be missing something. Everything compiles fine, but routing with the locale prefix (e.g. /ja ) is not working (get 404).

```haskell
-- Config.hs
import Application.Helper.Localization

config :: ConfigBuilder
config = do
    option (CustomMiddleware localeMiddleware)
```

```haskell
-- Localization.hs
{-# LANGUAGE DeriveGeneric #-}
module Application.Helper.Localization
    ( Localizable(..)
    , l10n
    , l10nText
    , defaultLocale
    , currentLocale
    , localeMiddleware
    ) where

import IHP.ViewPrelude
import Generated.Types
import IHP.ControllerPrelude
import qualified Data.Vault.Lazy as Vault
import System.IO.Unsafe (unsafePerformIO)
import Network.Wai

localeVaultKey :: Vault.Key Locale
localeVaultKey = unsafePerformIO Vault.newKey
{-# NOINLINE localeVaultKey #-}

localeMiddleware :: Middleware
localeMiddleware app request respond =
    let
        locale = localeFromPathInfo request.pathInfo
        vault' = Vault.insert localeVaultKey locale request.vault
        request' = request { vault = vault' }
    in
        app request' respond

localeFromPathInfo :: [Text] -> Locale
localeFromPathInfo pathInfo =
    case pathInfo of
        ("ja":_) -> JaJp
        otherwise -> EnUs

currentLocale :: (?context :: ControllerContext) => Locale
currentLocale =
    Vault.lookup localeVaultKey request.vault
    |> fromMaybe (error "no locale set")

data Localizable
    = Heading
    | Introduction
    | Body

l10nText :: _ => Localizable -> Text
l10nText l = l10nText' l currentLocale
```

stephenbenedict:
Figured it out. Forgot that I needed to keep this in Web.FrontController:

```haskell
router additionalControllers = do
        mlocale <- optional do
            string "/"
            string "ja"
        defaultRouter additionalControllers  
```

PR #1487, opened 2022-07-30
Use GHC 9.2.4

mpscholten:


mpscholten:
[ok, can you make a PR that overrides the `hs-brotli` package for IHP apps with your fork?](https://github.com/digitallyinduced/ihp/tree/master/NixSupport/haskell-packages) after that we can also readd it to the github actions builds

PR #1489, opened 2022-08-06
Update applyImageMagick parameters

GarrisonJ:
The parameters to this function have been updated.


mpscholten (approved):
Thanks 👍 good catch!

PR #1529, opened 2022-09-19
adding some docs on using scripts from ghci.

Montmorency:
Might be useful to have some documentation on running IHP Scripts from within ghci.

### DIFF
--- a/Guide/scripts.markdown
+++ b/Guide/scripts.markdown
@@ -87,6 +87,33 @@ The `ihpDefaultConfig` is made available from the `Application.Script.Prelude` i
 with your own `Config` data structure. This is particularly useful for adjusting logging levels or
 testing new APIs that require certain configuration variables.
 
+You can define a custom configuration (or import one from Config.hs) in your Script
+with Logging set to Debug:
+
+```haskell
+...
+import qualified IHP.Log as Log
+import IHP.Log.Types
+
+
+testConfig :: ConfigBuilder
+testConfig = do
+    logger <- liftIO $ newLogger def {
+        level = Debug,
+        formatter = withTimeAndLevelFormatter,
+        destination = File "Log/App.log" (SizeRotate (Bytes (4 * 1024 * 1024)) 4) defaultBufSize
+        }
+    option logger
+
+...
+```
+
+and then run the  script from ghci:
+
+```haskell
+IHP> runScript appConfig runTestNamee
+```
+
 ## Building a script
 
 In production, you might want to build a script to a binary for performance reasons. Use make like this:
--- a/Guide/scripts.markdown
+++ b/Guide/scripts.markdown
@@ -111,7 +111,7 @@ testConfig = do
 and then run the  script from ghci:
 
 ```haskell
-IHP> runScript appConfig runTestNamee
+IHP> runScript testConfig runTestNamee
 ```
 
 ## Building a script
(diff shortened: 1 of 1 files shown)


Guide/scripts.markdown
```diff
@@ -64,6 +64,29 @@ This is made possible because of the [she-bang line](https://en.wikipedia.org/wi
...
+```
+
+and run the script from the IHP ghci command line:
+
+```haskell
+IHP> runScript ihpDefaultConfig runTestName
+```
+
+The `ihpDefaultConfig` is made available from the `Application.Script.Prelude` import but can be substituted
+with your own `Config` data structure. This is particularly useful for adjusting logging levels or
```
mpscholten, line 87: Maybe we can have an example call to `runScript` with the custom config imported as well?

Montmorency: 👍 Can add that also I rewrote a version of runScript that allows you to run Scripts with a slightly different type signature. e.g. if you want your script to be able to return a Maybe, Either, List value etc. Essentially I wanted to get some Aeson/Lens stuff working (e.g. keep trying random filters until I had accessed the right part of a json data structure) on data I fetched from the db in a script: 

```
-- | Initializes IHP and then runs the script inside the framework context allows for Script to return arbitrary monad.
runCustomScript :: (Monad m) => ConfigBuilder -> ((?modelContext :: ModelContext, ?context :: FrameworkConfig) => IO (m a)) -> IO (m a)
runCustomScript configBuilder taskMain = withUtf8 do
    frameworkConfig@FrameworkConfig { environment, dbPoolIdleTime, dbPoolMaxConnections, databaseUrl, logger } <- buildFrameworkConfig configBuilder
    modelContext <- createModelContext dbPoolIdleTime dbPoolMaxConnections databaseUrl logger

    let ?modelContext = modelContext
    (taskMain `finally` cleanup logger)

runTestName :: (?modelContext :: ModelContext, ?context :: FrameworkConfig) => IO (Either String GoogleMaps.Result)
runTestName = do
    site <- fetchOne (query @Site |> filterWhere(#postcode, "XXXXXX"))
    let geoCodeResult = (eitherDecode $ cs (get #googleMapsJson site))
    pure geoCodeResult
    ``` 

This then lets you return data from the script.         
```    
  IHP> geoResult <- runCustomScript ihpDefaultConfig runTestName 
```    

There might already be a better way of getting Query results in ghci? But this is the most convenient way I could find rather than writing the full query + processing in the interpreter.

Comment (mpscholten) [reply Guide/scripts.markdown:87]: This looks like something that rather should be a high level IHP function. Otherwise the `runCustomScript` relies on a lot of IHP internals and will break when we change those. Basically we need a `withIHP` function 👍

File: Guide/scripts.markdown [hunk 2]
```diff
@@ -64,6 +64,56 @@ This is made possible because of the [she-bang line](https://en.wikipedia.org/wi
...
+testing new APIs that require certain configuration variables.
+
+You can define a custom configuration (or import one from Config.hs) in your Script
```
mpscholten, line 90: I think this is too generic. I rather want this documentation to show specifically on how to import Config.hs and then use that configuration for running the script as that's the most common case (and good docs need to be specific).

Guide/scripts.markdown
batrakeshav10: we can have an example call to runScript with the custom config imported

Guide/scripts.markdown
Montmorency: Updated to show the import in a script and then running it from ghci.

Guide/scripts.markdown
mpscholten: 👍 looks like what I had in mind.

I'd actually make this function part of IHP to avoid B.C. issues in the future when IHP internals change.

Issue #1495, opened 2022-08-30
HSX Template compiling lambda functions


Montmorency:
On upgrade to v0.20 some old hsx code with a lambda expression stopped compiling:
Old code:
```
     <input type="hidden" name="googlePlaceId" value={(\(PlaceId x) -> x)(place_id x)}/>
```
Error message:
```


 • Exception when trying to run compile-time code:
        IHP.HSX.HsExpToTH.toPat: not implemented: (PlaceId x)
CallStack (from HasCallStack):
  error, called at ./IHP/HSX/HsExpToTH.hs:195:18 in ihp-hsx-0.18.0-5G7n5wCYmB1CeYaNqV7khQ:IHP.HSX.HsExpToTH
      Code: Language.Haskell.TH.Quote.quoteExp
              hsx
              "\n\
              \                                 <div id=\"google-result\" data-lat-lon={encode $ (location (geometry x))}>  {formatted_address x} </div>\n\
              \                                 <input type=\"hidden\" name=\"eircode\" value={eircode} />\n\
              \                                 <input type=\"hidden\" name=\"geoCoords\" value={latlngParam x}/>\n\
              \                                 <input type=\"hidden\" name=\"googlePlaceId\" value={(\\(PlaceId x) -> x)(place_id x)}/>\n\
              \                                 <input type=\"hidden\" name=\"googleMapsJSON\" value={encode x}/>\n\
              \                                 {siteNameInputField sitename}\n\
              \                                 {submitButton {label=\"Confirm Location\"}}\n\
```

Fix: Move lambda expression out of HSX block and into where clause. 

Also some selectField code no longer compiling:
```
{(selectField #goal (Nothing:userGoals)){disableLabel=True}}
```

Montmorency:
Working on this in [branch](https://github.com/Montmorency/ihp/tree/hsx-conpat)
A lot of type familes make it tricky to find the right mapping (ConPat -> Th.ConP).


```
    html WelcomeView = [hsx|
                            <p> <a href={NewSessionAction}> Login </a> </p>
                            <p> <a href={DeleteSessionAction}> Logout </a> </p>
                            { renderGreetAdmin currentAdminOrNothing }

                            <form>
                            <!-- These work  -->
                            {infixList}
                            <input type="hidden" name="googlePlaceId" value={unpackedPlaceData}/>

                            <!-- These don't -->
<!--                             {1 : [2,3,42]} -->
[...]

PR #1535, opened 2022-09-26
Hsx conpat

Montmorency:
Mapped a few more data types from HsSyn to TH. Pattern matching on Constructors and infix operators like (:), (,) should now work in [hsx||] blocks. This should address issue #1495. Note the Constructors in hsx blocks don't support Type Annotations (may require ghc 9.4.2?).


Montmorency:
This should be ready for review 👍

mpscholten (approved):
Looks good! Thanks :)

PR #1541, opened 2022-10-02
Split IDE module into own cabal package

mpscholten:
This is useful for smaller binaries in production. There's no need for shipping all of the IHP IDE in production builds.

[Requires cabal 3.8 to build.](https://github.com/haskell/cabal/issues/5423#issuecomment-1222225119)


mpscholten:
This has been done via `ihp-ide`

PR #1554, opened 2022-10-17
adding turbo tags to hsx parser.

Montmorency:
[Turbo](https://turbo.hotwired.dev/reference/frames) Looks like a pretty natural out of the box fit for IHP apps that want to get some SPA features. Support for hsx parser requires the <turbo-frame> and <turbo-stream> tags.


mpscholten:
These should already work as they're web component tags https://github.com/digitallyinduced/ihp/blob/master/Test/HSX/QQSpec.hs#L104

Issue #1507, opened 2022-09-02
Declarative Nix installation how-to


sir4ur0n:
Hello,

I am fiddling with IHP these days and I noticed the [installation documentation](https://ihp.digitallyinduced.com/Guide/installation.html) does not provide instructions on declarative Nix installation. The only Nix installation mode is "stateful" using `nix-env`.

I also searched in this repo's issues but found no issue about Nix (Flake) installation.

How do you feel about adding instructions to install IHP in a declarative way?

Ideally both "Legacy Nix" and "Flake" instructions would be provided.

For what it's worth/in case anyone is interested/searches for information in this repository, I seem to have managed to install IHP on my flakified NixOS (I also use Home manager but that should be irrelevant) with the following:

```nix
# flake.nix
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
    ihp = {
      url = "https://downloads.digitallyinduced.com/ihp-new.tar.gz";
      # IHP does not currently provide a `flake.nix` file, only a `default.nix`, so we will manually import it in the `outputs`
      flake = false;
    };
  };

  outputs = { self, nixpkgs, ihp }:
    let
      system = "x86_64-linux";
      ihp-overlay = final: prev: {
        # Make sure that `ihp-new` uses the same Nix packages as the rest
        ihp-new = import ihp { pkgs = nixpkgs.legacyPackages.${system}; };
      };
    in
    {
      nixosConfigurations.nixos-sir4ur0n = nixpkgs.lib.nixosSystem {
        inherit system;
        modules = [
          ({ ... }: { nixpkgs.overlays = [ ihp-overlay ]; })
          # Other modules...
        ];
        # Other attributes...
      };
    };
}
```
Then add `ihp-new` in your installed packages; be it in `configuration.nix`, `home.nix`, or wherever you configure your packages.

Thank you for your work on IHP, it looks pretty neat!

mpscholten:
Thanks for the suggestion 👍 Making IHP easy to work with NixOS is definitely something we should do.

We could also get `ihp-new` into nixpkgs directly. That way it can be added to a NixOS system by e.g. adding it to `/etc/nixos/configuration.nix`.

Btw: Flakes also seem to be a good fit for replacing `nix-shell` inside IHP projects. Especially with the direnv integration and it's support for providing faster nix-shell's.

sir4ur0n:
[...]

PR #1559, opened 2022-10-20
Use the new ihp-new package from nixpkgs for install instructions in the docs

mpscholten:
Documentation for changes made in https://github.com/digitallyinduced/ihp/issues/1507

@sir4ur0n could you review this?

### DIFF
--- a/Guide/installation.markdown
+++ b/Guide/installation.markdown
@@ -117,6 +117,15 @@ Note that nix can gradually grow to use several GB of disk space, especially aft
 You can now install IHP by running:
 
 ```bash
+nix-env --install ihp-new
+```
+
+
+#### Package not Found?
+
+The package has only been added to nixpkgs recently. In case you're using an older nixpkgs version, the package might not be available yet. In that case you can install `ihp-new` like this:
+
+```bash
 nix-env -f https://downloads.digitallyinduced.com/ihp-new.tar.gz -i ihp-new
 ```
 


Guide/installation.markdown
```diff
@@ -116,6 +116,15 @@ Note that nix can gradually grow to use several GB of disk space, especially aft
 
 You can now install IHP by running:
 
+```bash
+nix-env --install ihp-new
+```
+
```
sir4ur0n, line 122: How about mentioning `ihp-new` can also be added to the NixOS or Home Manager list of packages?

E.g.:
```
If you use Home Manager or NixOS, you can also add `ihp-new` to the list of packages.
```

sir4ur0n: Related but different: as an alternative, users may try without installing by running either `nix-shell --packages ihp-new` or `nix shell nixpkgs#qmmp`.

How about also documenting it? Or is it overkill?

The reason I ask is because Nix/NixOS encourage declarative style/reproducibility, and installing with `nix-env -i` kinda goes against :sweat_smile:

mpscholten: > How about mentioning ihp-new can also be added to the NixOS or Home Manager list of packages?

👍  

> How about also documenting it? Or is it overkill?

I think these are all things only more experienced nix people know about. The documentation here is mostly not aimed at those people, the experts already know how to do it their own way and will skip our instructions. Also `nix shell` currently doesn't work without explicitly enabling it in the nix config (https://github.com/NixOS/nix/issues/7000).

sir4ur0n: Fair enough!

sir4ur0n (approved):

Issue #1560, opened 2022-10-24
SSC change the class attributes on input or textarea elements failed


leobm:
Hi! i have a strange problem in the hsx render function of a component.

It is about Dynamic Attributes, in my case a "class" attribute.
errorMessage is of type Maybe.
The first render works too, but when the errorMessage value changes not the class attribute is updated but e.g.
the value property of the DOM node. 
There is then in the input  field the "form-control is-invalid" text **(value attribute)**.
The class attribute itself is not updated. 

**[Platform/Environment]**
IHP v1.0.0

**[Steps to Reproduce]**
checkout this test IHP app: https://github.com/leobm/ihp-test-ssc-bug

**[Actual Result]**
see the text "form-control is-invalid" in the input field (value attribute)
![bug](https://user-images.githubusercontent.com/25712/197642985-bacbb1d7-cc48-485e-8332-db35a976d405.png)

**[Expected Result]**
"form-control is-invalid" should be added to the class attribute of the input field
![fixed](https://user-images.githubusercontent.com/25712/197643010-6e93e45c-bc9b-47ef-9b1c-10f6eb3a63b2.png)

**[Fix]**
see this branch here: https://github.com/leobm/ihp/tree/fix_eval_attr_op
#1561

mpscholten:
Fixed via #1561 👍 

Thanks for the great bug report btw.

PR #1561, opened 2022-10-24
Fix change the class attributes on input or textarea elements in SSC component actions

leobm:
See https://github.com/digitallyinduced/ihp/issues/1560


mpscholten (approved):
Very good catch! Thanks :)

PR #1568, opened 2022-11-30
Cache more Haskell packages

nomeata:
as advised in https://ihpframework.slack.com/archives/C01DX8592H0/p1669795611222709?thread_ts=1669749271.272259&cid=C01DX8592H0


mpscholten (approved):
Thanks 👍

PR #1582, opened 2023-01-16
doc Emacs eglot setup

unhammer:
Now that haskell-language-server is included with IHP, and LSP support is built-in to the newest release of Emacs, it makes sense to change the default recommendation to using those.

(I'm not very happy with the workaround to make emacsclient work as an IHP_EDITOR, though it's a fairly common trick since it expects args in a different syntax than `code --goto`.)


mpscholten (approved):
Thanks for updating this! :)

PR #1585, opened 2023-01-19
Shipnix instructions in "Your first project"

kodeFant:


amitaibu:
I think all the red markers should be done with a screenshot tool (I use Shutter on Ubuntu)

Currently on the PR it looks like this

image


But it would be better to use a rectangle or an arrow.

image

kodeFant:
> I think all the red markers should be done with a screenshot tool (I use Shutter on Ubuntu)
> 
> Currently on the PR it looks like this
> 
> image
> 
> But it would be better to use a rectangle or an arrow.
> 
> image

Sure, will check out Shutter!

PR #1592, opened 2023-01-25
minor typos

agreif:


mpscholten (approved):
Thanks :) Good catch with the `Violation` change 👍

PR #1604, opened 2023-02-13
Postgres Interval

Montmorency:
This PR allows for use of postgres interval in an IHP applications Schema. It also provides some helper parsers so that the default output style of postgres intervals can be parsed (faithfully) into a data structure that allows for accurate Date/Time arithmetic to be done using the standard Haskell Date/Time library and interval data.


IHP/Postgres/TimeParser.hs
```diff
@@ -0,0 +1,134 @@
...
+
+-- To support the default postgres output style PGInterval -> PGTimeInterval
+-- in Application Code we provide the parser combinators for the `postgres` output style.
+-- for parsing (optional combination of): Y year[s] M mon[s] D day[s] [-]HH:MM:SS.[SSSs].
+-- This corresponds to the default interval `postgres` format.
+-- (https://www.postgresql.org/docs/current/datatype-datetime.html).
+-- alternative parsers would need to be provided for the `sql_standard`, `postgres_verbose`, and `iso_8601`
+-- styles/
+
+unpackInterval (PGInterval bs) = case parseOnly pPGInterval bs of
```
mpscholten, line 50: Can you add a type sig here?

IHP/Postgres/TimeParser.hs
```diff
@@ -0,0 +1,134 @@
...
+-- `postgres` output style into a a PGTimeInterval data struct of years, months, days, and NominalDiffTime.
+--  These can be combined with the standard Calendar/Time library to perform Calendar Arithmetic
+--  Days and Years are big Integers and can be added to the Gregorian year and Julian Day respectively
+--  Months are small Int (up to 1-11) denoting a month of the year, and the pgClock is a Nominal DiffTime
+--  representing the time as measured by a clock without leap seconds.
+
+data PGTimeInterval = PGTimeInterval { pgYears :: Integer
+                                     , pgMonths :: Int
+                                     , pgDays :: Integer
+                                     , pgClock :: NominalDiffTime} deriving (Eq, Show)
```
mpscholten, line 40:
```suggestion
                                     , pgClock :: NominalDiffTime } deriving (Eq, Show)
```

IHP/Postgres/TimeParser.hs
```diff
@@ -0,0 +1,134 @@
...
+-- and the semantics of the interval quantity in the application code (if they differ from the postgres
+-- interpretation for some reason).
+-- This module provides the PGInterval wrapper type and a parser for the default
+-- `postgres` output style into a a PGTimeInterval data struct of years, months, days, and NominalDiffTime.
+--  These can be combined with the standard Calendar/Time library to perform Calendar Arithmetic
+--  Days and Years are big Integers and can be added to the Gregorian year and Julian Day respectively
+--  Months are small Int (up to 1-11) denoting a month of the year, and the pgClock is a Nominal DiffTime
+--  representing the time as measured by a clock without leap seconds.
+
+data PGTimeInterval = PGTimeInterval { pgYears :: Integer
```
mpscholten, line 37: Could we make these fields strict (add a `!`)?

Test/Postgres/Interval.hs
```diff
@@ -0,0 +1,23 @@
...
+import IHP.Postgres.Interval
+import IHP.Postgres.TimeParser
+import Database.PostgreSQL.Simple.ToField
+import qualified Data.Attoparsec.ByteString.Char8 as Attoparsec
+
+tests = do
+    describe "Interval" do
+        describe "Parser" do
+            it "Should Parse" do
+                unpackInterval (PGInterval "25 years 6 mons 4 days 114:00:00.123") `shouldBe`  (PGTimeInterval 25 6 4 410400.123)
```
mpscholten, line 19: Can we add a test case for uppercase strings like `"25 YEARS 6 MONS 4 DAYS 114:00:00.123"`?

Montmorency: The intention of PGInterval should only be wrapping bytestring output from postgres database rather than more arbitrary interval strings from user land. Probably want to keep this as fast possible given the strings will have a definite formatted style?

mpscholten: ok makes sense 👍

IHP/IDE/SchemaDesigner/Parser.hs
```diff
@@ -246,6 +246,7 @@ sqlType :: Parser PostgresType
 sqlType = choice $ map optionalArray
         [ uuid
         , text
+        , interval --Needs higher precedence eotherwise parsed as Int
```
mpscholten, line 249:
```suggestion
        , interval -- Needs higher precedence otherwise parsed as Int
```

IHP/IDE/SchemaDesigner/Types.hs
```diff
@@ -211,6 +211,7 @@ data PostgresType
     | PDate
     | PBinary
     | PTime
+    | PInterval {fields :: Maybe Text}
```
mpscholten, line 214:
```suggestion
    | PInterval { fields :: Maybe Text }
```

IHP/Postgres/Interval.hs
```diff
@@ -0,0 +1,43 @@
...
+
+instance ToField PGInterval where
+    toField (PGInterval interval) = toField (interval)
+
+instance FromJSON PGInterval where
+     parseJSON (Object v) = (PGInterval . encodeUtf8) <$> ((v .: "interval"))
+     parseJSON _ = mzero
+
+instance ToJSON PGInterval where
+    toJSON (PGInterval pgInterval) = object [ "interval" .= (tshow pgInterval)]
```
mpscholten, line 43:
```suggestion
    toJSON (PGInterval pgInterval) = object [ "interval" .= (tshow pgInterval) ]
```

Montmorency: After further testing I think the toJSON method should actually just return a string of the Postgres interval rather than an object. 

if you have a data structure:
data Activity = Activity {duration :: PGInterval}

in Client you would rather access this as:

```
const activity = useQuerySingleResult(query("activities"))
//interval string
<p> Interval: {activity.duration} </p>

// unnecessary object {activity.duration.interval}
```

mpscholten: agree 👍

mpscholten (approved):

PR #1606, opened 2023-02-15
Add info for VSCode Devcontainers

rvarun11:
The devcontainer config used for [Codespaces](https://github.com/rvarun11/codespaces-ihp) can be used for local devcontainers with a minor modification.


mpscholten (approved):
👍

PR #1612, opened 2023-02-25
Added onlyWhere, onlyWhereReferences and onlyWhereReferencesMaybe

mpscholten:
Related question in the IHP Slack:

> There's a function I find myself writing in specific over and over again, and I'd love to be able to write a general form. E.g.
getUserPosts user posts =
  filter (\p -> p.userId == user.id) posts
Given an A, and a list of Bs with an (Id A) field, find all Bs that refer to A.
Does anyone know how to write that? I don't know how to write the
p.userId
expression as a generic
p.(id A)

### DIFF
--- a/IHP/ModelSupport.hs
+++ b/IHP/ModelSupport.hs
@@ -957,18 +957,44 @@ withTableReadTracker trackedSection = do
 -- If the Id field is nullable, we need to use 'onlyWhereReferencesMaybe':
 --
 -- > getUserTasks user tasks =
--- >     tasks |> onlyWhereReferences #optionalUserId user
+-- >     tasks |> onlyWhereReferencesMaybe #optionalUserId user
 --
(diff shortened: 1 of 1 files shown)


IHP/ModelSupport.hs
```diff
@@ -935,3 +935,41 @@ withTableReadTracker trackedSection = do
...
+--
+-- > getUserTasks user tasks =
+-- >     tasks |> onlyWhereReferences #optionalUserId user
+--
+onlyWhere :: forall record fieldName value. (KnownSymbol fieldName, HasField fieldName record value, Eq value) => Proxy fieldName -> value -> [record] -> [record]
+onlyWhere field value records = filter (\record -> get field record == value) records
+
+-- | Shorthand filter function for Id fields
+--
+-- See 'onlyWhere' for more details.
```
amitaibu, line 967: I think worth adding docs also to this one, as for me it's not clear when/ how to use

mpscholten: Good point 👍  just pushed an update

amitaibu: Thanks, looks good

Issue #1621, opened 2023-03-15
PGListener causes 100% cpu usage when unable to connect to DB


mpscholten:
[This loop](https://github.com/digitallyinduced/ihp/blob/master/IHP/PGListener.hs#L235) can easily cause 100% cpu usage when it's unable to connect to postgres. We should add some kind off backoff strategy there to sleep for some time before retrying (after the first retry failed).

PR #1623, opened 2023-03-24
Add backoff strategy when a PGListener fails

kodeFant:
/claim #1621

Hi! 

I figured it would be fun to give this bounty hunt a try :)

I believe this should solve #1621 although I'm unsure of the best way to test a failed connection. Tested AutoRefresh with it and it seemed to work fine!

Added a sleep before re-running the function and replacing forever with a recursion.

Double the sleep delay for the next retry until it reaches a minute, then it retries every minute from then on.

I don't know what should be the best amounts here so feel free to suggest other values than min 500ms and max 1 minute.

### DIFF
--- a/IHP/PGListener.hs
+++ b/IHP/PGListener.hs
@@ -245,7 +245,7 @@ notifyLoop listeningToVar listenToVar subscriptions = do
                         Just (error :: AsyncCancelled) -> throw error
                         notification -> do
                             let ?context = ?modelContext -- Log onto the modelContext logger
-                            Log.info ("PGListener is going to restart, loop failed with exception: " <> displayException error <> ". Retrying in " <> formatTimeUnits nextDelay <> ".")
+                           
 
 
                             -- Sleep for the current delay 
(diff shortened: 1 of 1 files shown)


IHP/PGListener.hs
```diff
@@ -231,17 +232,30 @@ notifyLoop listeningToVar listenToVar subscriptions = do
...
-                    notification -> do
-                        let ?context = ?modelContext -- Log onto the modelContext logger
-                        Log.info ("PGListener is going to restart, loop failed with exception: " <> displayException error)
-            Right _ -> pure ()
+    -- Initial delay (in microseconds)
+    let initialDelay = 500 * 1000
+    -- Max delay (in microseconds)
+    let maxDelay = 60 * 1000 * 1000
+
+    let retryLoop delay = do
```
kodeFant, line 240: Could be that retryLoop should be separated into a separate function

IHP/PGListener.hs
```diff
@@ -231,17 +232,30 @@ notifyLoop listeningToVar listenToVar subscriptions = do
...
-            Right _ -> pure ()
+    let retryLoop delay = do
+            result <- Exception.try innerLoop
+            case result of
+                Left (error :: SomeException) -> do
+                    case fromException error of
+                        Just (error :: AsyncCancelled) -> throw error
+                        notification -> do
+                            let ?context = ?modelContext -- Log onto the modelContext logger
+                            Log.info ("PGListener is going to restart, loop failed with exception: " <> displayException error)
```
amitaibu, line 248: Maybe worth indicating in the message the delay (i.e. the `nextDelay` value)?

kodeFant: Yes sure! Does this look okay?

mpscholten:
First bounty submission, great! Code looks good 👍

One thing I've had in mind: It could happen that e.g. the database connection is lost for unrelated reasons (e.g. connection closed by postgres because of some timeout). In that case we can reconnect without any delay. It's important for the IHP server to keep the delays to a minimum to not loose any database changes accidentally (e.g. changes during a reconnect are not visible in AutoRefresh and DataSync).

So could we change the retry logic to retry without any delay on the first error, and then only add delay on the second retry?

kodeFant:
Thanks!

Yes. Now added an `isFirstError`  argument and only increasing delay from second retry and onward.

mpscholten (approved):

PR #1624, opened 2023-03-25
1601 default vals

tobz619:
I've figured out the pull request:

I've got 10 to go.

If you have any thoughts on some other things or style comments then let me know and I'll try and incorporate them.


Guide/search.js
mpscholten, line 1: Why was this file deleted?

Test/DefaultValues/CreateTableDefaults.hs
```diff
@@ -0,0 +1,84 @@
+{-# OPTIONS_GHC -Wno-ambiguous-fields #-}
+module Test.DefaultValues.CreateTableDefaults where
```
mpscholten, line 2: As this is used outside the testsuite scope (e.g. in `IHP/IDE/SchemaDesigner/SchemaOperations`) shouldn't we move this to something like `IHP.IDE.SchemaDesigner.Defaults`?

Test/DefaultValues/CreateTableDefaults.hs
```diff
@@ -0,0 +1,84 @@
...
+
+@setColumnType pgt = defColumn {columnType = pgt}@
+
+Or
+
+Just as part of where you're calling it:
+
+@someDefaultColumnType = defColumn {columnType = PDate}@
+-}
+defColumn :: Column
```
mpscholten, line 42: As `name` and `columnType` are always set, we could also make these two arguments to this function:

```
defColumn :: Text -> PostgresType -> Column

-- Usage:
defColumn "id" PUUID
```

Test/IDE/CodeGeneration/Defaults/CodeGeneratorDefaults.hs
mpscholten, line 1: What's the idea behind extracting this out? I'm worried this might make changing the tests are in the future as we now have to change things in two places.

ihp.nix
mpscholten, line 1: This change should not be committed

mpscholten (commented):
Thanks 👍 Good job so far!

Guide/search.js
tobz619: I can't remember explicitly where but I remember something saying to delete this:

I'll resolve this by redownloading the file from the master branch :)

Test/DefaultValues/CreateTableDefaults.hs
tobz619: That's understandable, I was unsure where best to place them as I was focused more on getting started on making the defaults. I'll make a note to move and rename as needed.

tobz619: That's understandable, I was unsure where best to place them as I was focused more on getting started on making the defaults. I'll make a note to move and rename as needed.

Test/DefaultValues/CreateTableDefaults.hs
tobz619: So I was debating with myself for this one as to what the default values were and how best to set them. I tend to have defaults with known constants and then change the values within as necessary -> make new defaults or constants from those, hence the current design. 

But I see the value in your approach and it makes a lot of sense!

ihp.nix
tobz619: Will add to `.gitignore`. Thanks :)

Test/IDE/CodeGeneration/Defaults/CodeGeneratorDefaults.hs
tobz619: You are correct! Initially I was unsure whether or not the tables would be used elsewhere. Realistically, they can all be kept in the same file and then only exported/imported as needed. I will move them all to the same file.

Guide/search.js
tobz619: I've found and remember why: I tried to `git rebase origin/master` and I had to remove it but I messed up the rebase. Will do it again now

tobz619:
I think one of the tests is incorrect:
`it "should parse a CREATE TABLE with REAL, FLOAT4, DOUBLE, FLOAT8 columns" do
            parseSql "CREATE TABLE bools (a REAL, b FLOAT4, c DOUBLE PRECISION, d FLOAT8);" 
                `shouldBe` StatementCreateTable realFloatTable`
                
  I think its name should be something other than bools. I will change the name to `realfloat` so it passes the test

Test/IDE/SchemaDesigner/ParserSpec.hs
tobz619, line 1: This is the change in question

Test/IDE/SchemaDesigner/ParserSpec.hs
```diff
@@ -417,7 +417,7 @@ tests = do
                 `shouldBe` StatementCreateTable boolTable
 
         it "should parse a CREATE TABLE with REAL, FLOAT4, DOUBLE, FLOAT8 columns" do
-            parseSql "CREATE TABLE bools (a REAL, b FLOAT4, c DOUBLE PRECISION, d FLOAT8);" 
+            parseSql "CREATE TABLE realfloat (a REAL, b FLOAT4, c DOUBLE PRECISION, d FLOAT8);"
```
tobz619, line 420: This is the change in question.

tobz619:
I have work tomorrow and plans in the evening so I will have to pick it up again on Tuesday to complete everything else. Or at least I hope what is currently here is enough for someone to take and complete! this is fun :)

mpscholten:
Just checked it. Newly added modules need to be added to the `ihp.cabal` file in the project, otherwise the linker fails when building IHP. Basically the linker cannot find the function implementations for `defCreateTable` etc. because the module has has not been part of the build process. To fix this, add the missing module to `ihp.cabal` and then everything should work 👍

Guide/search.js
mpscholten, line 1: This is still showing up as changed for some reason. Could you follow the steps here https://stackoverflow.com/questions/39459467/remove-a-modified-file-from-pull-request to fix this? :)

IHP/IDE/Defaults/TableColumnDefaults.hs
```diff
@@ -0,0 +1,425 @@
...
+import qualified IHP.IDE.CodeGen.ControllerGenerator as ControllerGenerator
+import IHP.ViewPrelude (cs, plain)
+import qualified Text.Megaparsec as Megaparsec
+import IHP.IDE.CodeGen.Types
+import IHP.IDE.SchemaDesigner.Types
+import IHP.NameSupport
+import IHP.IDE.SchemaDesigner.Types
+
+
+{- | Takes a `Text` value for the name and creates a default column where all values are empty lists,
```
mpscholten, line 18:
```suggestion
{- | Takes a 'Text' value for the name and creates a default column where all values are empty lists, 
```

IHP/IDE/Defaults/TableColumnDefaults.hs
```diff
@@ -0,0 +1,425 @@
...
+defColumn :: Column
+defColumn = Column { 
+                 name = ""
+               , columnType = PUUID 
+               , defaultValue = Nothing
+               , notNull = False 
+               , isUnique = False
+               , generator = Nothing
+               }
+
```
mpscholten, line 59: Typically IHP code is indentend a bit different, like this:

```suggestion
defColumn =
    Column
    { name = ""
    , columnType = PUUID 
    -- ....
    }
```

IHP/IDE/Defaults/TableColumnDefaults.hs
```diff
@@ -0,0 +1,425 @@
...
+                        , columns = []
+                        , primaryKeyConstraint = PrimaryKeyConstraint []
+                        , constraints = []
+                        , unlogged = False
+                        }
+@
+
+-}
+defCreateTable :: Text -> CreateTable
+defCreateTable t = CreateTable {
```
mpscholten, line 78: The variables here should be spelled out, e.g. `tableName` instead of `t`

```suggestion
defCreateTable tableName = CreateTable {
```

IHP/IDE/Defaults/TableColumnDefaults.hs
```diff
@@ -0,0 +1,425 @@
...
+                        name = t
+                        , columns = []
+                        , primaryKeyConstraint = PrimaryKeyConstraint []
+                        , constraints = []
+                        , unlogged = False
+                        }
+
+-- | Takes a name for our table and a list of column and inserts the list
+--  into to our default table.
+defCreateTableWCol :: Text -> [Column] -> CreateTable
```
mpscholten, line 88: IMO this can be just done via `defCreateTable`. When you have a table that has no columns, you could still pass an empty list like `defCreateTable "myTableWithoutCols" []`.

IHP/IDE/Defaults/TableColumnDefaults.hs
```diff
@@ -0,0 +1,425 @@
...
+
+-- | A version of `setColumn` where @notNull = True@
+setColumnN :: Text -> PostgresType -> Column
+setColumnN n p = (setColumn n p) {notNull = True}
+
+-- | Sets a column to have a default value. Would recommend using in conjunction with `setColumn`
+setColumnDefaultVal :: Maybe Expression -> Column -> Column
+setColumnDefaultVal expression column = column {defaultValue = expression}
+
+
```
mpscholten, line 170: Functions below this lines are IMO not a good idea. By having these very specific functions outside of the test suites we actually make tests harder to change, because you now need to look at two files to understands what's going on. So it would be better to undo the changes below and just make use of `defTable` and `defColumn` inside the tests.

README.md
mpscholten, line 1: This undos some changes from master again, could you revert this?

Test/IDE/SchemaDesigner/CompilerSpec.hs
```diff
@@ -491,130 +408,38 @@ tests = do
...
-                    , primaryKeyConstraint = PrimaryKeyConstraint ["order_id", "truck_id"]
-                    , constraints = []
-                    , unlogged = False
-                    }
+            let statement = StatementCreateTable orderTrucksTable
             compileSql [statement] `shouldBe` sql
 
         it "should compile a CREATE TABLE statement with an array column" do
             let sql = cs [plain|CREATE TABLE array_tests (\n    pay_by_quarter INT[]\n);\n|]
-            let statement = StatementCreateTable CreateTable
```
mpscholten, line 589: Here's an example how I imagine this to be working after the PR is merged:

```haskell
            let statement = StatementCreateTable (
                defCreateTable "array_tests"
                    [ defColumn "pay_by_quarter" (PArray PInt)
                    ]
            )
```

ihp.nix
mpscholten, line 1: This also needs to be reverted

mpscholten (commented):
We're getting closer :)

IHP/IDE/Defaults/TableColumnDefaults.hs
tobz619: Yes, I agree now, especially now I see how often each pattern appears, it makes more sense to define this in these ways - will alter :)

IHP/IDE/Defaults/TableColumnDefaults.hs
tobz619: So my thinking here was that if you had repeating units or tests, you can define them once, remember them and then use them as normal and so long as you know the definition of the table or column, you could:
1. reuse it
2. use it as a new base for a table/column

Perhaps it's not useful for whole tables but certainly default values/columns that may repeat and have complicated structures.

That said, I do see your point and there are several items that are superfluous; secondly, I'm sure if you needed to define repeatable units, you would define them as needed and perhaps on the test page like was done before.

Therefore, I will revert the tests to defintions of `defColumn` etc and leave space for IHP to define its own monomers as required.

tobz619: I was also thinking that the writer would check the definition of the test by ctrl+clicking on the name and opening up the references window -> but I guess not everyone uses vscode for IDE which is understandable.

tobz619:
Okay should all now be complete :)

IHP/IDE/Defaults/TableColumnDefaults.hs
```diff
@@ -0,0 +1,232 @@
+{-|
+Module: Test.IDE.Defaults.TableColumnDefaults
```
mpscholten, line 2:
```suggestion
Module: IHP.IDE.Defaults.TableColumnDefaults
```

IHP/IDE/Defaults/TableColumnDefaults.hs
```diff
@@ -0,0 +1,232 @@
...
+               , notNull = False 
+               , isUnique = False
+               , generator = Nothing
+               }
+@
+
+If you want a different 'PostgresType' you will need to 
+specify with like so by either using a function:
+
+@setColumnType pgt = defColumn {columnType = pgt}@
```
mpscholten, line 42: Some of the comments here are outdated, e.g. `setColumnType` doesn't exist anymore. Also there's sometimes still some markdown syntax \`something\` instead of 'something' .

IHP/IDE/Defaults/TableColumnDefaults.hs
```diff
@@ -0,0 +1,232 @@
...
+                               , colText  "password_hash"
+                               , colText "email"
+                               , setColumnN "company_id" PUUID
+                               , setColumn "picture_url" PText
+                               , setColumnDefaultVal (Just (CallExpression "NOW" [])) $ setColumnN "created_at" PTimestampWithTimezone
+                               ]
+@
+-}
+compilerSpecTable :: CreateTable
+compilerSpecTable = defCreateTablePKID "users" ["id"] cols
```
mpscholten, line 219: As this is test related logic it shouldn't live here

mpscholten: maybe we can move this into some test module instead

IHP/IDE/Defaults/TableColumnDefaults.hs
```diff
@@ -0,0 +1,232 @@
...
+@
+
+-}
+defCreateTableWSetCol :: Text -> Text -> PostgresType -> CreateTable
+defCreateTableWSetCol tablename columnname pgt = defCreateTable tablename (pure $ setColumn columnname pgt)
+
+{- | Same as its progenitor `defCreateTableWSetCol` except it uses `setColumnN`
+-}
+defCreateTableWSetColN :: Text -> Text -> PostgresType -> CreateTable
+defCreateTableWSetColN tablename columnname pgt = defCreateTable tablename (pure $ setColumnN columnname pgt)
```
mpscholten, line 114: Same with those functions, I think it's more clear to write:

```haskell
emptyTable { name = "users", columns = [ colUUID ] }
```

instead of
```haskell
defCreateTableWSetCol "users" "id" ...
```

IHP/IDE/Defaults/TableColumnDefaults.hs
```diff
@@ -0,0 +1,232 @@
...
+{- | A recurring unit as found in many tests and files.
+
+Defined as such:
+
+>>> colUUID
+Column {name = "id", columnType = PUUID, defaultValue = Just (CallExpression "uuid_generate_v4" []), notNull = True, isUnique = False, generator = Nothing}
+
+-}
+colUUID :: Column
+colUUID = setColumnDefaultVal (Just (CallExpression "uuid_generate_v4" [])) $ setColumnN "id" PUUID
```
mpscholten, line 185: Small thing but could we call it `idColumn`? :) I think it's more specific

IHP/IDE/Defaults/TableColumnDefaults.hs
```diff
@@ -0,0 +1,232 @@
...
+{- | Give a column the text defined by 'text' and sets its 'columnType' to 'PText'. Uses `setColumnN`.
+
+__Example:__
+
+>>> colText "example"
+Column {name = "example", columnType = PText, defaultValue = Nothing, notNull = True, isUnique = False, generator = Nothing}
+
+-}
+colText :: Text -> Column
+colText text = setColumnN text PText
```
mpscholten, line 196: I'd prefer to also just write `emptyColumn { name, columnType = PText }`

IHP/PGListener.hs
```diff
@@ -232,38 +231,18 @@ notifyLoop listeningToVar listenToVar subscriptions = do
                             let inChan = get #inChan subscription
                             Queue.writeChan inChan notification
 
-    -- Initial delay (in microseconds)
```
mpscholten, line 235: This change seems like unrelated to the PR. Could you revert this?

Test/IDE/CodeGeneration/ControllerGenerator.hs
```diff
@@ -12,61 +12,14 @@ import qualified Text.Megaparsec as Megaparsec
 import IHP.IDE.CodeGen.Types
 import IHP.IDE.SchemaDesigner.Types
 import IHP.NameSupport
+import IHP.IDE.Defaults.TableColumnDefaults
 
 tests = do
     describe "Controller Generator Tests:" do
         let schema = [
-                    StatementCreateTable CreateTable {
```
mpscholten, line 19: 👍

Test/IDE/SchemaDesigner/ParserSpec.hs
```diff
@@ -1134,10 +999,13 @@ COMMENT ON EXTENSION "uuid-ossp" IS 'generate universally unique identifiers (UU
...
+                                                                                  in defCreateTable "public_variables" idCol}
 
         it "should parse an 'CREATE UNLOGGED TABLE' statement" do
-            parseSql "CREATE UNLOGGED TABLE pg_large_notifications ();"  `shouldBe` StatementCreateTable CreateTable { name = "pg_large_notifications", columns = [], primaryKeyConstraint = PrimaryKeyConstraint [], constraints = [], unlogged = True }
+            parseSql "CREATE UNLOGGED TABLE pg_large_notifications ();"  
+                `shouldBe` 
+                    StatementCreateTable (defCreateTable "pg_large_notifications" []) {unlogged = True}
 
 
 col :: Column
```
mpscholten, line 1011: This `col` function here btw can likely be replaced with `defColumn` 👍

Test/IDE/SchemaDesigner/SchemaOperationsSpec.hs
```diff
@@ -450,39 +336,12 @@ tests = do
...
-                                        , generator = Nothing
-                                        }
-                            ]
-                            , primaryKeyConstraint = PrimaryKeyConstraint []
-                            , constraints = []
-                            , unlogged = False
-                            }
+                let tableAWithUpdatedColumn = StatementCreateTable 
+                                                $ defCreateTable "a" 
+                                                $ pure $ setColumn "created_at2" PText
```
mpscholten, line 343: It's more specific if we're using `[]` syntax here instead of `pure`

```suggestion
                let tableAWithUpdatedColumn = StatementCreateTable 
                                                $ defCreateTable "a" [ setColumn "created_at2" PText ]
```

IHP/IDE/Defaults/TableColumnDefaults.hs
```diff
@@ -0,0 +1,232 @@
...
+                  , columns = []
+                  , primaryKeyConstraint = PrimaryKeyConstraint []
+                  , constraints = []
+                  , unlogged = False
+                  }
+
+-- | Takes a name for our table and a list of column and inserts the list
+--   into to our empty table.
+defCreateTable :: Text -> [Column] -> CreateTable
+defCreateTable tablename columns = emptyTable { name = tablename
```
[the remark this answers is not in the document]
tobz619: It's more for speed: it means you don't have to write the record parts over and over again especially if they're the only two values in `emptyTable` being changed and you could quickly be certain that the only values changed were the name and the columns contained. If other values needed changing, then it's up to the writer to use `emptyTable {..}` or `(defCreateTable tablename columns){..})` -> both options are always available.

Since they were a recurring feature, it made sense to write them this way initially.

IHP/IDE/Defaults/TableColumnDefaults.hs
tobz619: Same as before, for speed (not needing to write records) and composability (ability to easily use (.) and ($) to change values or combine them with other things). I do understand the concerns for clarity's sake though.

Since there were quite a few examples of where only certain values in certain fields are changed, if you know what the functions do, you can have a quick idea of what the table or column looks like or have a fast way of defining a more complex table as a base for another table by using base functions. 

Again, the option the use the base tables and columns and record syntax is always there and would be functionally identical.

By using a specific function that encapsulates a pattern, I can quickly reason out the structure of the full table if I know what the function does UNLESS I modify it `( )` and adding more records with `{..}` or putting another function ahead of it which modifies the table further.

It really depends on much you want to represent specific patterns of test with functions. Also a lot of these functions can end up being helper functions much larger structures; but these will be made when needed I imagine.

tobz619:
I've been a bit busy so haven't been able to rewrite the tests but I have time to do so tomorrow evening :)

mpscholten:
Great, thanks for the update! Happy to review again later today 👍

mpscholten:
Is this ready for review again? :)

tobz619:
> Is this ready for review again? :)

Not yet, hopefully by the end of tonight! :)

I must admit my motivation to rewrite to instances of  `emptyTable` and `emptyColumn` is low but I'll get it done (y)

tobz619:
I think, I'm done with this here:

Rewriting instances of `defaultCreateTablePKID` will take a long time and I feel it does its job and I need to move onto other things.

PR #1629, opened 2023-04-05
Update README.md url to use updated github username

saxxie-dev:
Right now it will redirect there anyway - but probably a nice courtesy to not use her deadname


mpscholten (approved):
Thanks! Good catch 👍

PR #1647, opened 2023-04-17
Fix htmx-hyperscript behaviour by default

kodeFant:
Makes htmx and hyperscript work out of the box with IHP and Turbolinks

Otherwise, after turbolinks transition, or an AutoRefresh, nodes with htmx and hyperscript on them won't work anymore

Could possibly also just be added to a htmx/hyperscript guide in the docs


kodeFant:
Makes sense!

PR #1648, opened 2023-04-17
Docs for htmx and hyperscript

kodeFant:


Guide/htmx-and-hyperscript.markdown
```diff
@@ -0,0 +1,178 @@
...
+
+## htmx
+
+htmx gives you access to AJAX, as a way to update views locally with an API conviently residing in HTML attributes.
+
+Instead of the typical Single Page Application pattern parsing JSON and turning into html in the DOM, the API endpoints simply just return HTML to be patched into the DOM.
+
+### Installation
+
+The recommended way of installing htmx is downloading [htmx.min.js](https://unpkg.com/htmx.org/dist/htmx.min.js) from unpkg.com and save it to `static/vendor/htmx.min.js` in your project directory.
```
amitaibu, line 38: For Tailwind we recommend [using npm](https://github.com/digitallyinduced/ihp/blob/efeb31cd35c6f62effcaa82dac4aa45e2a322e02/Guide/tailwindcss.markdown#installing-tailwind) - can we do something similar here?

kodeFant: Yes, it's very doable with the standalone cli which also has all official plugins like the form plugin already built in: https://tailwindcss.com/blog/standalone-cli

Only problem with that is cross-platform support as one binary for each system would be required.

Would be great if some NixOS maintainers would add it in their package registry (they only have the nodePackages one without integrated plugins) as I don't have the energy for it right now :) But shouldn't be too complex I think

mpscholten (approved):

PR #1650, opened 2023-04-17
Added ihp-openai

mpscholten:


ihp-openai/README.md
```diff
@@ -0,0 +1,117 @@
...
+    action DeleteQuestionAction { questionId } = do
+        question <- fetch questionId
+        deleteRecord question
+        setSuccessMessage "Question deleted"
+        redirectTo QuestionsAction
+
+fillAnswer :: (?modelContext :: ModelContext) => Question -> IO (Async ())
+fillAnswer question = do
+    -- Put your OpenAI secret key below:
+    let secretKey = "sk-XXXXXXXX"
```
amitaibu, line 89: Maybe move that to `Config.hs`?

mpscholten: Yes this should be the case for production deployment, but I wanted to keep the example simple 👍

PR #1651, opened 2023-04-18
Replace Tailwind npm instructions with Tailwind Standalone CLI

kodeFant:
This simplifies the installation of Tailwind by using a pre-built Tailwind binary instead of npm


kodeFant:
Closing this, as I mistakenly assumed that everyone would like to avoid npm :) 

Will create a new PR just offering it as an alternative instead of removing npm guide

PR #1667, opened 2023-05-15
Update events listeners for page specific scripts

amitaibu:


lib/IHP/static/helpers.js
```diff
@@ -1,10 +1,3 @@
-var ihpLoadEvent = new Event('ihp:load');
```
mpscholten, line 1: This change is a BC break, not sure whether that's worth it. `ihpLoadEvent` is also used somewhere in the file, so we'd need to remove the usages as well in case we want to delete these variables

amitaibu: > This change is a BC break, not sure whether that's worth it

In my tests I saw that `ihp:load` was not working as expected. So on the one hand it's BC break, but on the other - people using it might think it's working. Should we ask on Slack who's using `ihp:load`, see if we get people to answer?

lib/IHP/static/helpers.js
```diff
@@ -417,8 +406,6 @@ window.transitionToNewPage = function (newHtml) {
         return;
     }
 
-    document.dispatchEvent(ihpUnloadEvent);
```
amitaibu, line 420: tbh, I'm not sure about this change

amitaibu: I should probably revert, as it's a custom event

lib/IHP/static/helpers.js
mpscholten: I just checked and saw we're ourselves using this in several projects:

image

e.g. like this:

image

So let's keep this event here. But I'm fine with removing the outdated documentation on this and replacing it

amitaibu:
Here's a reproducible example where the custom on ready handler isn't invoked transitioning into the page.

https://github.com/amitaibu/ihp-landing-page/pull/7

PR #1669, opened 2023-05-17
Document command to ignore changes to `IHP/ihp.nix`

amitaibu:
Make it easier not to accidentally commit changes to `ihp.nix`


mpscholten (approved):

mpscholten:
This is neat :) Wasn't aware that a command for this exists

amitaibu:
I tried to git pull and got 

image

So the solution isn't complete. I'll try to figure a nicer solution..

mpscholten:
that's only because there's been a few changes to `ihp.nix` today on master. The file doesn't change that often typically

PR #1677, opened 2023-05-21
helpers.js: Support formAction attribute on <button>

unhammer:
So you can have e.g.

            <button class="btn btn-primary" type="submit" formaction={CreateAction}>
              Submit
            </button>
            <button class="btn btn-secondary" type="submit" formaction={SaveProgressAction} formnovalidate>
              Save progress
            </button>


unhammer:
Won't work on `js-delete` where the function is called without `possibleButton`, but it seems that's only supported on `<a>` anyway

mpscholten (approved):
TIL about `formaction`! Thanks 👍

PR #1678, opened 2023-05-22
Actually use formForWithOptions in its doc example

unhammer:
also in formForWithoutJavascript; and actually set formAction as promised by the paragraph above


mpscholten (approved):
Good catch, thanks :)

PR #1679, opened 2023-05-23
Add docs about Multiple Nested Records

amitaibu:


Guide/relationships.markdown
```diff
@@ -337,6 +337,113 @@ defining the `instance FromRow Maybe a`. This approach to joins allows
...
+and the authors of those comments.
+-}
+fetchPostWithRecords :: (?modelContext :: ModelContext) => Id Post -> IO PostWithRecords
+fetchPostWithRecords postId = do
+    post <- fetch postId
+
+    -- Comments referencing the post ID.
+    comments <- query @Comment
+        |> filterWhere (#postId, postId)
+        |> fetch
```
mpscholten, line 388: Can also be written as

```suggestion
    comments <- fetch post.comments
```

Guide/relationships.markdown
```diff
@@ -337,6 +337,113 @@ defining the `instance FromRow Maybe a`. This approach to joins allows
...
+showComment :: Comment -> [User] -> Html
+showComment comment commentUsers =
+    [hsx|
+        {comment.body} authored by {authorName}
+    |]
+    where
+        authorName = commentUsers
+            |> filter (\user -> user.id == comment.userId)
+            -- There should be a single user.
+            |> head
```
mpscholten, line 442:
```suggestion
            |> find (\user -> user.id == comment.userId)
```

Guide/relationships.markdown
amitaibu: To confirm - will it fetch multiple comments? (if so, cool :smile: )

mpscholten: yes exactly 👍  when `fetchRelated` is not called, the `comments` field is a query builder with the right `WHERE`  condition already applied (so declared like `data Post = Post { .., comments :: QueryBuilder "comments" }` )

amitaibu: I tried to do it on my repo, and I get an error

https://github.com/amitaibu/ihp-landing-page/pull/13/files#r1205985723

amitaibu: As `Comment` is referencing the `Post`, I don't expect to have `post.comments`

amitaibu: Sorry, it's working. In my case I had to switch `landingPage.paragraphCtas` with  `landingPage.paragraphCtasRefLandingPages`

amitaibu: I'll add a comment in the docs to clarify

Guide/relationships.markdown
```diff
@@ -337,6 +337,115 @@ defining the `instance FromRow Maybe a`. This approach to joins allows
...
+fetchPostWithRecords postId = do
+    post <- fetch postId
+
+    -- Fetch Comments referencing the post ID.
+    -- Eventhough we haven't used `fetchRelated` on the Post we still have the `post.comments` field.
+    -- This field field is a query builder with the right `WHERE` condition already applied (referencing the Post ID),
+    -- so we only need to `fetch` and don't have to use the more verbose `query`:
+    -- comments <- query @Comment
+    --     |> filterWhere (#postId, postId)
+    --     |> fetch
```
amitaibu, line 391: Added this comment

Guide/relationships.markdown
```diff
@@ -337,6 +337,115 @@ defining the `instance FromRow Maybe a`. This approach to joins allows
...
+showComment comment commentUsers =
+    [hsx|
+        {comment.body} authored by {authorName}
+    |]
+    where
+        authorName = commentUsers
+            |> find (\user -> user.id == comment.userId)
+            -- Get the user's name. As `head` returns a Maybe value we need to use `maybe mempty` which means
+            -- if no user found, don't show anything.
+            |> maybe mempty (get #name)
```
mpscholten, line 447: We can use an empty string here instead of mempty, it's simpler 👍 

```suggestion
            |> maybe "" (get #name)
```

Guide/relationships.markdown
```diff
@@ -337,6 +337,115 @@ defining the `instance FromRow Maybe a`. This approach to joins allows
...
+
+showComment :: Comment -> [User] -> Html
+showComment comment commentUsers =
+    [hsx|
+        {comment.body} authored by {authorName}
+    |]
+    where
+        authorName = commentUsers
+            |> find (\user -> user.id == comment.userId)
+            -- Get the user's name. As `head` returns a Maybe value we need to use `maybe mempty` which means
```
amitaibu, line 445:
```suggestion
            -- Get the user's name. As `head` returns a Maybe value we need to use `maybe ""` which means
```

PR #1680, opened 2023-05-23
Add `radioField`

amitaibu:
| Bootstrap | Tailwind |
| -- | -- |
|  image | image |

- [x] Add Radio field
- [x] Theme for Bootstrap (default)
- [x] Theme for Tailwind
- [x] Write tests
- [x] Update guide


IHP/View/CSSFramework.hs
```diff
@@ -856,6 +902,52 @@ tailwind = def
...
+                            autofocus={autofocus}
+                            {...additionalAttributes}
+                        />
+                        {label}
+                    </div>
+                |]
+                    where
+                        -- @todo: Need to find a way to guarantee that the id is unique and html valid.
+                        -- use `generateUniqueId`?
+                        optionId = fieldInputId <> "-" <> optionValue
```
amitaibu, line 947: @mpscholten what do you think?

IHP/View/Form.hs
```diff
@@ -759,6 +759,86 @@ selectField field items = FormField
...
+        , cssFramework = ?formContext.cssFramework
+        , helpText = ""
+        , placeholder = ""
+        , required = False
+        , autofocus = False
+    }
+    where
+        fieldName = symbolVal field
+        FormContext { model } = ?formContext
+{-# INLINE radioField #-}
```
amitaibu, line 840: What's the purpose of `INLINE`?

IHP/View/Form.hs
```diff
@@ -759,6 +759,86 @@ selectField field items = FormField
...
+-- > </form>
+--
+-- If you want a certain value to be preselected, set the value in the controller. For example, to have the first user be preselected in the above example:
+--
+-- > action NewProjectAction = do
+-- >     users <- query @User |> fetch
+-- >     let userId = headMay users |> maybe def (.id)
+-- >     let target = newRecord @Project |> set #userId userId
+-- >     render NewView { .. }
+radioField :: forall fieldName model item.
```
amitaibu, line 803: Should we rename it to `radiosField` to indicate there could be multiple?

IHP/View/CSSFramework.hs
```diff
@@ -199,6 +201,47 @@ instance Default CSSFramework where
...
+                                autofocus={autofocus}
+                                {...additionalAttributes}
+                            />
+                            {label}
+                        </div>
+                    |]
+                        where
+                            -- @todo: Need to find a way to guarantee that the id is unique and html valid.
+                            -- use `generateUniqueId`?
+                            optionId = fieldInputId <> "-" <> optionValue
```
mpscholten, line 242: typically we use `_` instead of `-` for form Ids in IHP

IHP/View/Form.hs
mpscholten: Seems technically correct, but I'd propably frist try `radioField` when typing the code, so maybe we should stick with it as it feels more in line with the other form helpers

IHP/View/Form.hs
mpscholten: When the code is inlined, this makes it easier for GHC to optimize the other very generic code. E.g. the code `cs (lcfirst (getModelName @model) <> "_" <> cs fieldName)` can be simplified a lot when the `model` type variable is statically known (which is typically the case when the helper is inlined at the right place)

mpscholten:
Thanks, this is great work so far 👍

IHP/View/Form.hs
amitaibu: How do you know when to use it? Is there some rule of thumb?

IHP/View/CSSFramework.hs
amitaibu: Fixed

amitaibu:
@mpscholten fixed the `-` to `_`. Ready for review..

IHP/View/Form.hs
mpscholten: There's no hard rule. I've discovered some of those places by dumping the GHC Core output and checking for potential places. GHC can also report some of those places using the `missed-specialisations` flag (it's enabled in IHP)

IHP/View/Form.hs
```diff
@@ -759,6 +759,86 @@ selectField field items = FormField
         FormContext { model } = ?formContext
 {-# INLINE selectField #-}
 
+-- | Radio require you to pass a list of possible values to select. We use the same mechanism as for for `selectField`.
```
mpscholten, line 762:
```suggestion
-- | Radio require you to pass a list of possible values to select. We use the same mechanism as for for 'selectField'.
```

mpscholten:
Thanks!

PR #1686, opened 2023-05-28
Use filename on `storeFileFromPath`

amitaibu:
If there is an existing filename, use it. Otherwise assign it a new UUID.


amitaibu:
We should probably do the same for https://github.com/digitallyinduced/ihp/blob/248cb16b361adb4e8838b33a496f7c2be09541de/IHP/FileStorage/ControllerFunctions.hs#L153 , I'll try...

amitaibu:
An effect of this is that files uploaded via form are now saved with their original names, which I think is useful.
However, a negative effect is that if one would uploaded `attack.sh` the file could be dangerous, so maybe we need to filter out dangerous extensions?

PR #1688, opened 2023-05-30
Allow passing `fileName` in the `StoreFileOptions`

amitaibu:


IHP/FileStorage/Types.hs
```diff
@@ -25,13 +25,15 @@ data StoreFileOptions = StoreFileOptions
     { directory :: Text -- ^ Directory on S3 or inside the @static@ where all files are placed
     , contentDisposition :: Wai.FileInfo LByteString -> IO (Maybe Text) -- ^ The browser uses the content disposition header to detect if the file should be shown inside the browser or should be downloaded as a file attachment. You can provide a function here that returns a custom content-disposition header based on the uploaded file. This currently only works with the S3 storage. See 'contentDispositionAttachmentAndFileName' for standard configuration.
     , preprocess :: Wai.FileInfo LByteString -> IO (Wai.FileInfo LByteString) -- ^ Can be used to preprocess the file before storing it inside the storage. See 'applyImageMagick' for preprocessing images.
+    , fileName :: Maybe UUID -- ^ Optional filename. We use UUID for security measures.
```
amitaibu, line 28:
```suggestion
    , fileName :: Maybe UUID -- ^ Optional filename. We use UUID for security reasons.
```

amitaibu:
Failing test doesn't seem related

amitaibu:
@mpscholten is ready, but not sure what to do with the failing build

mpscholten:
The error on the build seems to be unrelated.

Can we also add some documentation to the Guide on how to use the new `fileName` attribute? Otherwise it's unlikely people will be aware that this exists now

mpscholten (approved):
Thanks 👍

PR #1690, opened 2023-05-31
Guide about Image style implementation

amitaibu:


Guide/file-storage.markdown
```diff
@@ -575,3 +575,248 @@ config = do
...
+```haskell
+[hsx|<img src={pathTo $ RenderImageStyleAction 400 200 imageUrl signed} />|]
+where
+    imageUrl = "http://localhost:8000/static/picture/b4c8f55c-16d6-41f0-9503-77352b134e14"
+
+    -- Sign the image URL to prevent tampering.
+    Config.RsaPublicAndPrivateKeys (_, privateKey) = getAppConfig @Config.RsaPublicAndPrivateKeys
+    signed = case RSA.sign Nothing (Just Hash.Algorithms.SHA256) privateKey (cs $ imageUrl <> "400x200") of
+        Left msg -> error $ "Cannot sign image URL, private key is invalid:" <> show msg
+        Right signature -> signature |> Base64.encode |> cs
```
amitaibu, line 821:
```suggestion
        -- Base 64 decode the token so that it can be used in the URL.
        Right signature -> signature |> Base64.encode |> cs
```

Guide/file-storage.markdown
```diff
@@ -575,3 +575,246 @@ config = do
...
 
 [You can find a full list of `WaiParser.set...` functions on the `Network.Wai.Parse` documentation](https://hackage.haskell.org/package/wai-extra-3.1.6/docs/Network-Wai-Parse.html#v:setMaxRequestKeyLength)
+
+## Image Style Implementation
+
+Image style is a common feature in web applications. It allows you to define different styles for an image, like a thumbnail or a preview image. The image style is then applied to the image when it's viewed. One image can have multiple image styles. So on one page it would appear as a thumbnail, on another page it would appear with a 400px x 200px dimension. The original image remains as is, untouched, and whenever there is a demand for an image style then it is generated on the fly. If an image style was already generated then it is served, saving the effort of re-generating it.
+
+Let's see how we can implement that. At first, without any access control, only accept the original image along with the desired dimensions and generate the image style on the fly.
+
+We add an action to the `ImageStyleController` that accepts the original image path, the desired width and height.
```
amitaibu, line 585:
```suggestion
Let's see how we can implement that. At first, without any access control. We add an action to the `ImageStyleController` that accepts the original image path, the desired width and height.
```

Guide/file-storage.markdown
```diff
@@ -575,3 +575,244 @@ config = do
...
+
+```haskell
+-- Web/Controller/ImageStyle.hs
+
+module Web.Controller.ImageStyle where
+
+import Web.Controller.Prelude
+import IHP.ControllerSupport
+import System.Directory (doesFileExist)
+import qualified Data.Text as T
```
mpscholten, line 629: In IHP we typically import `Data.Text` as `Text`, as just `T` doesn't say anything useful

```suggestion
import qualified Data.Text as Text
```

Guide/file-storage.markdown
```diff
@@ -575,3 +575,244 @@ config = do
...
+        let (originalImageDirectory, uuid) = extractDirectoryAndUUID originalImagePath
+
+        let size = show width <> "x" <> show height
+        let imageStylePathDirectory = originalImageDirectory <> "/imageStyles/" <> size
+        let imageStylePath = imageStylePathDirectory <> "/" <> uuid
+
+        -- If we use a StaticDirStorage storage then we need to prefix the path with the `static/` folder.
+        let storagePrefix = case storage of
+                StaticDirStorage -> "static/"
+                _ -> ""
```
mpscholten, line 644: This can likely be moved to a top level definition à la
```haskell
storagePrefix :: (?context :: ControllerContext) => Text
storagePrefix = ...
```

This way the actual action implementation is a bit shorter

Guide/file-storage.markdown
```diff
@@ -575,3 +575,244 @@ config = do
...
+import Web.Controller.Prelude
+import IHP.ControllerSupport
+import System.Directory (doesFileExist)
+import qualified Data.Text as T
+
+
+instance Controller ImageStyleController where
+    action RenderImageStyleAction { width, height, originalImagePath } = do
+        -- Get the original image directory and UUID from the path.
+        let (originalImageDirectory, uuid) = extractDirectoryAndUUID originalImagePath
```
mpscholten, line 635: I'm worried that `originalImageDirectory` could set to something like `../`. The signature basically fixes this, but we should still maybe warn about this here.

Guide/file-storage.markdown
```diff
@@ -575,3 +575,244 @@ config = do
...
+    -- See https://ihp.digitallyinduced.com/Guide/config.html
+    -- for what you can do here
+
+    -- ...
+
+    -- Private and public keys to sign and verify image style URLs.
+    -- The 300 means how large your URL can be.
+    -- The 65537 is the public exponent of RSA.
+    -- You can probably leave those values as is.
+    (publicKey, privateKey) <- liftIO $ liftIO $ RSA.generate 300 65537
```
mpscholten, line 720: This generates a new RSA key every time the app restarts. This invalidates all existing image urls generated, which could affect the end user experience in bad ways. It would be better to generate the RSA key only once and store it in `Config/image_signing_key.rsa` and then load it from here. The key can be generated by using `openssl` from the CLI, or even by some Haskell code here inside the `Config.hs` that checks whether the file already exists.

Guide/file-storage.markdown
```diff
@@ -575,3 +575,244 @@ config = do
...
+    ];
+```
+
+Then we need to generate a private and public key pair. We can do it in our Config file, and save the keys for later use.
+
+```haskell
+-- Config/Config.hs
+import Crypto.PubKey.RSA as RSA
+
+newtype RsaPublicAndPrivateKeys = RsaPublicAndPrivateKeys (RSA.PublicKey, RSA.PrivateKey)
```
mpscholten, line 707: I think this could also be a data structure:

```suggestion
data RsaPublicAndPrivateKeys = RsaPublicAndPrivateKeys { publicKey :: RSA.PublicKey, privatekey :: RSA.PrivateKey }
```

Guide/file-storage.markdown
```diff
@@ -575,3 +575,244 @@ config = do
...
+import Crypto.Hash.Algorithms as Hash.Algorithms
+import Config
+import Data.ByteString.Base64 as Base64
+
+instance Controller ImageStyleController where
+    action RenderImageStyleAction { width, height, originalImagePath, signed } = do
+        let size = show width <> "x" <> show height
+
+        -- Verify the signed token.
+        let Config.RsaPublicAndPrivateKeys (publicKey, _) = getAppConfig @Config.RsaPublicAndPrivateKeys
```
mpscholten, line 745: This can also be moved to a top level declaration:

```suggestion
publicKey :: (?context :: ControllerContext) => PublicKey
publicKey = (getAppConfig @Config.RsaPublicAndPrivateKeys).publicKey
```

Guide/file-storage.markdown
```diff
@@ -575,3 +575,244 @@ config = do
...
+
+        -- Verify the token, and deny or allow access based on the result.
+        accessDeniedUnless case cs signed |> Base64.decode of
+            Left msg -> False
+            Right signed -> RSA.verify (Just Hash.Algorithms.SHA256) publicKey (cs $ originalImagePath <> size) signed
```
mpscholten, line 750: Maybe this can be extracted into it's own function  and then used à la `accessDeniedUnless (signatureMatches imageUrl signature)`

Guide/file-storage.markdown
```diff
@@ -575,3 +575,244 @@ config = do
...
+where
+    imageUrl = "http://localhost:8000/static/picture/b4c8f55c-16d6-41f0-9503-77352b134e14"
+
+    -- Sign the image URL to prevent tampering.
+    Config.RsaPublicAndPrivateKeys (_, privateKey) = getAppConfig @Config.RsaPublicAndPrivateKeys
+    signed = case RSA.sign Nothing (Just Hash.Algorithms.SHA256) privateKey (cs $ imageUrl <> "400x200") of
+        Left msg -> error $ "Cannot sign image URL, private key is invalid:" <> show msg
+        -- Base 64 decode the token so that it can be used in the URL.
+        -- and use `cs` to convert it to `Text`.
+        Right signature -> signature |> Base64.encode |> cs
```
mpscholten, line 817: Maybe we can make a helper for this and put it into `Application/Helper/View.hs`?

Guide/file-storage.markdown
amitaibu: Should we move it to `IHP.FileStorage.ControllerFunctions`?

Guide/file-storage.markdown
amitaibu: That's a valid concern. How about:

a) `error` if there is `../` present in the path OR b) `Text.replace "../" -- so it would end with 403?

Guide/file-storage.markdown
amitaibu: :+1: I'll try to store it under `Config/private_key.rsa` (as it could be used for other stuff not just image signing).

Also, I think it's nicer that we do it on Config, rather than asking user to `openssl` -- so one could only copy/ paste code - and it should work.

Guide/file-storage.markdown
amitaibu: Should I move this and the below suggestions to `Application.Helper.Controller`. Or maybe  `Application.Helper.RSA` ?

Guide/file-storage.markdown
amitaibu: Having `../` will result with access denied

Guide/file-storage.markdown
amitaibu: Moved to `IHP.FileStorage.ControllerFunctions`

Guide/file-storage.markdown
amitaibu: Kept in Application.Helper.Controller

Guide/file-storage.markdown
amitaibu: Done

Guide/file-storage.markdown
amitaibu: Done

Guide/file-storage.markdown
```diff
@@ -575,3 +575,273 @@ config = do
...
+
+```haskell
+-- Application/Helper/View.hs
+module Application.Helper.View where
+
+import IHP.ViewPrelude
+import Crypto.PubKey.RSA.PKCS15 as RSA
+import Crypto.Hash.Algorithms as Hash.Algorithms
+import Data.ByteString.Base64 as Base64
+import Application.Helper.Controller
```
amitaibu, line 754: The compiler is ok with it - just confirming it's conceptually ok.

Guide/file-storage.markdown
mpscholten: I've used the JWT package in the past and it also uses RSA keys for signing. And it also uses cryptonite, so maybe we can just copy how the JWT package is doing it: https://github.com/puffnfresh/haskell-jwt/blob/master/src/Web/JWT.hs

Especially the [`readRsaSecret`](https://github.com/puffnfresh/haskell-jwt/blob/master/src/Web/JWT.hs#L436) function is basically doing what we want.

Guide/file-storage.markdown
mpscholten: I'd keep it in Config as it's only used for configuration purposes

Guide/file-storage.markdown
```diff
@@ -575,3 +575,275 @@ config = do
...
+import Config
+import Data.ByteString.Base64 as Base64
+
+instance Controller ImageStyleController where
+    action RenderImageStyleAction { width, height, originalImagePath, signed } = do
+        let size = show width <> "x" <> show height
+
+        -- Verify the token, and deny or allow access based on the result.
+        -- Also, deny access if there's `../` in the path, to prevent traversal attacks.
+        accessDeniedUnless (rsaSignatureMatches (originalImagePath <> size) signed || not (Text.isInfixOf "../" originalImagePath))
```
mpscholten, line 792: Shouldn't this be an `&&` here?

```suggestion
        accessDeniedUnless (rsaSignatureMatches (originalImagePath <> size) signed && not (Text.isInfixOf "../" originalImagePath))
```

amitaibu: Yikes :man_facepalming:

Guide/file-storage.markdown
amitaibu: Got it working, thanks

Guide/file-storage.markdown
```diff
@@ -575,3 +575,295 @@ config = do
...
+```
+
+```haskell
+-- Config/Config.hs
+import Crypto.PubKey.RSA as RSA
+import Control.Exception (catch)
+import qualified Data.ByteString as BS
+import Web.JWT
+
+data RsaPublicAndPrivateKeys = RsaPublicAndPrivateKeys { publicKey :: RSA.PublicKey, privateKey :: RSA.PrivateKey }
```
amitaibu, line 711: Maybe change to?

```suggestion
data RsaKeys = RsaKeys { publicKey :: RSA.PublicKey, privateKey :: RSA.PrivateKey }
```

Guide/file-storage.markdown
```diff
@@ -575,3 +575,295 @@ config = do
...
+
+config :: ConfigBuilder
+config = do
+    -- See https://ihp.digitallyinduced.com/Guide/config.html
+
```
amitaibu, line 716:
```suggestion
```

Guide/file-storage.markdown
mpscholten: Good idea 👍

amitaibu: Done

mpscholten (approved):
👍

Guide/file-storage.markdown
mpscholten: yes, that should work 👍

PR #1691, opened 2023-06-02
Remove Just prefix from query params

### LINKED ISSUE #1671: links to Actions with Maybe's get literal `Just` in query params
If I do

```hs
 | NewThingAction { redirect :: Maybe (Id Other) }  
```

and

```hs
        <a href={NewThingAction (Just user.id)}>New</a>
```

I get an unusable link
```
<a href="http://localhost:8000/NewThing?redirect=Just03142938-62d5-4de1-af90-49726ca98f35">
```
where I would expect 

```
<a href="http://localhost:8000/NewThing?redirect=03142938-62d5-4de1-af90-49726ca98f35">
```
for the Just case and

```
<a href="http://localhost:8000/NewThing">
```
for the Nothing case

amitaibu:
fixes #1671

 - [ ] Tests

### DIFF
--- a/IHP/RouterSupport.hs
+++ b/IHP/RouterSupport.hs
@@ -633,8 +633,9 @@ instance {-# OVERLAPPABLE #-} (Show controller, AutoRoute controller) => HasPath
                     |> \(kvs :: [(String, String)]) -> zip (showTerms action) kvs
                     -- If an Id type was present in the action, it will be returned as Nothing by @showTerms@
                     -- as we are not able to match on the type using reflection.
-                    -- In this case we default back to the @show@ representation.
-                    |> map (\(v1, (k, v2)) -> (k, fromMaybe v2 v1))
+                    -- In this case we default back to the @show@ representation, making sure to remove
+                    -- the "Just" prefix.
+                    |> map (\(v1, (k, v2)) -> (k, fromMaybe (cs $ Text.replace "Just" "" $ cs v2) v1))
                     |> map (\(k, v) -> if isEmpty v
                         then ""
                         else  k <> "=" <> URI.encode v)
@@ -645,7 +646,7 @@ instance {-# OVERLAPPABLE #-} (Show controller, AutoRoute controller) => HasPath
 -- | Parses the HTTP Method from the request and returns it.
 getMethod :: (?context :: RequestContext) => Parser StdMethod
 getMethod =
-    case parseMethod ?context.request.requestMethod of 
+    case parseMethod ?context.request.requestMethod of
         Left error -> fail (ByteString.unpack error)
         Right method -> pure method
 {-# INLINABLE getMethod #-}


amitaibu:
@mpscholten I've got the tests red, which is good - as I haven't fixed the route parsing yet :sweat_smile:  . Can you please point me in the right direction for the parsing part

amitaibu:
I suspect it's [here](https://github.com/digitallyinduced/ihp/blob/1c8756d85900bef453589142f53b806c6eb5dfc7/IHP/RouterSupport.hs#LL275C16-L275C16). Looks scary! :fearful:

amitaibu:
So if I understand the code correctly, if wanted to return a Just [here](https://github.com/digitallyinduced/ihp/blob/1c8756d85900bef453589142f53b806c6eb5dfc7/IHP/RouterSupport.hs#L288)

we'd have to do 

```diff
- Just uuid -> uuid |> unsafeCoerce |> Right
+ Just uuid -> Just uuid |> unsafeCoerce |> Right
```

But we can't know when it's a Maybe and when a regular. So one option would be keeping the `Just` in the URL, and parse that. So this URL will be valid `NewThing?redirect=Just03142938-62d5-4de1-af90-49726ca98f35` (note the Just in the query param)

IHP/RouterSupport.hs
```diff
@@ -286,8 +287,19 @@ parseFuncs parseIdType = [
                     |> fromASCIIBytes
                     |> \case
                         Just uuid -> uuid |> unsafeCoerce |> Right
-                        Nothing -> Left BadType { field = "", value = Just queryValue, expectedType = "UUID" }
+                        Nothing ->
+                            -- We couldn't parse the UUID, so try Maybe (Id record),
+                            -- where we have a @Just@ prefix before the UUID.
+                            queryValue
```
amitaibu, line 293: @mpscholten Is there a way to debug such a function and have it print the value of `queryValue` ?

amitaibu: Well, I've just fixed it `"Just" ` not "Just "`. But still would be nice to know if there's a way to debug

amitaibu:
After fiddling with the parser, I think it would be easier to keep the `Just` prefix and `Nothing` in the URL.

| Just | Nothing |
| -- | -- |
| `http://localhost:8000/NewThing?redirect=Justbbc7604b-09bd-4ce5-a0e4-92409d1de4d9` | `http://localhost:8000/NewThing?redirect=Nothing` |

Tests are passing now :smile:

mpscholten:
This implementation is inconsistent with how other `Maybe` types are working with AutoRoute. E.g. when I have a field `myField :: Maybe Int`, and I request `/MyAction?myField=` it will set `myField = Nothing`, and `/MyAction?myField=1337` means `Just 1337`.

This part of IHP is a bit tricky as you've seen already :D But there's likely a way we get this consistent with the other Maybe implementations

IHP/RouterSupport.hs
amitaibu: @mpscholten It's indeed a tricky problem. Is there a way to debug/print the values in those functions?

amitaibu:
I was looking at the code some more, but I don't know how to proceed.

Maybe I'm completely off, but I remember reading somewhere that in the past there was `Generic` that was added to data which was removed due to long comiple time. Could that be useful in our case, by adding it to to the `data FooController` ? Maybe with the GHC v9 we're using it would be quicker?

In short, I'm happy I was able to tackle it up to this point, but will need help/ someone to take it from here :v:

IHP/RouterSupport.hs
mpscholten: Check out https://hackage.haskell.org/package/base-4.18.0.0/docs/Debug-Trace.html#v:trace

mpscholten:
Yeah, this is a tricky problem. I've already tried it myself yesterday and couldn't find a quick solution so far. Will take a deeper look soon

amitaibu:
I just hit this one when implementing a Chat (with ihp openAI) on my app. Tried:

`AiChatsAction { maybeAiChatId :: !(Maybe (Id AiChat)) }`

Since it's not currently working, as a workaround I can change this to :

`AiChatsAction { AiChatId :: !(Id AiChat) }` and pass an empty UUID `0000-0000-...`. But it's not great.

amitaibu:
> as a workaround I can change this to :

Actually, I don't need a workaround (or this feature). I can simply go with 

```
= AiChatsAction
| NewAiChatAction
```

So I guess this feature is still needed from time to time, but likely there are workarounds.

mpscholten:
I run into this problem as well every once in a while. Hope we can fix this at some point 👍 I usually follow the same workaround

CSchank:
Just ran into this - I ended up converting everything in actions to be `Maybe Text` and then I converted back to UUIDs from there in the actions. A part of my Haskell soul died in doing so. But it works as expected, in terms of not including the parameters if they're `Nothing`. I wonder, if it can be done for `Text`, why can't it be done for UUIDs and other things? Shouldn't it work in the same way for all `Maybe` types, regardless of the type parameter?

mpscholten:
> Shouldn't it work in the same way for all Maybe types, regardless of the type parameter?

Yes. The current implementation of AutoRoute doesn't allow for it. It likely needs an internal refactoring to fix this issue

PR #1694, opened 2023-06-04
Update UPGRADE.md

mpscholten:


UPGRADE.md
```diff
@@ -169,6 +169,22 @@ Note that the upgrade will drop your existing _local_ database, so make sure to
...
+        ihp = ...;
+        haskellEnv = import "${ihp}/NixSupport/default.nix" {
+            ihp = ihp;
+            haskellDeps = (import ./devenv.nix { pkgs = {}; inputs = {}; config = {}; }).ihp.haskellPackages;
+            otherDeps = pkgs: (import ./devenv.nix { inherit pkgs; inputs = {}; config = {}; }).packages;
+            projectPath = ./.;
+        };
+    in
+        haskellEnv
+    ```
```
amitaibu, line 187:
```suggestion
        haskellEnv
    ```

This means that from now on when adding new packages, you need to do it in a single file - `devenv.nix`
```

amitaibu (commented):
Shall we replace on the IHP guide references from `default.nix` to `devenv.nix`?

mpscholten:
yes, makes sense 👍

mpscholten:
I've moved this to a tracking issue at https://github.com/digitallyinduced/ihp/issues/1695

amitaibu:
After changing the `default.nix` file - how does one indicate the `rev`? Also need to upgrade [this part](https://github.com/digitallyinduced/ihp/blob/master/CONTRIBUTING.md#running-an-application-against-a-local-copy-of-the-framework)

mpscholten:
> After changing the default.nix file - how does one indicate the rev?

It's now documented via https://github.com/digitallyinduced/ihp/pull/1697 

> Also need to upgrade [this part](https://github.com/digitallyinduced/ihp/blob/master/CONTRIBUTING.md#running-an-application-against-a-local-copy-of-the-framework)

I think everything there should still work, let me know in case you see any issues 👍

PR #1711, opened 2023-06-15
Guide on password handle for User edit

amitaibu:
- Improved guide on User create and edit
- Add `isEqual` field validation, with tests


IHP/ValidationSupport/ValidateField.hs
```diff
@@ -277,6 +277,18 @@ isGreaterThan min value = Failure ("has to be greater than " <> tshow min)
...
+--
+-- >>> isEqual "foo" "foo"
+-- Success
+--
+-- >>> isEqual "foo" "bar"
+-- Failure "has to be equal to \"foo\""
+isEqual :: (Show value, Eq value) => value -> value -> ValidatorResult
+isEqual expected value | value == expected = Success
+isEqual expected value = Failure ("has to be equal to " <> tshow expected)
+{-# INLINABLE isEqual #-}
```
amitaibu, line 290: Thanks Co-pilot for writing this for me :smile:

Guide/authentication.markdown
```diff
@@ -271,7 +281,77 @@ To create a user with a hashed password, you just need to call the hashing funct
                     redirectToPath "/"
 ```
 
-### Hashing a Password
+The view would look like this, removing the `failedLoginAttempts` that was generated by the IDE, and manually add a password confirmation input field.
+
+```haskell
+-- Web/View/Users/New.hs and Web/View/Users/Edit.hs
```
amitaibu, line 287:
```suggestion
-- Web/View/Users/New.hs
```

mpscholten (approved):
Thanks 👍

Guide/authentication.markdown
```diff
@@ -271,7 +281,78 @@ To create a user with a hashed password, you just need to call the hashing funct
...
+
+
+```haskell
+-- Web/View/Users/Edit.hs
+renderForm :: User -> Html
+renderForm user = formFor user [hsx|
+    {(textField #email)}
+    {(passwordField #passwordHash) {fieldLabel = "Password"}}
+
+    <input type="password" name="passwordConfirmation"/>
```
amitaibu, line 350: @mpscholten it could have been nicer if we would be able to write

```haskell
{(passwordField #passwordHash) {required = True, fieldLabel = "Password Confirmation", additionalAttributes = [("name", "passwordConfirmation")]}}
```

So the CssFramework would kick in. But now the [additionalAttributs](https://github.com/digitallyinduced/ihp/blob/1bf4fb248f29d9cd65e1b1b7013d34928e202c47/IHP/View/CSSFramework.hs#L112) don't override. 

Do you think they should be able to override?

amitaibu: I was able to achieve it with

`{(passwordField #passwordHash) {required = True, fieldLabel = "Password confirmation", fieldName = "passwordConfirmation", validatorResult = Nothing}}`

I'll update the docs

mpscholten: While `fieldName` works, I think the `additonalAttributes` should be able to override all the other attributes 👍  can you make a ticket for that?

Issue #1695, opened 2023-06-04
IHP v1.1.0 Release


mpscholten:
- [x] Upgrade internal projects to devenv and check that everything works fine
- [x] Update documentation to mention `devenv.nix` instead of `default.nix`
- [x] Update https://github.com/digitallyinduced/vscode-hsx to include new HSX comments
- [ ] Get nix dependencies below 10 GB of storage (see https://github.com/digitallyinduced/ihp/issues/1706)
- [x] Fix https://github.com/digitallyinduced/ihp/issues/1728

Release Checklist:
- [x] Pick a release commit
- [x] Merge into the `v1.1` branch
- [x] Upgrade active digitally induced projects over to 1.1 to make sure that everything is stable
- [x] Increment version numbers in `ihp.cabal`, `Paths_ihp.hs` and `ihp.nix`
- [x] Tag the release
- [x] Update the boilerplate
- [x] Test `ihp-new` still works
- [ ] Publish latest `ihp-new` version to `nixpkgs`
- [x] Announce the new release on newsletter, twitter, reddit, [haskell discourse](https://discourse.haskell.org/)

PR #1723, opened 2023-06-22
Add `NoRenderCommentNode` to hsx

### LINKED ISSUE #1722: Add non-render comments to HSX
Sometimes we'd like to add comments inside HSX, without them being printed out. Maybe with this syntax:

```haskell
[hsx|
<!-- This will print out to the HTML -->

{- This will NOT print out to the HTML -}
|]
```

amitaibu:
| hsx | Html |
| -- | -- |
| image | image |

fixes #1722

### DIFF
--- a/Guide/hsx.markdown
+++ b/Guide/hsx.markdown
@@ -228,9 +228,32 @@ Spaces and newline characters are removed where possible at HSX parse time.
 
 HTML Comments are supported and can be used like this:
 
+```haskell
+[hsx|
+<div>
+    <!-- Comment with be shown on HTML -->
+    <h1>Hello</h1>
+</div>
+|]
+
+```
+
+We also support a "no render" comments, that can be used for adding comments inside `hsx`, but without being
+rendered to HTML.
+
+```haskell
+[hsx|
+    <div>
+        {- Comment with be shown on HTML -}
+        <h1>Hello</h1>
+    </div>
+|]
+```
+
+Will render to:
+
 ```html
 <div>
-    <!-- Begin of Main Section -->
     <h1>Hello</h1>
 </div>
 ```
--- a/Guide/hsx.markdown
+++ b/Guide/hsx.markdown
@@ -231,7 +231,7 @@ HTML Comments are supported and can be used like this:
 ```haskell
 [hsx|
 <div>
-    <!-- Comment with be shown on HTML -->
+    <!-- Comment will be shown on HTML -->
     <h1>Hello</h1>
 </div>
 |]
@@ -244,7 +244,7 @@ rendered to HTML.
 ```haskell
 [hsx|
     <div>
-        {- Comment with be shown on HTML -}
+        {- Comment will not be shown on HTML -}
         <h1>Hello</h1>
     </div>
 |]
--- a/IHP/ServerSideComponent/HtmlDiff.hs
+++ b/IHP/ServerSideComponent/HtmlDiff.hs
@@ -49,6 +49,7 @@ diffNodes' path CommentNode { comment = oldComment } CommentNode { comment = new
         if oldComment == newComment
             then []
             else [UpdateComment { comment = newComment, path }]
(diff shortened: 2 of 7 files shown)


amitaibu:
@mpscholten Are you ok with the direction of the PR? If so - I tried adding a test, but I'm getting this error

image

mpscholten:
Yes, direction is rigth. I'd prefer the `{- Syntax -}` actually, as it's closer to the normal haskell multi line comments syntax.

The error is casued by the wrong entry point. If you look at the parser entry point, it will go into `parseChildren` first instead of `parseElement`. The `parseChildren` calls some other functions that eventually will interpret the `{` as the start of the haskell expression.

amitaibu:
> I'd prefer the {- Syntax -} actually

Same here. I've started with it, but it messes up the syntax highlight.

image

Can we (easily) change the syntax highlight code as well?

mpscholten:
Yes, it's up at https://github.com/digitallyinduced/vscode-hsx

mpscholten:
cc @s0kil maybe you can take a look at the syntax highlighting here?

amitaibu:
> The error is casued by the wrong entry point. If you look at the parser entry point, it will go into parseChildren first instead of parseElement

This code is for the `ServerSideComponent`. For the code inside `IHP.HSX.Parser`, if I understand correctly (and I don't - since I can't get it right), I need to have `hsxNoRenderComment` executed before `hsxSplicedAttributes`, `hsxSplicedValue` and `hsxSplicedNode` are executed.

That's why I've placed it here:

https://github.com/digitallyinduced/ihp/blob/f303ecce55819795e9d982d26dfc637f2ca7f282/ihp-hsx/IHP/HSX/Parser.hs#L82-L83

What am I'm missing?

amitaibu:
I'm suspecting my local is not showing the correct info:

1. GitHub actions shows tests are green.
2. I've commented out my test on local, and when I try to load it I get

image

This branch is updated with `master`, that got this test [here](https://github.com/digitallyinduced/ihp/pull/1719/files#diff-542a4402951db912e76097f9bbf751911038ed62043794c1507491778503c2efR184-R188).

I've ran `nix-shell --run 'make -B build/ihp-lib'` and my project's flake.nix is pointing to latest IHP `master`.

Did I miss a step?

amitaibu:
> Did I miss a step?

It was a case of `rm -rf .direnv .devenv` from inside the IHP dir. And now it's passing locally for me as well

Guide/hsx.markdown
```diff
@@ -228,9 +228,32 @@ Spaces and newline characters are removed where possible at HSX parse time.
 
 HTML Comments are supported and can be used like this:
 
+```haskell
+[hsx|
+<div>
+    <!-- Comment with be shown on HTML -->
```
mpscholten, line 234:
```suggestion
    <!-- Comment will be shown on HTML -->
```

Guide/hsx.markdown
```diff
@@ -228,9 +228,32 @@ Spaces and newline characters are removed where possible at HSX parse time.
...
+
+```
+
+We also support a "no render" comments, that can be used for adding comments inside `hsx`, but without being
+rendered to HTML.
+
+```haskell
+[hsx|
+    <div>
+        {- Comment with be shown on HTML -}
```
mpscholten, line 247:
```suggestion
        {- Comment will not be shown on HTML -}
```

mpscholten:
Regarding the error: It could be the case that ihp-hsx is not loaded from source but instead the compiled binary is used. Inside ghci you can run `:set -iihp-hsx` to add the `ihp-hsx` directory to the load path manually

mpscholten (approved):
Thanks! Another nice feature for IHP 1.1 💯

amitaibu:
:sunglasses: 

Before 1.1, we need to remember to have https://github.com/digitallyinduced/vscode-hsx adapted.

mpscholten:
👍 added to https://github.com/digitallyinduced/ihp/issues/1695

PR #1724, opened 2023-06-24
Add docs about GitHub Action

amitaibu:


Guide/testing.markdown
```diff
@@ -163,3 +170,62 @@ tests = aroundAll (withIHPApp WebApplication config) do
...
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  tests:
+    name: Run Tests
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: cachix/install-nix-action@v20
+      with:
+        nix_path: nixpkgs=https://github.com/NixOS/nixpkgs/archive/51bcdc4cdaac48535dabf0ad4642a66774c609ed.tar.gz
```
amitaibu, line 203: Not sure what this is needed for

Guide/testing.markdown
```diff
@@ -163,3 +170,62 @@ tests = aroundAll (withIHPApp WebApplication config) do
...
+
+    # Use the cachix cache for faster builds.
+    - name: Cachix Init
+      uses: cachix/cachix-action@v12
+      with:
+        name: digitallyinduced
+        skipPush: true
```
amitaibu, line 210: Maybe can be removed, as I think only you have write access?

Guide/testing.markdown
mpscholten: This specifies the nixpkgs revision. If you set this to the nixpkgs revision used by IHP, you can save a bit of build time. But can be dropped as well as it's just an optimization

Guide/testing.markdown
mpscholten: `skipPush: true` needs to set to make it a read-only cache. Otherwise it will try to push any new binaries nix created at the end of the build (which will fail due to missing write access)

Guide/testing.markdown
mpscholten: The failures should be unrelated to this change. In case they're related, there is a bug in IHP as IHP should always use it's internal nixpkgs revision instead of the global one

amitaibu:
@mpscholten reviving this. We had some effort in the past with https://github.com/amitaibu/ihp-landing-page/pull/22 , but I think in the meanwhile we can get this one in.

mpscholten (approved):
Looks good, thanks 👍

PR #1725, opened 2023-06-25
`isActiveAction` should take query string into account

amitaibu:


Test/ViewSupportSpec.hs
```diff
@@ -0,0 +1,111 @@
...
+    action TestAction = do
+        renderPlain "TestAction"
+    action TestWithParamAction { .. } = do
+        let output = [plain|
+            isActiveAction #{param}: #{isActiveAction $ TestWithParamAction param}
+            isActiveAction bar: #{isActiveAction $ TestWithParamAction "bar"}
+
+            isActivePath #{param}: #{isActivePath $ "/test/TestWithParam?param=" <> param}
+            isActivePath bar: #{isActivePath ("/test/TestWithParam?param=bar" :: Text)}
+        |]
```
amitaibu, line 56: @mpscholten added tests so we have more confidence over the change! :beers: 

When I try to add

`isActiveController TestController: #{isActiveController @TestController}`

I get:

image

mpscholten: The function can only be called from a view. To fix the test you can add a view data structure and render it to HSX

amitaibu: > The function can only be called from a view

What indicates it? Is it the use of `fromFrozenContext`?

mpscholten: Yep exactly. The `renderHtml` does call `Context.freeze` before rendering the html https://github.com/digitallyinduced/ihp/blob/master/IHP/Controller/Render.hs#L60

mpscholten: Some explanation behind this is here https://github.com/digitallyinduced/ihp/blob/master/IHP/Controller/Context.hs#L14

amitaibu: @mpscholten I've tried to add a `View` in https://github.com/digitallyinduced/ihp/pull/1725/commits/f790af3233b3cee01c1dfc6cc6e91a21b3a1318f  but I'm getting a 500 error, but I don't know how to get the error message. Trying to print the output of the body ```"" `shouldBe` body``` showed an empty string

image

amitaibu: @mpscholten If it's not an easy fix, I can remove the `isControllerActive` test, in favor of getting this in.

amitaibu: @mpscholten ^^

mpscholten: Sorry, forgot to comment here, pushed a fix to https://github.com/amitaibu/ihp/pull/4/files Once merged the tests should be green 👍

amitaibu: Thanks. @mpscholten can you give a short explanation on how you knew what this fix is?

mpscholten: I dumped the response body of the the 500 Internal Server Error response. It showed that the error was happening here https://github.com/digitallyinduced/ihp/blob/master/IHP/Controller/Session.hs#L162 That error should never happen. I was first confused why the session was even used, because no code in the tests is doing anything with the session. I placed many `error` calls in my local IHP codebase to trace it back. Turns out that [`renderHtml` calls `initFlashMessages` which then accesses the session](https://github.com/digitallyinduced/ihp/blob/master/IHP/Controller/Render.hs#L58). 

Then I looked thorugh the test code and checked what could be wrong with the session setup. I figured out that we never called the session middleware. After that it was an easy fix, just wiring up the middleware so that the session is available 👍

amitaibu: > I dumped the response body of the the 500 Internal Server Error response

How did you do it?

mpscholten: Changed the

```haskell
assertTextExists body response = do
    response.simpleStatus `shouldBe` status200
```

function to this:

```haskell
assertTextExists body response = do
    error (show response)
    response.simpleStatus `shouldBe` status200
```

amitaibu:
@mpscholten tests are passing, but it shows the build as failing.

mpscholten (approved):
Thanks, looks good 👍

amitaibu:
I think for that we'd need an `isActiveActionByBasePath` (or some better but still explicit name).

Issue #1714, opened 2023-06-19
Nix-Flakes: Dev-Server does not recognize file changes on new IHP version


InftyCat:
Both in the new disaya version and textcontent I have trouble to autoreload. F5 also does not work. The only thing that worked yet is restarting the server.

Eisfunke:
So when you change the source nothing happens at all? Is there any output in the terminal? Can you post the terminal output beginning from you starting the dev server?

InftyCat:
No, there is no output at all. Marc already had a look on the terminal output but I send it to you anyhow:

20:56:11 system      | devServer.1 started (pid=252603)
20:56:11 devServer.1 | IHP Version: 1.0.1
20:56:11 devServer.1 | waiting for server to shut down.... done
20:56:11 devServer.1 | server stopped
20:56:12 devServer.1 | GET /
20:56:12 devServer.1 |   Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8
20:56:12 devServer.1 |   Status: 302 Found 0.183559304s
20:56:12 devServer.1 |
20:56:12 devServer.1 |  ===> AppState {postgresState = NotStarted, appGHCIState = NotStarted, statusServerState = NotStarted, databaseNeedsMigration = (..), lastSchemaCompilerError = (..)}
20:56:12 devServer.1 | UpdateStatusServerState Started
20:56:12 devServer.1 |  ===> AppState {postgresState = NotStarted, appGHCIState = NotStarted, statusServerState = Started, databaseNeedsMigration = (..), lastSchemaCompilerError = (..)}
20:56:12 devServer.1 | [("ihpVersion","1.0.1"),("os","linux"),("arch","x86_64"),("projectId","b7160aa803ab07b85d72149bb5cfbbce6f3e5c70d8352b0e5d8dab959e995d0545751f7ca77afd9b4f6e648e5d6c279e11b7df1f377c17b258b59f4c2d783c4d")]
20:56:12 devServer.1 | UpdatePostgresState Started
20:56:12 devServer.1 |  ===> AppState {postgresState = Started, appGHCIState = NotStarted, statusServerState = Started, databaseNeedsMigration = (..), lastSchemaCompilerError = (..)}
20:56:12 devServer.1 | GHCI: :script build/ihp-lib//applicationGhciConfig
20:56:12 devServer.1 | GHCI: :set prompt ""
[...]

PR #1726, opened 2023-06-26
Try to remove indirection for devServer process

Eisfunke:
This is an wild guess at fixing #1714 by trying to remove the inderection via a devenv script for running the DevServer. Will have to try it out to see if it helps.


mpscholten:
Is this still needed?

Eisfunke:
Nope, the error was something else, and fixed in #1733

Issue #1734, opened 2023-07-01
`accessDeniedWhen` and `accessDeniedUnless` should return 403


amitaibu:
Currently, they return 500.

![image](https://github.com/digitallyinduced/ihp/assets/125707/cd80928a-44fc-4ea6-9de2-0c3a5375d199)

@mpscholten I suppose we should have a `buildAccessDeniedResponse` similar to how we handle 404?

https://github.com/digitallyinduced/ihp/blob/0bf5b47e4adc1ad8825180d1627565beab936db1/IHP/ErrorController.hs#L60-L61

PR #1736, opened 2023-07-01
Add custom 403 response

amitaibu:
fixes https://github.com/digitallyinduced/ihp/issues/1734


IHP/AuthSupport/Authorization.hs
```diff
@@ -5,21 +5,21 @@ Copyright: (c) digitally induced GmbH, 2020
 -}
 module IHP.AuthSupport.Authorization where
 
-import IHP.Prelude
+import IHP.Controller.Render
```
amitaibu, line 8: @mpscholten I might need your help in untangling this part which creates a cyclic import. I couldn't find an easy way 

image

Another option is to move `accessDeniedUnless` into `IHP.ErrorController`

mpscholten: What do you think about moving all the functions related to 403 responses (including `accessDeniedUnless`) to it's own `IHP.Controller.AccessDenied` module?

amitaibu: > What do you think about moving all the functions related to 403 responses

I'm not sure.

1. IMO we should keep `handleAccessDeniedFound` in `IHP/ErrorController.hs` next to the `handleNotFound` - they are very similar, and maybe we can even create helper functions to reduce them.
2. `renderNotFound` and `renderAccessDenied` also feel natural inside `IHP/Controller/Render.hs`

So I think the question is only about `accessDeniedUnless` and `accessDeniedWhen`.  Why not inside `IHP.ErrorController`?

mpscholten: The `IHP.ErrorController` should only have code related to actually error handling. `accesDeniedUnless` is not directly related to that.

Maybe we can move `accessDeniedUnless` to `IHP.ControllerSupport`?

amitaibu: Yeah, I like that - let me try it out.

Another concern I have. We have this code that I think caches the the 404 pages?

https://github.com/digitallyinduced/ihp/blob/45902cce813a378f413f05628defec2e1e493a7d/IHP/Server.hs#L100

But there's no `Static.ss403Handler`. Should we create a PR for [Network.Wai.Application.Static](https://hackage.haskell.org/package/wai-app-static-3.1.7.4/docs/Network-Wai-Application-Static.html)?

mpscholten: The `ss404Handler` is called whenever a file is not found in `static/` that should match the request.

With latest master IHP routing prio is:
- app routes (e.g. controllers and actions)
- files that exist in `static/`
- files that exists in `IHP/static`
- when nothing matches, render 404 (via `ss404Handler`)

amitaibu: > Maybe we can move accessDeniedUnless to IHP.ControllerSupport?

I think the place is right, but still got a cyclic import. I don't have a clear mental model of who should be importing what.

amitaibu: @mpscholten How about `IHP.Controller.Render`? As it's really just a render function in the end

mpscholten: Hm, doesn't feel like it belongs there. I think this might get tricky when we don't want to move it into it's own dedicated module.

If we end up with something like `IHP.Controller.AccessDenied` we should also consider doing `IHP.Controller.NotFound` btw. The `ErrorController` module is already quite large, so it would be good to split up (this helps with compile times as well)

amitaibu: Ok, I'll try so separate to `IHP.Controller.AccessDenied` and `IHP.Controller.NotFound`

amitaibu: I'll also add `notFoundWhen`. Sometimes one would like to return 404 instead of 403 -- to prevent sniffing URLs and knowing they hit the right one by getting a 403.

IHP/ErrorController.hs
```diff
@@ -210,6 +114,7 @@ postgresHandler exception controller additionalInfo = do
         handlePostgresOutdatedError exception errorText = do
             let ihpIdeBaseUrl = ?context.frameworkConfig.ideBaseUrl
             let title = [hsx|Database looks outdated. {errorText}|]
+            -- @todo: Causes error
```
amitaibu, line 117: @mpscholten A bunch of surprising errors as I try to split modules.

image

amitaibu: I think again ihp-hsx isn't loaded properly for me. I'll try to grab later `master`

amitaibu: vscode is showing me an error, which tells me it's somehow not updated

image

Changing `flake.nix` to lastest `master` didn't help, also `rm -rf .devenv .direnv`. 

@mpscholten when you get a chance, can you please check if you see the same error?

amitaibu: I did `nix flake update` and the error on VScode is gone. But the compiler error [from above](https://github.com/digitallyinduced/ihp/pull/1736/files#r1248972353) is still there

mpscholten: I could not reproduce this error. I've pushed a fix for all the other errors I did hit when running locally. It's up at https://github.com/amitaibu/ihp/pull/3

It seems HSX is not correctly loaded from the ihp-hsx directory. Inside my `ghci` I had to type this to get it working:

```
:set -iihp-hsx
:set -package ghc
:l Test/Main
```

amitaibu:
@mpscholten We need to untangle more, but I need some advice here:

`ControllerSupport` -> `ErrorController` -> `NotFound`/`AccessDenied`

But they need `respondAndExit` exposed in `ControllerSupport`

amitaibu:
I'll try to move `respondAndExit` to own file

amitaibu:
@mpscholten I've added `IHP.Controller.Response` but need your approval it's the right direction, as it's likely a breaking change. On the other hand, it's cleans `IHP.ControllerSupport` further.

IHP/Controller/AccessDenied.hs
```diff
@@ -0,0 +1,158 @@
+module IHP.Controller.AccessDenied
+    ( accessDeniedWhen
+    , accessDeniedUnless
+    , handleAccessDeniedFound
+    , buildAccessDeniedResponse
+    )
+ where
```
mpscholten, line 7: IHP code typically doesn't have indentation here

```suggestion
( accessDeniedWhen
, accessDeniedUnless
, handleAccessDeniedFound
, buildAccessDeniedResponse
)
where
```

IHP/Controller/AccessDenied.hs
```diff
@@ -0,0 +1,158 @@
...
+-- __Example:__ Checking a user is the author of a blog post.
+--
+-- > action EditPostAction { postId } = do
+-- >     post <- fetch postId
+-- >     accessDeniedWhen (post.authorId /= currentUserId)
+-- >
+-- >     renderHtml EditView { .. }
+--
+-- This will throw an error and prevent the view from being rendered when the current user is not the author of the post.
+accessDeniedWhen :: (?context::ControllerContext) => Bool -> IO ()
```
mpscholten, line 33:
```suggestion
accessDeniedWhen :: (?context :: ControllerContext) => Bool -> IO ()
```

IHP/Controller/AccessDenied.hs
```diff
@@ -0,0 +1,158 @@
...
+-- __Example:__ Checking a user is the author of a blog post.
+--
+-- > action EditPostAction { postId } = do
+-- >     post <- fetch postId
+-- >     accessDeniedUnless (post.authorId == currentUserId)
+-- >
+-- >     renderHtml EditView { .. }
+--
+-- This will throw an error and prevent the view from being rendered when the current user is not the author of the post.
+accessDeniedUnless :: (?context::ControllerContext) => Bool -> IO ()
```
mpscholten, line 47:
```suggestion
accessDeniedUnless :: (?context :: ControllerContext) => Bool -> IO ()
```

IHP/Controller/NotFound.hs
```diff
@@ -0,0 +1,158 @@
+module IHP.Controller.NotFound
+    ( notFoundWhen
+    , notFoundUnless
+    , handleNotFound
+    , buildNotFoundResponse
+    )
+ where
```
mpscholten, line 7: Same indentation issue here

IHP/Controller/Render.hs
```diff
@@ -100,7 +101,21 @@ renderJson' additionalHeaders json = respondAndExit $ responseLBS status200 ([(h
 --
 renderNotFound :: (?context :: ControllerContext) => IO ()
 renderNotFound = do
```
mpscholten, line 103: This is the same as  `IHP.Controller.NotFound.renderNotFound`, can we remove this one in `IHP.Controller.Render`?

IHP/Controller/Render.hs
```diff
@@ -100,7 +101,21 @@ renderJson' additionalHeaders json = respondAndExit $ responseLBS status200 ([(h
...
+--
+-- This can be useful e.g. when an entity cannot be access:
+--
+-- > action ExampleAction = do
+-- >     renderAccessDenied
+--
+-- You can override the default not found error page by creating a new file at @static/403.html@. Then IHP will render that HTML file instead of displaying the default IHP access denied page.
+--
+renderAccessDenied :: (?context :: ControllerContext) => IO ()
+renderAccessDenied = do
```
mpscholten, line 117: Same here, duplicate of `IHP.Controller.AccessDenied.renderAccessDenied`

ihp.cabal
```diff
@@ -187,12 +186,13 @@ library
         , IHP.Controller.Layout
         , IHP.Controller.BasicAuth
         , IHP.Controller.Cookie
+        , IHP.Controller.AccessDenied
+        , IHP.Controller.NotFound
```
mpscholten, line 190:
```suggestion
        , IHP.Controller.NotFound
        , IHP.Controller.Response
```

mpscholten (commented):
Looks good 👍

mpscholten:
> I've added IHP.Controller.Response and it's compiling! But it needs your approval that it's in the right direction, as it's likely a breaking change. On the other hand, it cleans IHP.ControllerSupport further.

Looks like the right direction to me 👍 As these functions are typically imported via some Prelude, it shouldn't cause much breaking in most IHP app

IHP/Controller/AccessDenied.hs
amitaibu: Wish we had an `Elm format` equivalent, that "simply works" without any settings

mpscholten: yeah true :D

amitaibu:
I've updated [my project](http://github.com/amitaibu/ihp-landing-page/) to the latest  IHP `master`. Trying to run IHP tests, it's takes much longer than before. I don't have exact stats, but I think it's about ~40 sec, rather than previously ~2 sec.

Known issue?

mpscholten:
This shouldn't happen. What command do you run for the tests?

amitaibu:
From IHP core's dir:

```
ghci
:l Test/Main
main
```

mpscholten:
That's strange. For me it takes a couple of seconds to do the `:l Test/Main`, but after that the `main` is pretty quick. Can you record a video of the test run?

amitaibu:
[2023-07-18-10-30-43.webm](https://github.com/digitallyinduced/ihp/assets/125707/a620d52c-b30c-4031-b738-38a50408701d)

mpscholten:
Thanks 👍 

Can you adjust the `Test/Main.hs` like this:

```haskell

main :: IO ()
main = do
    putStrLn "Main entered"
    hspec do
        Test.IDE.SchemaDesigner.CompilerSpec.tests
        Test.IDE.SchemaDesigner.ParserSpec.tests
        Test.IDE.SchemaDesigner.Controller.EnumValuesSpec.tests
        Test.IDE.SchemaDesigner.Controller.HelperSpec.tests
        Test.IDE.SchemaDesigner.Controller.ValidationSpec.tests
        Test.ValidationSupport.ValidateFieldSpec.tests
        Test.IDE.CodeGeneration.ControllerGenerator.tests
        Test.IDE.CodeGeneration.ViewGenerator.tests
        Test.IDE.CodeGeneration.MailGenerator.tests
        Test.IDE.CodeGeneration.JobGenerator.tests
        Test.HSX.QQSpec.tests
        Test.NameSupportSpec.tests
        Test.HaskellSupportSpec.tests
        Test.HSX.ParserSpec.tests
        Test.View.CSSFrameworkSpec.tests
        Test.View.FormSpec.tests
        Test.Controller.ContextSpec.tests
        Test.Controller.ParamSpec.tests
        Test.SchemaMigrationSpec.tests
        Test.ModelSupportSpec.tests
        Test.SchemaCompilerSpec.tests
        Test.QueryBuilderSpec.tests
        Test.RouterSupportSpec.tests
        Test.ServerSideComponent.HtmlParserSpec.tests
        Test.ServerSideComponent.HtmlDiffSpec.tests
        Test.FileStorage.MimeTypesSpec.tests
        Test.DataSync.DynamicQueryCompiler.tests
        Test.IDE.SchemaDesigner.SchemaOperationsSpec.tests
        Test.IDE.CodeGeneration.MigrationGenerator.tests
        Test.Controller.CookieSpec.tests
        Test.PGListenerSpec.tests
        Test.SEO.Sitemap.tests
```

And then do another video?

amitaibu:
"Main entered" doesn't show for a long time until tests start to run.

[2023-07-18-11-50-38.webm](https://github.com/digitallyinduced/ihp/assets/125707/225fa629-9b18-4296-8eda-fbd88fcf45ae)

mpscholten:
Ok thanks! I think the issue causing this is https://gitlab.haskell.org/ghc/ghc/-/issues/23415#note_499590 (if you watch the video on the gitlab it's the same behaviour) For me the loading time is a bit shorter, but I also have a delay there. Maybe the new nix setup is somehow adding to this bug.

amitaibu:
@mpscholten I think something is wrong with the GH action:

image


https://github.com/digitallyinduced/ihp/actions/runs/5589604810/jobs/10218133895?pr=1736#step:5:2784

amitaibu:
Tests are passing, but the build is showing as failed. Anyway, the last thing for the PR - I'll update the docs.

ihp.cabal
mpscholten: @amitaibu  `IHP.Controller.Response` still needs to be added here to fix the build

amitaibu:
I've updated the docs, and the tests are passing. Not sure why other builds [have not](https://github.com/digitallyinduced/ihp/actions/runs/5689495968/job/15421106008?pr=1736#step:5:2774)

mpscholten:
Can you run `nix build` locally in the IHP directory? This will output some instructions to run `nix log ...`. If you run that `nix log` command then, it will show the full error (you might need to scroll a bit up to find it, but it's typically highlighted red in my terminal)

amitaibu:
I'm not getting instructions:

image

IHP/ControllerPrelude.hs
```diff
@@ -82,3 +86,4 @@ import IHP.FileStorage.Preprocessor.ImageMagick
 import IHP.Pagination.ControllerFunctions
 import IHP.HSX.QQ (hsx)
 import IHP.HSX.ToHtml ()
+import qualified Network.TLS as IHP.Controller
```
mpscholten, line 89:
```
[115 of 203] Compiling IHP.ControllerPrelude ( IHP/ControllerPrelude.hs, dist/build/IHP/ControllerPrelude.o, dist/build/IHP/ControllerPrelude.dyn_o )

IHP/ControllerPrelude.hs:89:1: error:
    Could not load module ‘Network.TLS’
    It is a member of the hidden package ‘tls-1.5.8’.
    Perhaps you need to add ‘tls’ to the build-depends in your .cabal file.
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
89 | import qualified Network.TLS as IHP.Controller
```

This line here is causing the error @amitaibu

mpscholten:
> I'm not getting instructions:

Oh same here, looks like we don't export any kind of package from the IHP framework itself.

I did get it to work by putting this into an empty IHP project and then setting `ihp.url = "path:///Users/marc/digitallyinduced/ihp";` in the project `flake.nix`

amitaibu:
Fixed the error. The [new error](https://github.com/amitaibu/ihp/actions/runs/5700690346/job/15450719318#step:7:56) looks related to the fact my branch is on my own fork, not on IHP itself

image

mpscholten:
Just pushed a fix for the github actions onto master via https://github.com/digitallyinduced/ihp/commit/8c44e841df5b32c7d12bb483a5502ff487302087 Can you merge master into this branch and then let the github actions rerun?

amitaibu:
@mpscholten there's a new error on the [ARM64 build](https://github.com/digitallyinduced/ihp/actions/runs/5708083099/job/15473709580#step:10:37)

amitaibu:
Still red

mpscholten:
I could fix this for the mac build, but the ubuntu build is still failing. I guess it's something with the github action secrets. Likely the secret key for cachix is not shared with your fork. Will merge it now 👍

mpscholten (approved):

PR #1737, opened 2023-07-02
Add `TransversingCSS` module to allow tests with CSS selectors

### LINKED ISSUE #1735: Testing helpers, add traversing over an HTML, and finding an element by CSS selector
Yesod has some [handy functions](https://github.com/yesodweb/yesod/blob/197ecb409fcc71d90cc353f13d6437ee73845a96/yesod-test/Yesod/Test.hs#L202-L207).

My goal is to be able to do the equivalent to this [PHP example](https://github.com/Gizra/drupal-starter/blob/0ce1b43273fa1cf970e7e90138967a78287396d1/web/modules/custom/server_general/tests/src/ExistingSite/ServerGeneralParagraphCtaTest.php#L68-L72).

amitaibu:
fixes #1735

1. Currently, it's a shameless copy/ paste from [here](https://github.com/yesodweb/yesod/blob/393954d802f16e88db8a72aa143f3aaf813fa277/yesod-test/Yesod/Test/TransversingCSS.hs) and [here](https://github.com/yesodweb/yesod/blob/393954d802f16e88db8a72aa143f3aaf813fa277/yesod-test/Yesod/Test/CssQuery.hs)
2. It Would have been nicer if Yesod had exposed it as an importable package. Makes sense? If so, I'll open an issue there.
3. `html-conduit` seems to be needed. Did I add it in the correct places? How to re-build IHP with this new package?

### DIFF
--- a/IHP/Test/CssQuery.hs
+++ b/IHP/Test/CssQuery.hs
@@ -1,0 +1,97 @@
+module IHP.Test.CssQuery where
+
+import IHP.Prelude hiding (takeWhile)
+import Data.Text as Text (Text)
+import Data.Attoparsec.Text
+import Control.Applicative
+import Data.Char
+
+data SelectorGroup
+  = DirectChildren [Selector]
+  | DeepChildren [Selector]
+  deriving (Show, Eq)
+
+data Selector
+  = ById Text
+  | ByClass Text
+  | ByTagName Text
+  | ByAttrExists Text
+  | ByAttrEquals Text Text
+  | ByAttrContains Text Text
+  | ByAttrStarts Text Text
+  | ByAttrEnds Text Text
+  deriving (Show, Eq)
+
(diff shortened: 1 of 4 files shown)


mpscholten:
> It Would have been nicer if Yesod had exposed it as an importable package. Makes sense? If so, I'll open an issue there.

Yes, that seems useful to me 👍 

> html-conduit seems to be needed. Did I add it in the correct places? How to re-build IHP with this new package?

In `ihp.nix` it needs to be used somewhere else a bit lower in the file as well, so it should be two places where it's added in `ihp.nix` (once to declare it as the dependency at the top level of the file, and then somewhere in the file to actually use it)

amitaibu:
Created issue on Yesod - https://github.com/yesodweb/yesod/issues/1810

PR #1741, opened 2023-07-06
Install `cachix` command

amitaibu:
We should also let people know they need to install `direnv` and hook to their shell- or is it changing with the new flake?


Eisfunke:
I think that's not part of the migration instructions because you needed direnv and cachix before as well, so you don't need to install them as part of the migration.

What you won't need to install anymore with the new flake from #1705 is actually devenv itself, as the shell environment pulls that in itself, so that part of the migration instructions will be removed with that PR (and replaced with enabling flake support in Nix).

Eisfunke:
Although, now that I'm thinking about it, I actually think we don't need users to install cachix at all with the new flake, as the substituters are included in the flake's `nixConfig` attribute, so users shouldn't need the `cachix` command for configuring the substituters anymore. :thinking:

PR #1752, opened 2023-07-17
Improved ihp lib

mpscholten:
This PR removes all of `build/ihp-lib` and replaces with an env var managed by devenv.

Fixes https://github.com/digitallyinduced/ihp/issues/1751


mpscholten:
> As a workaround, I'm setting export IHP_LIB=IHP/lib/IHP in .envrc file.

This is actually the solution I had in mind as well 👍

PR #1755, opened 2023-07-18
Improve docs on how to get latest `master`

amitaibu:


CONTRIBUTING.md
```diff
@@ -41,42 +41,46 @@ To help us with that, you can run this from the root of your project `cd IHP &&
...
+Then every time you'd like to update to the latest master, you'll copy/paste the [flake.lock](https://github.com/digitallyinduced/ihp-boilerplate/blob/master/flake.lock) to your own `flake.lock` and run:
 
 ```
-main
+nix flake update
+direnv allow
 ```
 
-After you have made modifications to files inside `IHP`, you need to press `CTRL + C` to stop the process running in `ghci` and then type `:r` to refresh the haskell modules. Now type `main` to start the server again.
+Note that the `flake.lock` on ihp-boilerplate's is rebuilt by IHP's GitHub action whenever a new commit is done on `master`. So be aware that if there's a new commit on master, and its GitHub action isn't done yet, then there will be a mismatch between IHP's commit and the `flake.lock`. You will notice in this case that trying to rebuild IHP takes a really long time, as it will try to compile it, instead of getting it from cachix. In that case, you can wait for the GitHub action to complete or point to a specific IHP commit.
```
amitaibu, line 64: > or point to a specific IHP commit.

@mpscholten Right now the [flake.lock](https://github.com/digitallyinduced/ihp-boilerplate/blob/master/flake.lock) commit shows the IHP commit message. Would be nice to also include the commit hash.

So it's easier to match an IHP commit and the right ihp-boilerplate `flake.lock`

mpscholten: I think the `flake.lock` can be ignored mostly now that the cachix situation is fixed. To update to master you only need to make sure that the url is set to ` ihp.url = "github:digitallyinduced/ihp";` and then run `nix flake update`. The `flake.lock` doesn't need to be copied manually.

Eisfunke: Yeah, you should generally avoid touching lockfiles manually. We don't even need a pre-defined lockfile in the boilerplate repo, I opened a PR for that: https://github.com/digitallyinduced/ihp-boilerplate/pull/29

amitaibu: Even better, closing 👍

Eisfunke: Also: We include the boilerplate repo itself as input in the IHP flake and expose it as template. So conceivably in the future we could use the Nix-/flake-native `nix flake init github:digitallyinduced/ihp/v1.1` (or any other release or commit) in ihp-new.

If we keep the locked boilerplate commit in sync from the IHP side, it will then use the appropriate boilerplate commit and init a folder from that.

(advanced users could use `nix flake init` directly as well)

amitaibu: Actually, maybe we should keep a note about the mismatch of IHP commit, while GitHub action is being built? Because on that phase, cachix didn't cache yet.

amitaibu: Updated PR

CONTRIBUTING.md
```diff
@@ -41,42 +41,42 @@ To help us with that, you can run this from the root of your project `cd IHP &&
...
+Then every time you'd like to update to the latest master, you'll run:
 
 ```
-main
+nix flake update
+direnv allow
 ```
 
-After you have made modifications to files inside `IHP`, you need to press `CTRL + C` to stop the process running in `ghci` and then type `:r` to refresh the haskell modules. Now type `main` to start the server again.
+Note that there's a time window that there could be a mismatch between IHP commit and cachix's cache. When a PR is merged into IHP core, a GitHub action runs and updates `flake.lock` on IHP-boilerplate repository. This means that while GitHub action on `master` is still running you will notice that rebuilding IHP takes a long time. Instead of building from the cache, it will try to compile it. You can wait for the GitHub action to complete or point to a specific IHP commit.
```
mpscholten, line 61:
```suggestion
Note that it takes around 30 minutes for the IHP GitHub actions to prepare a binary build of IHP. If you run latest master and the GitHub actions aren't finished yet, you will notice that your computer needs to build IHP from scratch which takes a lot of time. You can wait for the GitHub action to complete or point to a specific IHP commit to avoid long build times.
```

CONTRIBUTING.md
```diff
@@ -41,42 +41,42 @@ To help us with that, you can run this from the root of your project `cd IHP &&
...
 
 ### Running the development server
 
-When making changes to the development tooling, follow the setup above, except don't start postgres (the IDE starts it automatically).
 
-Instead of starting your application, start the development server:
+When making changes to the development tooling, we have to start the server differently, without `devenv up`. We have to
+use `make console` to load your application together with the framework located in `IHP`. In a `nix-shell`:
```
mpscholten, line 67: Nix-shell is not needed anymore

```suggestion
use `make console` to load your application together with the framework located in `IHP`.
```

PR #1756, opened 2023-07-18
Add docs for enabling flakes, recommend direnv extension

Eisfunke:
Some doc stuff for the new flake architecture.

While nix-env-selector *should* work through the flake-compat default.nix, we should rather recommend the direnv extension for VSCode so it uses the correct environment. That extension also tends to work more reliably and can be used in non-nix projects as well, so it's the better choice IMO.

### DIFF
--- a/Guide/installation.markdown
+++ b/Guide/installation.markdown
@@ -166,9 +166,9 @@ If you want to try it out before making your own repo, use this button below:
 
 ## 3. Setting up your editor
 
-Also check this out if you're editor is already set up. You might miss a plugin that's recommended for IHP to work well.
+Also check this out if your editor is already set up. You might miss a plugin that's recommended for IHP to work well.
 
--   [VS Code](https://ihp.digitallyinduced.com/Guide/editors.html#using-ihp-with-visual-studio-code-vscode) **Don't have nix-env-selector installed? Read this link**
+-   [VS Code](https://ihp.digitallyinduced.com/Guide/editors.html#using-ihp-with-visual-studio-code-vscode) **Don't have the direnv extension installed? Read this link**
 -   [Sublime Text](https://ihp.digitallyinduced.com/Guide/editors.html#using-ihp-with-sublime-text)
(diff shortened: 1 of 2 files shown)


Guide/installation.markdown
```diff
@@ -112,6 +112,14 @@ Installing nix for IHP was done using [this guide](https://nathan.gs/2019/04/12/
 
 Note that nix can gradually grow to use several GB of disk space, especially after upgrading IHP. You can always run `nix-collect-garbage` in a `nix-shell` which will delete older/unused files.
 
+### Enabling flakes
```
mpscholten, line 115: Maybe we can add, that this step is optional. If you disable flakes in your nix.conf, you can still run `ihp-new` and then still use the project (I've tried this yesterday myself)

mpscholten: Yes when direnv is working, there's no manual need for `nix develop`

Guide/installation.markdown
```diff
@@ -112,6 +112,14 @@ Installing nix for IHP was done using [this guide](https://nathan.gs/2019/04/12/
...
+
+IHP uses Nix flakes. While already widely used, they are not yet enabled by default by the Nix install scripts. To enable them, eithe edit `~/.config/nix/nix.conf` (to enable flakes just for your user) or `/etc/nix/nix.conf` (to enable flakes globally) and add the following line:
```
amitaibu, line 117:
```suggestion
IHP uses Nix flakes. While already widely used, they are not yet enabled by default by the Nix install scripts. To enable them, either edit `~/.config/nix/nix.conf` (to enable flakes just for your user) or `/etc/nix/nix.conf` (to enable flakes globally) and add the following line:
```

Eisfunke: Fixed, thanks!

Guide/installation.markdown
Eisfunke: I added some explaination there, what do you think?

mpscholten (approved):

Issue #1760, opened 2023-07-19
Compile `migrate` along with `RunDevServer`?


kodeFant:
Is there a way of compiling the `migrate` script, for example through `nix build` for use inside a NixOS configuration? 

I have tried a bit to reverse-engineer it to work on a test server, but it seems a bit sealed inside of the dev environment, so workarounds I try feel a bit too hacky for it to make sense.

Would for example be great to have it available through `${ihpApp}/bin/migrate` through nix to avoid the need of tapping into the development shell just for running the migrate script on a production server

mpscholten:
It should work with `${ihp}/bin/migrate` actually

kodeFant:
Took it to the test now! Seems like it's not found.

Reproducable step: Made an executable in `configuration.nix` trying both to hook on to the `ihp` library and the `ihpApp` itself coming from the flake.nix.

```nix
  environment.systemPackages = with pkgs; [
    (pkgs.writeShellScriptBin "ihp-migrate" ''
      ${ihp}/bin/migrate
    '')
    (pkgs.writeShellScriptBin "ihp-app-migrate" ''
      ${ihpApp}/bin/migrate
    '')
  ];

```

Here's the result trying to run them:

```sh
[ship@test-server-one:~]$ ihp-app-migrate
/run/current-system/sw/bin/ihp-app-migrate: line 2: /nix/store/m8f52k0i4pqjnkv2xk148fs1kxw07xl8-app/bin/migrate: No such file or directory

[ship@test-server-one:~]$ ihp-migrate
/run/current-system/sw/bin/ihp-migrate: line 2: /nix/store/d46kna0fszi06zhagmmfzg35hw3fg90c-source/bin/migrate: No such file or directory
```

Eisfunke:
I'm a little confused as to why the migrate script package from #1762 works for you while `${ihp}/bin/migrate` doesn't, because the migrate script over there literally just calls `${ghcCompiler.ihp}/bin/migrate`.

Where exactly did you take the `ihp` variable from in your test with `${ihp}/bin/migrate`?

kodeFant:
It comes from the `ihp` of specialArgs that is passed into the `configuration.nix`

```nix
            flake.nixosConfigurations."test-server-one" = nixpkgs.lib.nixosSystem {
                system = "x86_64-linux";
                specialArgs = inputs // {
                    environment = "production";
                    ihp = ihp;
                    migrate = self.packages.x86_64-linux.migrate;
[...]

PR #1762, opened 2023-07-19
Added a migrate output to the IHP flake

mpscholten:
Potential fix for https://github.com/digitallyinduced/ihp/issues/1760

```bash
nix build .#migrate

# Now this runs the migration script, and automatically sets the working directory to the Project src
./result/bin/migrate
```

Likely (but untested) this can be used in nixos as:

```nix
environment.systemPackages = with pkgs; [ ihpApp.migrate ];
```

Then a global `migrate` command would be available

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -127,7 +127,6 @@ ihpFlake:
                         projectSrc = pkgs.nix-gitignore.gitignoreSource [] cfg.projectPath;
                     in
                         pkgs.writeScriptBin "migrate" ''
-                            cd ${projectSrc}
                             ${ghcCompiler.ihp}/bin/migrate
                         '';
 
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -120,6 +120,15 @@ ihpFlake:
                     includeDevTools = false;
                     optimized = false;
                 };
+
+                migrate =
+                    let
+                        projectSrc = pkgs.nix-gitignore.gitignoreSource [] cfg.projectPath;
+                    in
+                        pkgs.writeScriptBin "migrate" ''
+                            cd ${projectSrc}
+                            ${ghcCompiler.ihp}/bin/migrate
+                        '';
             };
 
             devenv.shells.default = lib.mkIf cfg.enable {
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -122,13 +122,9 @@ ihpFlake:
                 };
 
 
-                migrate =
-                    let
-                        projectSrc = pkgs.nix-gitignore.gitignoreSource [] cfg.projectPath;
-                    in
-                        pkgs.writeScriptBin "migrate" ''
-                            ${ghcCompiler.ihp}/bin/migrate
-                        '';
+                migrate = pkgs.writeScriptBin "migrate" ''
+                    ${ghcCompiler.ihp}/bin/migrate
+                '';
 
                 ihp-schema = pkgs.stdenv.mkDerivation {
                     name = "ihp-schema";


kodeFant:
Tested now, with `nix build --impure` in the project and added as system package on the server. Both results in this error:

```
> Makefile:30: *** IHP not found! Run the following command to fix this:    nix-shell --run 'make .envrc'    .  Stop.
```

mpscholten:
You're missing some steps from the UPGRADE.md (there's been some recent changes to the Makefile boilerplate)

kodeFant:
`ihpApp.migrate` didn't work.

I got it to run by adding the migrate script like this in the flake.nix.

```nix
            flake.nixosConfigurations."test-server-one" = nixpkgs.lib.nixosSystem {
                system = "x86_64-linux";
                specialArgs = inputs // {
                    environment = "production";
                    ihp = ihp;
                    migrate = self.packages.x86_64-linux.migrate;
                    ihpApp = self.packages.x86_64-linux.default;
                };
                modules = [
                    ./nixos/configuration.nix
                ];
            };
```

and then added `migrate` as arg to `configuration.nix` and added it as a system package.

```
{ config, pkgs, modulesPath, lib, environment, ihpApp, ihp, migrate, ... }:
{

  # ...
  
  # Add system-level packages for your server here
  environment.systemPackages = with pkgs; [
    migrate
  ];
  
  # ...
  
}
```

Now the error is 

```sh
migrate: Application/Migration: getDirectoryContents:openDirStream: does not exist (No such file or directory)
```

mpscholten:
Does the test app have atleast one migration? Otherwise the `Application/Migration` directory might not been created yet

kodeFant:
Ah of course, it didn't. Created now and it run properly :1

kodeFant:
It works smoothly, but there is one little issue:

Since it cds into the previously built `ihpApp`,  `projectSrc` is for the previous build.

This makes it impossble to run the proper migrations before the nixos-rebuild.

Maybe we could take responsibility ourselves for cd-ing into the project directory and that it only provides the `migrate` binary directly?

flake-module.nix
```diff
@@ -120,6 +120,15 @@ ihpFlake:
                     includeDevTools = false;
                     optimized = false;
                 };
+
+                migrate =
+                    let
+                        projectSrc = pkgs.nix-gitignore.gitignoreSource [] cfg.projectPath;
```
Eisfunke, line 126: This is supposed to get the root path of the IHP project, yes? You could get that via a `self` argument in the arguments in line 11, I believe

Eisfunke:
I'd recommend making this an app instead of an package, so you could directly run it with `nix run .#migrate`. Just put it under apps instead of packages, and add a `type = "app"`. See: https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-run.html#apps

Could look like this:

```
apps.default = {
      type = "app";
      program = pkgs.writeScriptBin "migrate" ''
          ${ghcCompiler.ihp}/bin/migrate
      '';
};
```

I would also recommend letting out the `cd`, by the way, that should be more flexible.

(also as a note, if I'm not mistaken, the migrate script isn't unique to a project but rather a IHP feature, so we could add it directly to the root IHP flake.nix so it could be used independently of having a specific project with `nix run github:digitallyinduced/ihp#migrate`, but that's something for another PR, probably together with expanding the root flake.nix in general)

kodeFant:
Tried out this in my fork based on @Eisfunke suggestion, and it seems to work well.

```nix
            apps.migrate = {
                type = "app";
                program = "${ghcCompiler.ihp}/bin/migrate";
            };
```

Then super nice to cd into project folder and run

```sh
nix run .#migrate
```

To me this looks like the perfect solution!

kodeFant:
One consequence of the app approach is that the whole IHP app is being built twice, here and during the rebuild phase, so less efficient (e.g. more disk space used, and longer build time if not hitting binary cache) as opposed to having it as a package that uses the same `ihpApp` used in NixOS rebuild.

mpscholten:
> With the package appoach, it's instantly present.

Then let's stick with the package approach for now. I've just removed the `cd PROJECT_DIR`

Eisfunke:
> One consequence of the app approach is that the whole IHP app is being built twice, here and during the rebuild phase, so less efficient (e.g. more disk space/cpu/ram used, and longer build time if not hitting binary cache).
> 
> With the package appoach, it's instantly present.
> 
> So for that reason, I'm leaning towards a package without the `cd` to project path.

I'm not sure *why* it'd be built twice, the app should just use the package that was built before, and why that doesn't happecn with packages. Apps and packages are both just derivations, apps just have a defined execution start point.

We could also add both, by the way: a package with just the script, and then an app, which just call the script from the migration script package. That would be maximally flexible.

flake-module.nix
mpscholten: How exactly would the code look like here then?

kodeFant:
> I'm not sure why it'd be built twice, the app should just use the package that was built before, and why that doesn't happecn with packages. Apps and packages are both just derivations, apps just have a defined execution start point.

@Eisfunke  I'm not sure why myself. Either it has something to do with that it's impure according to nix standards (I think mainly because of `import <nixpkgs> ` still present some places in the IHP nix files), or simply that a `nix build` is not using the same Nix store addresses as the one in the nixos configuration.

I'm simply not smart enough to figure this out 😄

flake-module.nix
kodeFant: I think this line is not needed anymore since we don't cd into the `projectSrc`

kodeFant:
```nix
migrate =  pkgs.writeScriptBin "migrate" ''
    ${ghcCompiler.ihp}/bin/migrate
  '';
`

File: flake-module.nix [hunk 2]
```diff
@@ -120,6 +120,15 @@ ihpFlake:
...
+                    in
+                        pkgs.writeScriptBin "migrate" ''
```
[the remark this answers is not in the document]
mpscholten: As long as this package is build with `nix build` it shouldn't cause any collision. I'd prefer to not introduce two names for the same binary, otherwise it will be confusing to other people in the future 👍

PR #1769, opened 2023-07-23
Update oauth with the new `flake.nix`

amitaibu:
I tried adding `ihp-oauth-google` to `flake.nix`, but I'm getting this error

image


mpscholten:
`ihp-oauth-google` is only available in IHP Pro builds. That's causing the error

mpscholten (approved):

PR #1772, opened 2023-07-25
Add make to devenv

Eisfunke:
(gnu)make is (still) needed for various stuff in the devenv, including the haskell IDE stuff.

I noticed that we don't yet have make added to the IHP devenv because don't have a global make install on my NixOS system as I rely on direnv and flakes to make it available where I need it. That led to unexpected and hard to debug errors in my VSCode haskell IDE stuff, because the hie-bios stuff is generated through a make task, which didn't work without make.

By adding make to the devenv, users don't need a global make install and we make sure it's available.

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -142,7 +142,7 @@ ihpFlake:
             };
 
             devenv.shells.default = lib.mkIf cfg.enable {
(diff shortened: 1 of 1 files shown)


mpscholten (approved):
👍 good catch

Issue #1776, opened 2023-07-27
SSC Accept nullary types


s0kil:
With:

```
data PolicyEditorController
    = OpenModalAction
    | CloseModalAction
    deriving (Data, Eq, Show)
```

Calling `OpenModalAction` will fail with:
```
Error in $: When parsing Web.Component.PolicyEditor.PolicyEditorController expected String but got Object.
Error in $: When parsing Web.Component.PolicyEditor.PolicyEditorController expected String but got Object.
Error in $: When parsing Web.Component.PolicyEditor.PolicyEditorController expected String but got Object.
```

mpscholten:
Duplicate of https://github.com/digitallyinduced/ihp/issues/923

Issue #923, opened 2021-06-14
SSC: Component with a single action doesn't work


mpscholten:
When only a single action is available the Aeson encoding is done differently and then the JS fails to generate a valid JS message causing a `Error in $: When parsing the constructor IncrementCounterAction of type Web.Component.Counter.CounterController expected Array but got Object.` on the server-side.

zacwood9:
Example code that ran into this issue:

```haskell
{-# LANGUAGE TemplateHaskell #-}
module Admin.Component.Payments where
import IHP.ViewPrelude hiding (fetch, query)
import Web.Controller.Prelude hiding (render, setState, getState)
import IHP.ModelSupport
import IHP.ServerSideComponent.Types
import IHP.ServerSideComponent.ControllerFunctions
data FormType
    = QuotePayment    { quote :: !Quote }
    | WorkshopPayment { workshop :: !Workshop }
    | ClassPayment    { _class :: !Class }
    | StandalonePayment
    | Initial
data FormState = FormState {
    payment :: !Payment,
    formType :: !FormType
}
data PaymentFormController
    = SelectQuoteAction
    | SelectWorkshopAction
    deriving (Eq, Show, Data)
$(deriveSSC ''PaymentFormController)
instance Component FormState PaymentFormController where
    initialState = FormState {
        payment = newRecord @Payment,
        formType = Initial
    }
    componentDidMount state = do
        pure state
    action state SelectQuoteAction = do
        pure state
    action state SelectWorkshopAction = do
        pure state
    render FormState { payment, formType } =
        case formType of
            Initial -> [hsx|
                <h6>Choose Payment Type</h6>
                <nav class="nav nav-pills flex-column flex-sm-row">
                    <button
                        class="btn btn-outline-primary flex-sm-fill text-sm-center nav-link"
                        onclick="callServerAction('SelectQuoteAction')"
                    >
                        Standalone
                    </button>
                    <button class="btn btn-outline-primary flex-sm-fill text-sm-center nav-link">Class</button>
                    <button class="btn btn-outline-primary flex-sm-fill text-sm-center nav-link">Workshop</button>
                </nav>
            |]
            _ -> [hsx|State not handled|]
```

PR #1777, opened 2023-07-27
SSC do not encode nullary constructors to string

s0kil:
Fixes issue https://github.com/digitallyinduced/ihp/issues/1776

### DIFF
--- a/IHP/ServerSideComponent/Controller/ComponentsController.hs
+++ b/IHP/ServerSideComponent/Controller/ComponentsController.hs
@@ -7,7 +7,7 @@ import IHP.ServerSideComponent.ControllerFunctions as SSC
 
 import qualified Data.Aeson as Aeson
 
-instance (Component component controller, FromJSON controller) => WSApp (ComponentsController component) where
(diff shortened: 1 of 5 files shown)


mpscholten:
Also fxies https://github.com/digitallyinduced/ihp/issues/923

IHP/ServerSideComponent/Controller/ComponentsController.hs
```diff
@@ -29,4 +29,7 @@ instance (Component component controller, FromJSON controller) => WSApp (Compone
 
                     nextState <- SSC.action currentState theAction
                     SSC.setState nextState
-                Left error -> putStrLn (cs error)
+                Left error -> do
+                    putStrLn "Failed Parsing Server Side Component Message As JSON"
+                    putStrLn (cs actionPayload)
+                    putStrLn (cs error)
```
mpscholten, line 35: Could we use `Log.info` here instead of `putStrLn`?

mpscholten (approved):
Thanks! Nice that this bug is finally fixed :)

Issue #936, opened 2021-06-24
SSC: Support Passing In `initialValue` During Instantiation


s0kil:
`componentFromState @Counter (Counter { value = 100 })` should use the provided  initial state instead of relying on the `initialValue` defined in the Counter component module.

MurakamiKennzo:
any updated here?

mpscholten:
Not yet 👍

MurakamiKennzo:
😭, What's the main problem here, I read some ssc code just now and think It's may not a tricky.

fidel-ml:
What about something as this:
```
class Component state action configData | state -> action, state -> configData where
  initialState :: configData -> state
  ...
```
Then you may have both static and dynamic configuration data for the components (as you are able to choose the `configData`...).
For example
```
instance Component Counter CounterController Int where
  initialState n = Counter { value = n }
```

mpscholten:
I like this idea. It's very close to how react.js works as well 👍

s0kil:
With `class Component state action configData | state -> action, state -> configData where` it will require ConfigData to be provided on every instance of the component, correct? `component @Counter configData`, I can no longer do `component @Counter`, or am I misunderstanding something?

mpscholten:
yep that's correct. If you set `configData = ()` you could use `component @Counter ()` as a workaround to get the old behaviour

fidel-ml:
You may also provide a `defaultConfig`, and let `component` use it, and another `componentWith` for the new behaviour...

mpscholten:
> You could maybe put the props as data attribute into the html and send it somehow with the first websocket call and set it before you change the state. But this is probably a stupid idea....

I think this might actually be a good option here.

The ControllerContext cannot be used as we're having two requests here. The first that renders the SSC `componentWithProps Counter { value = 100 }`. Then the JS connects via the WebSocket which is a second request.

[...]

PR #1778, opened 2023-07-27
Initial State Is Sent Back To The SSC

s0kil:
My attempt at fixing issue https://github.com/digitallyinduced/ihp/issues/936

Send the Initial State To the Browser, And send it back when the component is initialized.

This requires the component to be an instance of `FromJSON` and `ToJSON`.

```haskell
module Web.Component.PolicyEditor where

import Web.Component.Prelude
import Data.Aeson.TH

data PolicyEditor
    = PolicyEditor
          { maybeFacilityId :: Maybe (Id' "company_facilities")
          , maybePolicyId   :: Maybe (Id' "facility_policies")
          , modalOpen :: Bool
          }
    deriving (Data, Eq, Show)

$(deriveJSON defaultOptions 'PolicyEditor)

data PolicyEditorController
    = OpenModalAction
    | CloseModalAction
    deriving (Data, Eq, Show)

$(deriveSSC ''PolicyEditorController)

....
```

### DIFF
--- a/IHP/ServerSideComponent/Controller/ComponentsController.hs
+++ b/IHP/ServerSideComponent/Controller/ComponentsController.hs
@@ -7,7 +7,7 @@ import IHP.ServerSideComponent.ControllerFunctions as SSC
 
 import qualified Data.Aeson as Aeson
 
-instance (Component component controller, FromJSON controller) => WSApp (ComponentsController component) where
+instance (Component component controller, Aeson.FromJSON component, FromJSON controller, ToJSON controller) => WSApp (ComponentsController component) where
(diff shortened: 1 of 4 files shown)


mpscholten:
This would allow anyone to write any kind of component state which could be insecure. It might be better to cryptographically sign the initial state JSON and then check the signature in the component JSON decoder

mpscholten:
>  Is this a good approach in the first place, can't we save initial state in some ?context?

Yes, this would be even better. AutoRefresh is working similiar to this. So that might be a good inspiration for designing this

PR #1779, opened 2023-07-29
Guide on how to test emails with Mailhog API

amitaibu:
With Mailhog we can catch emails, and use the API to get the contents of the emails in tests.


mpscholten (approved):
Thanks 👍

PR #1786, opened 2023-08-02
Disable -fexpose-all-unfoldings

mpscholten:
This will speed up linking time on macOS

See https://gitlab.haskell.org/ghc/ghc/-/issues/23415


mpscholten:
This has been integrated into IHP already

PR #1790, opened 2023-08-03
Add guide on conditional `fill` and security concerns

amitaibu:
PR also changes the generator to always use `@'[` for `fill` instead of the `@[` -- as it seems it works regardless of the number of elements in the list.

### DIFF
--- a/Guide/validation.markdown
+++ b/Guide/validation.markdown
@@ -293,14 +293,13 @@ buildComment comment = comment
     |> fill @'["body", "commentModeration"]
     |> fillIsNew
     where
-        fillIsNew record
+        fillIsNew record =
+            if isNew record
             -- Record is new, so fill the `postId`.
-            | isNew record = fill @'["postId"] record
+            then fill @'["postId"] record
             -- Otherwise, leave the record as is.
-            | otherwise = record
+            else record
 ```
-
-We are using Haskell guards instead of the regular `IF` statement to make the code more consice.
 
 Next, imagine we have a `currentUserIsAdmin` indicating if the current user is an admin. We'd like to allow only admins to set the moderation status of a comment. So we'll allow `fill` on the `commentModeration` field only in that case:
 
@@ -318,16 +317,17 @@ buildComment comment = comment
     |> fillIsNew
     |> fillCurrentUserIsAdmin
     where
-        fillIsNew record
+        fillIsNew record =
+            if isNew record
             -- Record is new, so fill the `postId`.
-            | isNew record = fill @'["postId"] record
+            then fill @'["postId"] record
             -- Otherwise, leave the record as is.
-            | otherwise = record
+            else record
 
-        fillCurrentUserIsAdmin record
-            -- User is admin.
-            | currentUserIsAdmin = fill @'["commentModeration"] record
-            | otherwise = record
+        fillCurrentUserIsAdmin record =
+            if currentUserIsAdmin
+            then fill @'["commentModeration"] record
+            else record
 ```
 
 Let's finish with a final example. Let's assume there was also a `score` integer field between 1 - 5 that only the admin could set. As mentioned, we'd need to have a validation on the backend to ensure that the user didn't manipulate the form data. And use `fill` to ensure that a non-admin user can't set the score in the first place. Here's the final code, where we conditionally `fill` the `score` field only if the user is an admin, and perform validation on it:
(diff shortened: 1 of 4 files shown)


Guide/validation.markdown
```diff
@@ -250,6 +250,116 @@ user
...
+        fillIsNew record
+            -- Record is new, so fill the `postId`.
+            | isNew record = fill @'["postId"] record
+            -- Otherwise, leave the record as is.
+            | otherwise = record
+```
+
+We are using Haskell guards instead of the regular `IF` statement to make the code more consice.
+
+Next let's imagine we have a `currentUserIsAdmin` indicating if the current user is an admin. We'd like to only allow admins to set the moderation status of a comment. So we'll allowing `fill` on the `commentModeration` field only in that case:
```
amitaibu, line 305:
```suggestion
Next, imagine we have a `currentUserIsAdmin` indicating if the current user is an admin. We'd like to allow only admins to set the moderation status of a comment. So we'll allow `fill` on the `commentModeration` field only in that case:
```

Guide/validation.markdown
```diff
@@ -250,6 +250,116 @@ user
...
+            -- Otherwise, leave the record as is.
+            | otherwise = record
+
+        fillCurrentUserIsAdmin record
+            -- User is admin.
+            | currentUserIsAdmin = fill @'["commentModeration"] record
+            | otherwise = record
+```
+
+Let's finish with a final example. What if there was also a `score` integer field between 1 - 5, that only the admin could set. As mentioned, we'd need to have a validation on the backend to ensure that the user didn't manipulate the form data. And use `fill` to ensure that a non admin user can't set the score in the first place. Here's the final code, where we conditionally `fill` the `score` field only if the user is an admin, and perform a validation on it:
```
amitaibu, line 333:
```suggestion
Let's finish with a final example. Let's assume there was also a `score` integer field between 1 - 5 that only the admin could set. As mentioned, we'd need to have a validation on the backend to ensure that the user didn't manipulate the form data. And use `fill` to ensure that a non-admin user can't set the score in the first place. Here's the final code, where we conditionally `fill` the `score` field only if the user is an admin, and perform validation on it:
```

Guide/validation.markdown
```diff
@@ -250,6 +250,116 @@ user
...
+
+It's important to remember that any kind of validations you might have on the form level are not enough to ensure the security of your application. You should always have validations on the backend as well. This is because the user might manipulate the form data and send invalid data to your application.
+
+So if a field is disabled, or if you have an integer field with a min/max value, you should always have a validation on the backend as well.
+
+Another point to think about is that you don't have to always `fill` all fields in one go. Sometimes you'd like to conditionally `fill` based on the current user or based on the current logic.
+
+Let's see those examples in action. Let's say we have a `Comment` record that has a `postId` that references a `Post`, a `body` field, and a modeartion field allowing admin users to indicate if they are approved or rejected.
+
+Here's an excerpt from the `Schema.sql`:
```
amitaibu, line 263:
```suggestion
It's important to remember that any kind of validations you might have on the form level are not enough to ensure the security of your application. You should always have validations on the backend as well. The user might manipulate the form data and send invalid data to your application.

So if a field is disabled or has an integer field with a min/max value, you should always have a validation on the backend.

Another point is that you don't have to always `fill` all fields in one go. Sometimes you'd like to conditionally `fill` based on the current user or based on the current logic.

Let's see those examples in action. Let's say we have a `Comment` record that has a `postId` that references a `Post`, a `body` field, and a moderation field allowing admin users to indicate if they are approved or rejected.

Here's an excerpt from the `Schema.sql`:
```

Guide/validation.markdown
```diff
@@ -250,6 +250,116 @@ user
...
+buildComment comment = comment
+    |> fill @'["body", "commentModeration"]
+    |> fillIsNew
+    where
+        fillIsNew record
+            -- Record is new, so fill the `postId`.
+            | isNew record = fill @'["postId"] record
+            -- Otherwise, leave the record as is.
+            | otherwise = record
+```
```
mpscholten, line 301: Likely it would be better if we stick to normal `if .. then .. else ..` in the documentation to avoid more complex syntax beginners might not be aware about

amitaibu: Done

mpscholten (approved):
Great, thanks :)

Issue #1696, opened 2023-06-04
Support deployments via nixos-rebuild


mpscholten:
https://www.haskellforall.com/2023/01/announcing-nixos-rebuild-new-deployment.html

Now that we support nix flakes, we can use nixos-rebuild for a very nice deployment experience

PR #1792, opened 2023-08-05
deploy-to-nixos Command

mpscholten:
This branch enables easy deployments to any NixOS server that's reachable via SSH:

```bash
deploy-to-nixos myapp.example.com
```

Before that command the server needs to be configured in `flake.nix`:

```nix
{
    inputs = {
        ihp.url = "path:///Users/marc/digitallyinduced/ihp";
        nixpkgs.follows = "ihp/nixpkgs";
        flake-parts.follows = "ihp/flake-parts";
        devenv.follows = "ihp/devenv";
        systems.follows = "ihp/systems";
    };

    outputs = inputs@{ ihp, flake-parts, systems, nixpkgs, self, ... }:
        flake-parts.lib.mkFlake { inherit inputs; } {

            systems = import systems;
            imports = [ ihp.flakeModules.default ];

            perSystem = { pkgs, ... }: {
                ihp = {
                    enable = true;
                    projectPath = ./.;
                    packages = with pkgs; [
                        # Native dependencies, e.g. imagemagick
                    ];
                    haskellPackages = p: with p; [
                        # ....
                    ];
                };
            };

            flake.nixosConfigurations."myapp.example.com" = nixpkgs.lib.nixosSystem {
                system = "x86_64-linux";
                specialArgs = inputs;
                modules = [
                    "${nixpkgs}/nixos/modules/virtualisation/amazon-image.nix"
                    ihp.nixosModules.appWithPostgres
                    ({ ... }: {
                        security.acme.defaults.email = "me@example.com";

                        services.ihp = {
                            domain = "myapp.example.com";
                            migrations = ./Application/Migration;
                            schema = ./Application/Schema.sql; # TODO: Remove
                            fixtures = ./Application/Fixtures.sql; # TODO: Remove
                            sessionSecret = "xxx";
                        };
                    })
                ];
            };
        };
}
```

Migrations can be run by calling the global `migrate` command:

```bash
ssh myapp.example.com migrate
```

Cronjobs can be scheduled like this:

```nix
{
    inputs = {
        ihp.url = "path:///Users/marc/digitallyinduced/ihp";
        nixpkgs.follows = "ihp/nixpkgs";
        flake-parts.follows = "ihp/flake-parts";
        devenv.follows = "ihp/devenv";
        systems.follows = "ihp/systems";
    };

    outputs = inputs@{ ihp, flake-parts, systems, nixpkgs, self, ... }:
        flake-parts.lib.mkFlake { inherit inputs; } {

            systems = import systems;
            imports = [ ihp.flakeModules.default ];

            perSystem = { pkgs, ... }: {
                ihp = {
                    enable = true;
                    projectPath = ./.;
                    packages = with pkgs; [
                        # Native dependencies, e.g. imagemagick
                    ];
                    haskellPackages = p: with p; [
                        # ....
                    ];
                };
            };

            flake.nixosConfigurations."myapp.example.com" = nixpkgs.lib.nixosSystem {
                system = "x86_64-linux";
                specialArgs = inputs;
                modules = [
                    "${nixpkgs}/nixos/modules/virtualisation/amazon-image.nix"
                    ihp.nixosModules.appWithPostgres
                    ({ ... }: {
                        security.acme.defaults.email = "me@example.com";

                        services.ihp = {
                            domain = "myapp.example.com";
                            migrations = ./Application/Migration;
                            schema = ./Application/Schema.sql; # TODO: Remove
                            fixtures = ./Application/Fixtures.sql; # TODO: Remove
                            sessionSecret = "xxx";
                        };

                        # Define an IHP service as a systemd service
                        systemd.services.scheduleCampaigns = {
                            serviceConfig = {
                                Type = "oneshot";
                                WorkingDirectory = "${ihpApp}/lib";
                                ExecStart = "${ihpApp}/bin/ScheduleCampaigns";
                            };
                        };

                        # Run the systemd service daily
                        systemd.timers.scheduleCampaignsDaily = {
                            wantedBy = [ "timers.target" ];
                            partOf = [ "scheduleCampaigns.service" ];
                            timerConfig = {
                                OnCalendar = "*-*-* 01:00:00";
                                Unit = "scheduleCampaigns.service";
                            };
                        };

                    })
                ];
            };
        };
}
```

Fixes https://github.com/digitallyinduced/ihp/issues/1696


kodeFant:
I haven't tried it in practice yet, but looks very good :+1:

I think it could be a good idea to have the option, or even have it mandatory, to specify the major PostgreSQL version.

If not specifically set, there is a risk of accidentally NixOS bumping the `pkgs.postgresql` to a new major version during a future NixOS upgrade without performing the [necessary upgrading procedures](https://nixos.org/manual/nixos/stable/#module-postgresql)

kodeFant:
At some point, I can see foresee a near future where this eventually deprecates Shipnix, which I strongly support!

Due to so many changes and innovations in Nix and the IHP tooling the last year, it already feels like I am working on a legacy project, and the efforts in working on it is a huge net minimum for me.

My main motivation was that I wanted to deploy IHP on NixOS servers with custom machine specifications and with full flexibility, and now we are getting there at the framework level. 

I think it will mostly boil down to two things: Provisioning and environment variable management.

For provisioning, `nixos-anywhere` looks like a powerful option: 
https://github.com/numtide/nixos-anywhere

I can imagine production environment variables could be handled from the IHP IDE with SSH under the hood

kodeFant:
At some point, I can see foresee a near future where this eventually deprecates Shipnix, which I strongly support!

Due to so many changes and innovations in Nix and the IHP tooling the last year, it already feels like I am working on a legacy project, and the efforts in working on it is a continuous net loss for me.

My main motivation was that I wanted to deploy IHP on NixOS servers with custom machine specifications and with full flexibility, and now we are getting there at the framework level. 

I think it will mostly boil down to two things: Provisioning and environment variable management.

For provisioning, `nixos-anywhere` looks like a powerful option: 
https://github.com/numtide/nixos-anywhere

I can imagine production environment variables could be handled from the IHP IDE with SSH under the hood

mpscholten:
> If not specifically set, there is a risk of accidentally NixOS bumping the pkgs.postgresql to a new major version during a future NixOS upgrade without performing the [necessary upgrading procedures](https://nixos.org/manual/nixos/stable/#module-postgresql)

This is automatically handled by NixOS as long as we set the stateVersion (which is done in [NixSupport/nixosModules/app.nix](https://github.com/digitallyinduced/ihp/pull/1792/files#diff-c48c9df63aa26359c56335a0608fa1ef1ebbc4cffc41342ad57615ff7f2330a8)): `system.stateVersion = "23.05"; `

NixOS then selects the right postgres package based on the version here https://github.com/NixOS/nixpkgs/blob/nixos-23.05/nixos/modules/services/databases/postgresql.nix#L454

kodeFant:
Ah, great! I wasn't aware of this!

NixSupport/nixosModules/app.nix
```diff
@@ -0,0 +1,25 @@
...
+        ihp.nixosModules.services_worker
+        ihp.nixosModules.services_migrate
+    ];
+
+    # Speed up builds with the IHP binary cache
+    nix.settings.substituters = [ "https://digitallyinduced.cachix.org" ];
+    nix.settings.trusted-public-keys = [ "digitallyinduced.cachix.org-1:y+wQvrnxQ+PdEsCt91rmvv39qRCYzEgGQaldK26hCKE=" ];
+    
+    # Add swap to avoid running out of memory during builds
+    swapDevices = [ { device = "/swapfile"; size = 8192; } ];
```
kodeFant, line 18: Could it be a custom option in the `flake.nix` for swapfile size? 8 gigs could be a lot for a tiny server.

NixSupport/nixosModules/app.nix
```diff
@@ -0,0 +1,25 @@
...
+
+    # Pin the nixpkgs to the IHP nixpkgs
+    nix.registry.nixpkgs.flake = nixpkgs;
+
+    system.stateVersion = "23.05";
```
kodeFant, line 23: I think this should be set as a mandatory option from the `flake.nix` then to avoid the stateVersion being changed in the future

NixSupport/nixosModules/services/worker.nix
```diff
@@ -0,0 +1,29 @@
...
+                defaultEnv = {
+                    PORT = "${toString cfg.appPort}";
+                    IHP_ENV = cfg.ihpEnv;
+                    IHP_BASEURL = cfg.baseUrl;
+                    IHP_REQUEST_LOGGER_IP_ADDR_SOURCE = cfg.requestLoggerIPAddrSource;
+                    DATABASE_RUL = cfg.databaseUrl;
+                    IHP_SESSION_SECRET = cfg.sessionSecret;
+                };
+            in
+                defaultEnv // cfg.additionalEnvVars;
```
kodeFant, line 27: Could the systemd services also support environment files so secrets are not stored in version control or in the nix store?

flake-module.nix
```diff
@@ -192,6 +188,13 @@ ihpFlake:
...
                 env.IHP_LIB = "${ihp}/lib/IHP";
                 env.IHP = "${ihp}/lib/IHP"; # Used in the Makefile
+
+                scripts.deploy-to-nixos.exec = ''
+                    echo "usage: deploy-to-nixos <flake> <target-host>"
+                    echo "example: deploy-to-nixos stagging.example.com"
+                    nix shell nixpkgs#nixos-rebuild --command bash -c "nixos-rebuild switch -j auto --use-substitutes --fast --flake .#$1 --target-host $1 --build-host $1 --option substituters https://digitallyinduced.cachix.org --option trusted-public-keys digitallyinduced.cachix.org:digitallyinduced.cachix.org-1:y+wQvrnxQ+PdEsCt91rmvv39qRCYzEgGQaldK26hCKE="
+                    # ssh ihp-deploy-test systemctl start migrate
+                '';
             };
```
kodeFant, line 198: What if for example I need to run some custom custom build steps, for example a custom make command, or npm that some people use? Would it work with what's being done locally on the machine?

kodeFant (commented):
Added some comments.

Also would recommend copying the default auto garbage collection options that are default in Shipnix to avoid disk space filling up. It's was a recurring problem on Shipnix that people ran out of disk space and the database service crashed without this

```ssh
  nix.gc = {
    automatic = true;
    dates = "weekly";
    options = "--delete-older-than 30d";
  };

  # Saves disk space by detecting and handling identical contents in the Nix Store
  nix.settings.auto-optimise-store = true;
```

NixSupport/nixosModules/app.nix
mpscholten: Good point 👍 should be moved to flake.nix

NixSupport/nixosModules/app.nix
mpscholten: Makes sense 👍

NixSupport/nixosModules/services/worker.nix
mpscholten: 👍

flake-module.nix
mpscholten: Yes that should work by adding custom build steps to the `Makefile`. E.g. by making your custom build task a dependency of `build/bin/RunUnoptimizedProdServer`. The it will be triggered automatically during the `nix build` call

Issue #1820, opened 2023-09-14
Inconsistent behaviour for selectField when using enum vs references


mariano-dagostino:
:wave: I'm pretty new to IHP, forgive me if I'm not correct enough describing this problem.

I've seen that using `selectField` in combination with enums leads to inconsistent rendering of `<select>` inputs.

Below there is an example of 2 fields, one populated with enums. and the other one with references from a table. This code was generated using some boilerplate code generated by the IHP IDE. And following the [documentation](https://ihp.digitallyinduced.com/Guide/form.html#select-inputs-with-custom-enums).

https://github.com/digitallyinduced/ihp/assets/5313631/36efb6f8-c904-4d5b-9570-2234714eb8ce

As you can observe. When using enums. The option `Please select` is disabled. But when populating the same field with the ids of an external table, it works.

This is the generated HTML for the form:

![Screenshot_2023-09-13_23-23-06](https://github.com/digitallyinduced/ihp/assets/5313631/7eb00bd1-67dd-4248-9347-fd7746b098d9)

Both columns for the Music Album table are set to `NOT Nullable` so attempting to save the form as is will trigger a SQL error because `00000000-0000-0000-0000-000000000000` is not a valid value for the `genre_id`. The enums field does not allow selecting that option, which seems more robust, but at the same time, does not allow you to let users choose, it always select the first option. Which can lead other problems from a UX point of view.

![Screenshot_2023-09-13_23-33-59](https://github.com/digitallyinduced/ihp/assets/5313631/7d2a6753-a768-4e5f-8170-e9ed2f6d1da6)

I don't have the answer to how to fix this. In my opinion the correct behavior should be the presented by the second dropdown. With some extra validation.

mpscholten:
Thanks for reporting this 👍 

The select field is always displaying the value it gets from the music album passed to the `formFor`. The issue is introduced basically by the `newRecord @MusicAlbum`. The `newRecord @MusicAlbum` call in the controller creates an empty record setting all fields to a default empty value. The default empty value for UUIDs is the `00000000-0000-0000-0000-000000000000` and the default empty value for enums is the first enum value.

> The enums field does not allow selecting that option, which seems more robust, but at the same time, does not allow you to let users choose,

👍  this would atleast make it consistent

> Both columns for the Music Album table are set to NOT Nullable so attempting to save the form as is will trigger a SQL error because 00000000-0000-0000-0000-000000000000 is not a valid value for the genre_i

[...]

PR #1821, opened 2023-09-14
Allow empty select field for enums to force user selection

amitaibu:
fixes https://github.com/digitallyinduced/ihp/issues/1820


mpscholten:
This will accidentally reset the form field whenever the validation fails. If we'd use [`record |> didChange #field`](https://ihp.digitallyinduced.com/api-docs/IHP-ModelSupport.html#v:didChange) we could avoid this

amitaibu:
@mpscholten how would we write it for a generic `fieldName`?

`isNew model && not (didChange @fieldName model)` won't compile

mpscholten:
We have the `field` as a `Proxy` value in scope, so try this:

```haskell
isNew model && not (didChange field model)
```

amitaibu:
I think GH actions are still not fixed, so I ran tests locally

image

IHP/View/Form.hs
```diff
@@ -757,6 +759,17 @@ selectField field items = FormField
...
         FormContext { model } = ?formContext
+        -- The select field is always displaying the value it gets from the nodel passed to the formFor.
+        -- The issue is introduced basically by the `newRecord @Record`. The `newRecord` call in the controller creates an empty record setting all fields to a default empty value.
+        -- The default empty value for UUIDs is the 00000000-0000-0000-0000-000000000000 and the default empty value for
+        -- enums is the first enum value.
+        -- Now, if we have a required field, we want to make sure the user selects a value, in the same
+        -- way they have to select for a reference field.
+        -- So we check if the model is new and the field was not submitted yet, then we set the
+        -- field value to an empty string. Otherwise, we use the value from the model.
+        fieldValue = if isNew model && null (paramList @Text (cs fieldName))
```
amitaibu, line 771: fyi, `didChange` couldn't be used here, since the model was still holding the first enum. Instead, we're using `paramList` to check if the field was submitted.

amitaibu:
Re-setting to draft, as need to tackle the `radioField` as well.

amitaibu:
Adapted and simplified `radioField`

image

Tests still passing.

image

amitaibu:
Ready for review.

IHP/View/Form.hs
mpscholten: I've tried this and it worked locally:

```haskell
        fieldValue = if didChange field model
                    then inputValue (getField @fieldName model :: SelectValue item)
                    else ""
```

amitaibu:
> I've tried this and it worked locally:

Try selecting the 1st enum. In that case, from the model perspective, nothing has changed.

amitaibu:
I realized the PR is wrong, as with it we can't set a default value for an enum.

I'll convert this to a recipe instead, having a `fieldValueForceSelect` helper function.

mpscholten:
> I realized the PR is wrong, as with it we can't set a default value for an enum.

Good point. Didn't realize this

IHP/View/Form.hs
```diff
@@ -722,6 +723,7 @@ checkboxField field = FormField
 -- >     render NewView { .. }
 selectField :: forall fieldName model item.
     ( ?formContext :: FormContext model
+    , ?context::ControllerContext
```
amitaibu, line 726: remove

IHP/View/Form.hs
```diff
@@ -802,43 +834,47 @@ selectField field items = FormField
...
+    , placeholder = ""
     }
-    where
-        fieldName = symbolVal field
-        FormContext { model } = ?formContext
 {-# INLINE radioField #-}
 
+{- Radio field where @fieldValue@ is empty text when the model is new.
+
+See `selectFieldEemptyFieldValueWhenIsNew` for more details.
```
amitaibu, line 858: Add link

IHP/View/Form.hs
```diff
@@ -759,6 +761,36 @@ selectField field items = FormField
...
+enums is the first enum value.
+Now, if we have a required field, we want to make sure the user selects a value, in the same
+way they have to select for a reference field.
+So we check if the model is new and the field was not submitted yet, then we set the
+field value to an empty string. Otherwise, we use the value from the model.
+If the select field is empty, then the param would be empty as well.
+-}
+selectFieldEemptyFieldValueWhenIsNew :: forall fieldName model item.
+    ( ?formContext :: FormContext model
+    , ?context::ControllerContext
```
amitaibu, line 779:
```suggestion
```

IHP/View/Form.hs
```diff
@@ -19,12 +19,13 @@ import           IHP.ViewSupport
 import qualified Text.Blaze.Html5                   as Html5
 import IHP.HSX.ToHtml
 import GHC.Types
-import IHP.ModelSupport (getModelName, inputValue, isNew, Id', InputValue)
+import IHP.ModelSupport (didChange, getModelName, inputValue, isNew, Id', InputValue)
```
amitaibu, line 22:
```suggestion
import IHP.ModelSupport (getModelName, inputValue, isNew, Id', InputValue)
```

IHP/View/Form.hs
```diff
@@ -759,6 +761,39 @@ selectField field items = FormField
...
+model passed to @formFor@.
+
+For a new record, that value will always be the first enum. And the reason lies in the newRecord @Record call in the controller.
+The @newRecord@ creates an empty record setting all fields to a default empty value.
+The default empty value for UUIDs is 00000000-0000-0000-0000-000000000000 and the default empty value for enums is the first enum value.
+
+Now, if we have a required field, and we want to make sure the user selects a value, in the same way they have to select for a reference field.
+So we check if the model is new and the field was not submitted yet, then we set the field value to an empty string. Otherwise, we use the value from the model.
+-}
+selectFieldEmptyFieldValueWhenIsNew :: forall fieldName model item.
```
amitaibu, line 777: @mpscholten do you have an idea for a better name?

amitaibu:
@mpscholten ready for review

mpscholten:
I don't like that this change adds two new functions. IMO this needs to be simpler.

Why didn't `didChange` work for you?

```haskell
        fieldValue = if didChange field model
                    then inputValue (getField @fieldName model :: SelectValue item)
                    else ""
```

amitaibu:
> I don't like that this change adds two new functions

Same, but I didn't find a better solution so far.

> Why didn't didChange work for you?

We can't select the first enum: The record is holding the first enum by default, and only the HTML field that has `fieldValue = ""` -- then we user will select the first enum, then `didChange` will return False.

amitaibu:
Another option would be to extend the existing `FieldForm` to have a `resetFieldValue :: Bool`, and have the logic inside the existing select/radio fields.

mpscholten:
>  then didChange will return False

Maybe we can check `record.meta.touchedFields` directly. Whenever `set` or `fill` is called, the field name will be added to the Set in `record.meta.touchedField`. The `touchedFields` Set is used for `updateRecord`, so that `updateRecord` only updates fields that have actually been written to.

When you look at the implementation of `didChange`, it actually checks 1) the field has been written to and 2) the field value is different from the initial value. We basically want to skip step 2 here:

```haskell
didChange :: forall fieldName fieldValue record. (KnownSymbol fieldName, HasField fieldName record fieldValue, HasField "meta" record MetaBag, Eq fieldValue, Typeable record) => Proxy fieldName -> record -> Bool
didChange field record = didTouchField && didChangeField
    where
        didTouchField :: Bool
        didTouchField =
            record.meta.touchedFields
            |> includes (cs $! symbolVal field)

        didChangeField :: Bool
        didChangeField = originalFieldValue /= fieldValue

        fieldValue :: fieldValue
        fieldValue = record |> getField @fieldName

        originalFieldValue :: fieldValue
        originalFieldValue =
            record.meta.originalDatabaseRecord
            |> fromMaybe (error "didChange called on a record without originalDatabaseRecord")
            |> fromDynamic @record
            |> fromMaybe (error "didChange failed to retrieve originalDatabaseRecord")
            |> getField @fieldName
```

So we want something like:

```haskell
didTouch :: forall fieldName fieldValue record. (KnownSymbol fieldName, HasField fieldName record fieldValue, HasField "meta" record MetaBag, Eq fieldValue, Typeable record) => Proxy fieldName -> record -> Bool
didTouch field record = didTouchField
    where
        didTouchField :: Bool
        didTouchField =
            record.meta.touchedFields
            |> includes (cs $! symbolVal field)
```

With the `didTouch` it should work (or maybe I miss something here)

amitaibu:
Cool, I'll look into it.

amitaibu:
I think we still need to add a function like `fieldValueEmptyByDefault`

Here's the current code:
https://github.com/digitallyinduced/ihp/blob/35d5d0cc1006bfc1d32b173a187b9c1be06ff0f0/IHP/View/Form.hs#L741

So ideally, one would be able to do something like:

```
{(selectField #enumTest) {fieldValue = fieldValueEmptyByDefault}}
```

1. @mpscholten sounds right?
2. If so, not sure how to allow `{fieldValue = fieldValueEmptyByDefault}` to work (i.e. passing a partial function)

amitaibu:
Another option is to add

```diff
data FormField = FormField
    { fieldType :: !InputType
    , fieldName :: !Text
    , fieldLabel :: !Text
    , fieldValue :: !Text
+   , fieldValueEmptyByDefault :: !Bool
```

Then inside `selectField` we would know if we'd need to set `fieldValue = ""` or `fieldValue = inputValue ...`

amitaibu:
A nicer option, but would break BC would be something like: 

`fieldValue :: !FieldValueSetting`

Where `FieldValueSetting = FieldValueEmptyByDefault | FieldValueFixedValue`

mpscholten:
I think this is still too complicated.

Here's a simpler solution based on `didTouch`: https://github.com/digitallyinduced/ihp/commit/72fc0984b35874fbf580ed95303a342381489e23

Here's a project implementing this:

https://github.com/mpscholten/issue-1820/blob/main/Web/Controller/Posts.hs#L14

```haskell
    action NewPostAction = do
        let post = newRecord
        let postWithDefault = newRecord |> set #postType Article
        render NewView { .. }
```

```haskell
module Web.View.Posts.New where
import Web.View.Prelude

data NewView = NewView
    { post :: Post
    , postWithDefault :: Post
    }

instance View NewView where
    html NewView { .. } = [hsx|
        {breadcrumb}
        <h1>New Post, without default</h1>
        {renderForm post}

        <h1>New Post, with default</h1>
        {renderForm postWithDefault}
    |]
        where
            breadcrumb = renderBreadcrumb
                [ breadcrumbLink "Posts" PostsAction
                , breadcrumbText "New Post"
                ]

renderForm :: Post -> Html
renderForm post = formFor post [hsx|
    {(textField #title)}
    {(selectField #postType values)}
    {submitButton}
|]
    where
        values = allEnumValues @PostType


instance CanSelect PostType where
    type SelectValue PostType = PostType
    selectValue value = value
    selectLabel = show
```

Rendering that form I get this:

image

amitaibu:
Ok, that's much nicer. I like that now one has to set the field explicitly.

IHP/ModelSupport.hs
```diff
@@ -802,6 +797,27 @@ didChange field record = didTouchField && didChangeField
...
+--
+-- >>> let projectId = "227fbba3-0578-4eb8-807d-b9b692c3644f" :: Id Project
+-- >>> project <- fetch projectId
+-- >>> project |> set #name project.name |> didChange #name
+-- True
+--
+didTouchField :: forall fieldName fieldValue record. (KnownSymbol fieldName, HasField fieldName record fieldValue, HasField "meta" record MetaBag, Eq fieldValue, Typeable record) => Proxy fieldName -> record -> Bool
+didTouchField field record =
+    record.meta.touchedFields
+    |> includes (cs $! symbolVal field)
```
amitaibu, line 819: Why is `$!` required here?

Guide/form.markdown
```diff
@@ -666,6 +666,19 @@ formFor subscription [hsx|
         allContentTypes = allEnumValues @ContentType
 ```
 
+### Set Default Value for Custom Enums
+
+When creating a new record, by default the field value will be empty. If you'd like to set a default enum, you can set it from the controller.
+
+Note that by default the `newRecord` populates the first enum on the record. However, when showing the form, IHP will check if the field was not explicetly set, and if so, will not render the default value.
```
amitaibu, line 673:
```suggestion
Note that by default the `newRecord` populates the first enum on the record. However, when showing the form, IHP will check if the field was not explicitly set, and if so, will not render the default value.
```

IHP/ModelSupport.hs
mpscholten: it's a strict version of `$`, I used it here because I felt like it could improve performance here.

I think we can actually use [`IHP.HaskellSupport.symbolToText`](https://github.com/digitallyinduced/ihp/blob/master/IHP/HaskellSupport.hs#L374) here:

```haskell
|> includes (symbolToText @fieldName)
```

This is even a bit shorter

IHP/ModelSupport.hs
```diff
@@ -802,6 +797,27 @@ didChange field record = didTouchField && didChangeField
...
+-- >>> let projectId = "227fbba3-0578-4eb8-807d-b9b692c3644f" :: Id Project
+-- >>> project <- fetch projectId
+-- >>> didTouchField #name project
+-- False
+--
+-- __Example:__ Returns 'True' after setting a field
+--
+-- >>> let projectId = "227fbba3-0578-4eb8-807d-b9b692c3644f" :: Id Project
+-- >>> project <- fetch projectId
+-- >>> project |> set #name project.name |> didChange #name
```
mpscholten, line 813:
```suggestion
-- >>> project |> set #name project.name |> didTouchField #name
```

mpscholten: Missed this one when copy pasting

IHP/ModelSupport.hs
amitaibu: Sure, let me re-roll the PR.

mpscholten (approved):
Thanks 👍

PR #1823, opened 2023-09-19
Add back withHoogle support

nvmd:
- Adds `withHoogle` support back: it was missing since the migration to flakes/flake-parts.
- Updates documentation to reflect the changes in configuration and usage since the migration to flake-parts.


mpscholten (approved):
Thanks 👍

mpscholten:
No, this will only enable it for the dev shell

PR #1834, opened 2023-10-02
Fix imports for signed token example

amitaibu:
I've updated `nix flake update`on https://github.com/amitaibu/ihp-landing-page/ , and had to fix the error:

```
Ambiguous module name ‘Crypto.PubKey.RSA’:
      it was found in multiple packages: crypton-0.32 cryptonite-0.30
```


mpscholten (approved):
👍

Issue #1028, opened 2021-08-07
Blog demo app: passing malformed UUID causes page to not reload


dharmatech:
Hey y'all!

This is regarding the blog demo app from the guide.

In the file `Web/View/Comments/Index.hs`, we're instructed to remove the code that calls `NewCommentAction`.

If I add back in, a call to `NewCommentAction` with a hardcoded UUID like this:

    <a href={NewCommentAction "0f671934-0971-402e-ab70-f17e91b56a82"}>abc</a>

that works fine and the page renders properly with the new link.

However, if I remove a character from the UUID (the first one in this case) making it a malformed UUID:

    <a href={NewCommentAction "f671934-0971-402e-ab70-f17e91b56a82"}>abc</a>

the page does not reload. If I manually reload the page, the page doesn't load at all.

In my experience so far with IHP, I'd normally expect an error message if there was a problem running the application. So this behaviour seems unusual.

Here's a video demonstrating the issue:

https://youtu.be/9TlUOGasXWU

Just wondering, is this the expected behaviour? Or would it be possible to provide an error message to the user in a case like this?

Is there somewhere else in the logs that makes the page reload issue apparent?

Thanks!

mpscholten:
Hey, thanks for the great error report! :) This is a bug. I'll take a look in a few minutes.

mpscholten:
This is now fixed via https://github.com/digitallyinduced/ihp/commit/e6b1b642746005bb035cf533b42faaa1418a1e2b :) The issue was that the error was thrown too late in the request handling lifecycle because of lazyness. We've disabled a bit of the lazyness in dev mode to be able to handle the error with the normal IHP error tooling.

dharmatech:
@mpscholten,

Well that was fast! :-) Thanks! And thanks for the explanation.

PR #1837, opened 2023-10-05
Add renderHtmlPartial function

kodeFant:
When using `respondHtml`, rendering partial hsx functions will fail if trying to use `fromFrozenContext` and related utilities  like `currentUser`.

This makes patching a view with html partials a bit less inconvenient, for example for htmx users.

With `renderHtmlPartial` this will work.

### DIFF
--- a/Guide/htmx-and-hyperscript.markdown
+++ b/Guide/htmx-and-hyperscript.markdown
@@ -63,7 +63,7 @@ data CounterController
     deriving (Eq, Show, Data)
 ```
 
-Instead of using the `render` function, htmx routes are better used with `respondHtml` to avoid the layout being shipped as part of the response. The same function can be used for initializing the view as well as upating.
+Instead of using the `render` function, htmx routes are better used with `renderHtmlPartial` to avoid the layout being shipped as part of the response. The same function can be used for initializing the view as well as upating.
 
 ```haskell
 module Web.Controller.Counter where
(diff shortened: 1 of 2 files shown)


kodeFant:
I noticed an issue with this on a production server, but not in development mode.

I tried to manually run a partial endpoint with a missing required parameter. 

This resulted in `502 Bad Gateway` and looks like the server panics globally for 2 seconds and  also errors globally on the site, then it recovers works fine again. 

So that seems to be a little risk factor if an attacker would feel like bringing a site down.

I would guess it's the same issue as this one: https://github.com/digitallyinduced/ihp/issues/1028

kodeFant:
Here are some error logs from the server:

```sh
Oct 10 13:50:42 kompak-prototype RunProdServer[215109]: ParamNotFoundException {name = "framework"}
Oct 10 13:50:42 kompak-prototype nginx[215099]: 2023/10/10 13:50:42 [error] 215099#215099: *22448 upstream prematurely closed connection while reading response header from upstream, client: 84.234.224.78, server: kompak.ai, request: "GET /GetFrameworkHelpTextPartial HTTP/2.0", upstream: "http://127.0.0.1:8000/GetFrameworkHelpTextPartial", host: "kompak.ai"
Oct 10 13:50:42 kompak-prototype nginx[215099]: 2023/10/10 13:50:42 [error] 215099#215099: *22448 connect() failed (111: Connection refused) while connecting to upstream, client: 84.234.224.78, server: kompak.ai, request: "GET /GetFrameworkHelpTextPartial HTTP/2.0", upstream: "http://[::1]:8000/GetFrameworkHelpTextPartial", host: "kompak.ai"
```

PR #1840, opened 2023-10-08
Bringing Test Suite sqlImport in line with IDE. 

Montmorency:
The old importSql function in createTestDatabase for Mocking used a raw byte string and then imported Schema as a PGConnect/PGQuery. This ran into parsing issues when init a complex schema into test database. The IDE had no problem importing same Schema with psql. This PR updates testDatabase to work with psql import.

### DIFF
--- a/IHP/Test/Database.hs
+++ b/IHP/Test/Database.hs
@@ -10,6 +10,9 @@ import qualified Data.ByteString as ByteString
 import qualified IHP.LibDir as LibDir
 import qualified Control.Exception as Exception
 
+import qualified System.Process as Process
+import qualified System.Directory as Directory
+
 data TestDatabase = TestDatabase
     { name :: Text
     , url :: ByteString
@@ -22,6 +25,7 @@ data TestDatabase = TestDatabase
 --
 createTestDatabase :: ByteString -> IO TestDatabase
 createTestDatabase databaseUrl = do
+    currentDir <- Directory.getCurrentDirectory
     databaseId <- UUID.nextRandom
     let databaseName = "test-" <> UUID.toText databaseId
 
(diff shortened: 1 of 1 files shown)


IHP/Test/Database.hs
```diff
@@ -34,8 +38,11 @@ createTestDatabase databaseUrl = do
             |> cs
 
     libDir <- LibDir.findLibDirectory
-    importSql newUrl (cs libDir <> "/IHPSchema.sql")
-    importSql newUrl "Application/Schema.sql"
+
+    let importSql file = Process.callCommand ("psql -h '" <> currentDir <> "/build/db' -d " <> (cs databaseName) <> " < " <> file)
```
mpscholten, line 42: I think this should be using `databaseUrl` instead of the current dir, otherwise the function might fail if the database url passed to this function is different than the current directory

```suggestion
    let importSql file = Process.callCommand ("psql -h '" <> databaseUrl <> "/build/db' -d " <> (cs databaseName) <> " < " <> file)
```

Montmorency: Agree using `newUrl` which will have the temporary database name patched over the connection from databaseurl

IHP/Test/Database.hs
```diff
@@ -34,8 +38,11 @@ createTestDatabase databaseUrl = do
             |> cs
 
     libDir <- LibDir.findLibDirectory
-    importSql newUrl (cs libDir <> "/IHPSchema.sql")
-    importSql newUrl "Application/Schema.sql"
+
+    let importSql file = Process.callCommand ("psql " <> (cs newUrl) <> " < " <> file)
```
mpscholten, line 42: Maybe we should also add a comment why `psql` is used here isntead of the old approach. Otherwise we might change that back in the future because someone doesn't know about the encoding probelms :D

PR #1848, opened 2023-10-20
Add `Ord` to Enums

amitaibu:
I'd like to have an associate list with an enum as key.

Assuming Foo is an Enum:

```haskell
import qualified Data.Map as Map

let value = Map.lookup Foo treeJson
```

Without this PR we get error:

```
Could not deduce (Ord ProductItemType)
        arising from a use of ‘Map.lookup’
```

### DIFF
--- a/Test/SchemaCompilerSpec.hs
+++ b/Test/SchemaCompilerSpec.hs
@@ -178,7 +178,7 @@ tests = do
 
                     type instance PrimaryKey "users" = UUID
 
-                    type User = User'
+                    type User = User' 
 
                     type instance GetTableName (User' ) = "users"
                     type instance GetModelByTableName "users" = User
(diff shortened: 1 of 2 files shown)


Test/SchemaCompilerSpec.hs
```diff
@@ -178,7 +178,7 @@ tests = do
 
                     type instance PrimaryKey "users" = UUID
 
-                    type User = User' 
+                    type User = User'
```
amitaibu, line 181: My editor is trimming white space, but the white space is needed for the tests. I'll fix it manually for now.

```suggestion
                    type User = User' 
```

Test/SchemaCompilerSpec.hs
```diff
@@ -247,7 +247,7 @@ tests = do
...
```
amitaibu, line 250:
```suggestion
                    type User = User' 
```

Test/SchemaCompilerSpec.hs
```diff
@@ -316,7 +316,7 @@ tests = do
...
```
amitaibu, line 319:
```suggestion
                    type User = User' 
```

mpscholten (approved):

PR #1861, opened 2023-11-17
Update deployment.markdown

AronNovak:
It is not trivial to follow the deployment instructions


mpscholten (approved):
Thanks, good improvements 👍 I added some specific code example additionally, as that might be even more clear

PR #1863, opened 2023-11-17
Docs on how to use ToField to query custom tuples in IN clause

amitaibu:


Guide/database.markdown
```diff
@@ -334,6 +334,72 @@ do
...
+    , Plain (char8 ')')
+    ]
+```
+
+Then, in our controller, we can execute the query:
+
+```haskell
+-- Web/Controller/Projects.hs
+action ProjectsAction = do
+    -- let pairs = projects |> fmap (\project -> (project.projectType, project.participants))
```
amitaibu, line 396:
```suggestion
```

Guide/database.markdown
```diff
@@ -334,6 +334,71 @@ do
...
+    ]
+```
+
+Then, in our controller, we can execute the query:
+
+```haskell
+-- Web/Controller/Projects.hs
+action ProjectsAction = do
+    let pairs = [(ProjectTypeOngoing, 1 :: Int), (ProjectTypeFinished, 2)]
+    projects :: [Project] <- sqlQuery "SELECT * FROM projects WHERE (project_type, participants) IN ?" (Only $ In pairs)
```
amitaibu, line 397: Is using `*` correct?

amitaibu: RE-checked, yes, it's working fine

Guide/database.markdown
```diff
@@ -334,6 +334,75 @@ do
                 |> map (\(Only uuid) -> Id uuid :: Id Project)
 ```
 
+### Implementing `ToField` For Custom Types.
+
+This is a more advanced topic, and before trying to implement it, we encourage you to
+try and use the [Query Builder](https://ihp.digitallyinduced.com/Guide/querybuilder.html), as it provides a better type safety. Even the below example can be accomplished with the Query Builder.
```
amitaibu, line 340: > Even the below example can be accomplished with the Query Builder.

Once we have https://github.com/digitallyinduced/ihp/issues/1867 , we should link it here.

amitaibu:
Made this a guide on the `ToField` implementation.

amitaibu:
@mpscholten ready for review

mpscholten:
Shouldn't we wait until https://github.com/digitallyinduced/ihp/issues/1867 is merged? That seems like a lot simpler solution to me

PR #1870, opened 2023-11-28
Remove shipnix from docs

kodeFant:
Although Shipnix still exists for now, I think it's time to remove it from the IHP docs. The service is likely going to be phased out after a while. 

The stuff in the beginner guide is a bit outdated and might not be in IHP's best interest to have it there.

Thank you so much for the positivity, support and encouragement for this project!


mpscholten (approved):
😢 sad to see shipnix go. Hope the deploy-to-nixos command can be a good replacement

PR #1875, opened 2023-12-23
Allow a submit button to be disabled

amitaibu:
Sometimes we may wish to have the submit button disabled. In my case, I later enable it via JS, if the form has changed.

```haskell
renderForm :: LandingPage -> Html
renderForm landingPage = formFor landingPage [hsx|
    {(textField #title)}

    {submitButton {label = "Save Landing page", buttonDisabled = True}}
|]
```

image

### DIFF
--- a/IHP/View/CSSFramework.hs
+++ b/IHP/View/CSSFramework.hs
@@ -272,9 +272,9 @@ instance Default CSSFramework where
 
             styledValidationResultClass = ""
 
-            styledSubmitButton cssFramework SubmitButton { label, buttonClass } =
+            styledSubmitButton cssFramework SubmitButton { label, buttonClass, disabled } =
                 let className :: Text = cssFramework.styledSubmitButtonClass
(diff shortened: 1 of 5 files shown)


mpscholten:
Can we name the field `disabled` instead of `buttonDisabled`? This would be [more consistent with the other field helpers](https://ihp.digitallyinduced.com/Guide/form.html#disabled-fields) (maybe you tried that already and it led to some compiler issue because of the duplicate field names)

amitaibu:
> maybe you tried that already and it led to some compiler issue because of the duplicate field names

Indeed, I've started with `disabled`, but it complained about ambiguous property [here](https://github.com/digitallyinduced/ihp/blob/ae76491c1ac655e4ad18e0502f5b7a11443fc73b/Test/View/CSSFrameworkSpec.hs#L77) . If there's a solution for that, I'd also prefer to use `disabled`

mpscholten:
This error has no solution :D so then we're stuck with `buttonDisabled`

mpscholten (approved):

PR #1877, opened 2023-12-28
`fileOrNothing` should check `fileContent` isn't empty

amitaibu:
Right now we return a `Just file`, even when there's no file.

This is similar to the check we already have here https://github.com/digitallyinduced/ihp/blob/d0658832f6d51160678f3a03d0920f3621b844e2/IHP/FileStorage/ControllerFunctions.hs#L333

### DIFF
--- a/IHP/Controller/FileUpload.hs
+++ b/IHP/Controller/FileUpload.hs
@@ -57,7 +57,7 @@ fileOrNothing !name =
             FormBody { files } ->
                 -- Search for the file, and confirm it's not an empty one.
                 case lookup name files of
-                    Just file | fileContent file /= "" -> Just file
+                    Just fileInfo | not (LBS.null (fileInfo.fileContent)) -> Just fileInfo
                     _ -> Nothing
             _ -> Nothing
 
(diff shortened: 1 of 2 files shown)


IHP/Controller/FileUpload.hs
```diff
@@ -49,12 +49,16 @@ import qualified System.Process as Process
...
+-- See 'IHP.FileStorage.ControllerFunctions.storeFile' to upload the file to S3 or similar cloud storages.
 --
 fileOrNothing :: (?context :: ControllerContext) => ByteString -> Maybe (FileInfo LBS.ByteString)
 fileOrNothing !name =
         case ?context.requestContext.requestBody of
-            FormBody { files } -> lookup name files
+            FormBody { files } ->
+                -- Search for the file, and confirm it's not an empty one.
+                case lookup name files of
+                    Just file | fileContent file /= "" -> Just file
```
mpscholten, line 60: Performance wise it's better to use a `not (null fileContent)` here instead, like in ` Just fileInfo | not (LBS.null (fileInfo.fileContent))`

```suggestion
                    Just file | fileContent file /= "" -> Just file
```

mpscholten (commented):
When the `fileOrNothing` function now already checks for emptiness, we can remove the emptiness checks from the call-sites of `fileOrNothing`. E.g. in `uploadToStorageWithOptions`, we can now remove the `| not (LBS.null (fileInfo.fileContent))` condition, as `fileOrNothing` already checked that

mpscholten (approved):

PR #1882, opened 2024-01-01
Add debounce to fsnotify

### LINKED ISSUE #1809: Re-add debounce to fsnotify
The latest version of fsnotify has removed debouncing from the library, so we likely need to implement this ourselves.

Related to https://github.com/digitallyinduced/ihp/pull/1771

mpscholten:
Fixes #1809

### DIFF
--- a/ihp.cabal
+++ b/ihp.cabal
@@ -105,7 +105,7 @@ common shared-properties
             , with-utf8
             , ihp-hsx
             , ihp-postgresql-simple-extra
-            , auto-update
+            , auto-update -- Used for 'Control.Debounce' in 'IHP.IDE.FileWatcher'
     default-extensions:
         OverloadedStrings
         , NoImplicitPrelude
--- a/ihp.cabal
+++ b/ihp.cabal
@@ -105,6 +105,7 @@ common shared-properties
             , with-utf8
             , ihp-hsx
             , ihp-postgresql-simple-extra
+            , auto-update
     default-extensions:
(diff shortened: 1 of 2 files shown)


ihp.cabal
```diff
@@ -105,6 +105,7 @@ common shared-properties
             , with-utf8
             , ihp-hsx
             , ihp-postgresql-simple-extra
+            , auto-update
```
amitaibu, line 108: Is this related to the debounce?

mpscholten: Yes the `Control.Debounce` module is part of that package

amitaibu: Should we add a comment to indicate it?

amitaibu (commented):
Locally I see the "modules loaded" appears now only twice rather than three times, when I switch git branches.

image

Not sure if we can reduce to a single load?

ihp.cabal
mpscholten: Good point, added 👍

mpscholten:
> Not sure if we can reduce to a single load?

Not possible without adding latency. We could use https://hackage.haskell.org/package/auto-update-0.1.6/docs/Control-Debounce.html#v:trailingEdge to first cool down and then trigger the reloads. The issue is that this adds latency to every normal single file change

amitaibu:
Yeah, I was just reading about the `trailingEdge`, and it seems like a no go.

Thanks 🙏

PR #1884, opened 2024-01-02
Fix comment on ihp.cabal

amitaibu:
Follow up https://github.com/digitallyinduced/ihp/pull/1882

@mpscholten Fixes the build error 
image


mpscholten (approved):
Thanks :) Sorry for breaking the build

Issue #1883, opened 2024-01-01
Remove cachix install step from `ihp-new`


mpscholten:
With the new flake based dev env we don't need to install cachix anymore to add the digitallyinduced binary cache. So the step here can be removed https://github.com/digitallyinduced/ihp/blob/master/ProjectGenerator/bin/ihp-new#L219

rvarun11:
Apart from removing cachix install step, do we need: https://github.com/digitallyinduced/ihp/blob/33df9f448cd2bda19f43b7fa60b9a2407b4dff3c/ProjectGenerator/bin/ihp-new#L237

Or can this be removed as well?

mpscholten:
Yes very likely

PR #1886, opened 2024-01-03
Update ihp-new

rvarun11:
Updates `ihp-new` by:

- Fixing direnv install to use `nix profile` instead of `nix-env` 
- Removing cachix install step - https://github.com/digitallyinduced/ihp/issues/1883

### DIFF
--- a/ProjectGenerator/bin/ihp-new
+++ b/ProjectGenerator/bin/ihp-new
@@ -83,7 +83,9 @@ if ! [ -x "$(command -v direnv)" ]; then
 	read -r -n1 choice
 	echo ""
 	case "$choice" in
-	y | Y) nix-env -iA ${NIXPKGS_ATTR}.direnv ;;
+    y | Y)
+        nix profile install nixpkgs#nix-direnv
+        ;;
 	*)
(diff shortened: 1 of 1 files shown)


mpscholten (approved):

ProjectGenerator/bin/ihp-new
```diff
@@ -83,7 +83,9 @@ if ! [ -x "$(command -v direnv)" ]; then
 	read -r -n1 choice
 	echo ""
 	case "$choice" in
-	y | Y) nix-env -iA ${NIXPKGS_ATTR}.direnv ;;
+    y | Y)
+        nix profile install nixpkgs#nix-direnv
```
Eisfunke, line 87: If I'm not mistaken, [`nix profile`](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-profile) needs the experimental features `nix-command` and `flakes`activated, so this would fail on a system without them activated beforehand. We should probably add `--extra-experimental-features` similarly to the other nix commands below in this file.

mpscholten: This has been fixed later on master after this PR

Eisfunke: Ah oh, I see, thanks

Issue #1889, opened 2024-01-09
Can't redirect to remote server


amitaibu:
I was trying to see if I can redirect to a controller, and add a hash. So I tried [this function](https://github.com/digitallyinduced/ihp/blob/028525673b63b775cbbbcce0f478c2a9fd0744fa/IHP/Controller/Redirect.hs#L53C6-L53C58):

`redirectToUrl "https://github.com/digitallyinduced/ihp?query=1#hash=2"`

But this results with an error - likely because we are trying to Ajax it.

![image](https://github.com/digitallyinduced/ihp/assets/125707/b59163b4-5a3e-45f5-ace6-191f3c805e95)

mpscholten:
You can use `formForWithoutJavascript` (https://ihp.digitallyinduced.com/Guide/form.html#disable-form-submission-via-javascript) to workaround this issue (assuming you have a form)

amitaibu:
I think that won't work in all cases. Silly example:

A user has to enter a URL in the field, and when submitted, we will redirect there. Before submit, we still don't know if there's CORS enabled on that URL.

What this PR does is try to fetch it via JS. If it fails, instead of returning an error, it does a full page refresh for the redirect.

PR #1891, opened 2024-01-10
Allow redirect to external sites if CORS isn't enabled

amitaibu:
fixes https://github.com/digitallyinduced/ihp/issues/1889

### DIFF
--- a/lib/IHP/static/helpers.js
+++ b/lib/IHP/static/helpers.js
@@ -296,7 +296,15 @@ window.submitForm = function (form, possibleClickedButton) {
             'Content-Type',
             'application/x-www-form-urlencoded'
         );
-        request.send(parameters.join('&'));
+
+        try {
+            request.send(parameters.join('&'));
+        } catch (e) {
+            // If the send method throws an exception, redirect to the form action URL.
+            // This can happen for example if redirecting to an external site, which doesn't have CORS enabled,
+            // so we can't use Ajax.
+            window.location.href = url;
+        }
     }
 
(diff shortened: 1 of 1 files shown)


lib/IHP/static/helpers.js
```diff
@@ -296,7 +296,15 @@ window.submitForm = function (form, possibleClickedButton) {
...
         );
-        request.send(parameters.join('&'));
+
+        try {
+            request.send(parameters.join('&'));
+        } catch (e) {
+            // If the send method throws an exception, redirect to the form action URL.
+            // This can happen for example if redirecting to an external site, which doesn't have CORS enabled,
+            // so we can't use Ajax.
+            window.location.href = url;
```
mpscholten, line 306: This won't work for forms that are submitted via POST. I don't think this can be really fixed at all except by disabling the JS form submission in those forms

amitaibu: Ok, I'll share my example code soon to show my use case

amitaibu: Here's the example: https://github.com/amitaibu/ihp-landing-page/pull/35

If you navigate to `http://localhost:8000` without the PR it will result with console error. With the PR you'll be redirected to the repo of one of the best Haskell frameworks! :smile:

amitaibu: @mpscholten ^^

amitaibu:
@copilot can you resolve conflicts with `master`?

Issue #1787, opened 2023-08-02
New project appears to be broken right from the start.


seanparsons:
When I create a new project and try to get it up and running the project just appears to be broken:

```
sean@tower ~/workspace (master)> ihp-new ihp-test

Direnv shell hook doesn't seem correct.
Direnv needs to be hooked into your shell. Please follow the steps below:
Bash: Add eval "$(direnv hook bash)" to ~/.bashrc
ZSH: Add eval "$(direnv hook zsh)" to ~/.zshrc
Other shells: See https://direnv.net/#README
This could be a false negative if you use a more complex shell profile

We will now create your new IHP project. This may take up to 30 seconds.
warning: Pathname can't be converted from UTF-8 to current locale.
error: getting status of '/nix/store/h8c7bnrnldxyi94hvmslxid9k87x7yid-source/workspace/ihp-test': No such file or directory


sean@tower ~/workspace (master)> cd ihp-test/
direnv: loading ~/workspace/ihp-test/.envrc
direnv: loading https://raw.githubusercontent.com/nix-community/nix-direnv/2.3.0/direnvrc (sha256-Dmd+j63L84wuzgyjITIfSxSD57Tx7v51DMxVZOsiUD8=)
direnv: using flake . --impure --accept-flake-config
error: getting status of '/nix/store/h8c7bnrnldxyi94hvmslxid9k87x7yid-source/workspace/ihp-test': No such file or directory
error: getting status of '/home/sean/workspace/ihp-test/.direnv/flake-profile.1771549': No such file or directory
error: getting status of '/nix/store/h8c7bnrnldxyi94hvmslxid9k87x7yid-source/workspace/ihp-test': No such file or directory
direnv: nix-direnv: renewed cache
direnv: export ~XDG_DATA_DIRS


sean@tower ~/w/ihp-test (master)> git init .
Initialized empty Git repository in /home/sean/workspace/ihp-test/.git/


sean@tower ~/w/ihp-test (trunk)> direnv allow
direnv: loading ~/workspace/ihp-test/.envrc
direnv: loading https://raw.githubusercontent.com/nix-community/nix-direnv/2.3.0/direnvrc (sha256-Dmd+j63L84wuzgyjITIfSxSD57Tx7v51DMxVZOsiUD8=)
direnv: using flake . --impure --accept-flake-config
warning: Git tree '/home/sean/workspace/ihp-test' is dirty
[...]

PR #1898, opened 2024-01-23
ihp-new: Create git repo to avoid issues when IHP is part of a larger git repo

mpscholten:
Fixes https://github.com/digitallyinduced/ihp/issues/1787


Eisfunke (approved):
LGTM

PR #1904, opened 2024-02-09
Fix naming in test docs

svenk-ct:
I guess the naming got updated, but not in the docs. In my tests the attribute is called `haskellPackages`, not `haskellDeps`.


amitaibu (approved):
:-1: Thanks

PR #1905, opened 2024-02-11
Update tailwindcss instructions

billksun:
I updated the tailwindcss instructions to use `devenv` to launch tailwind.

I removed the redundant `make tailwind-dev` instructions to keep things simple. But let me know if you prefer to keep them for users of older IHP versions.

### DIFF
--- a/Guide/search/package-lock.json
+++ b/Guide/search/package-lock.json
@@ -177,6 +177,29 @@
         "react-dom": ">= 16.8.0 < 18.0.0"
       }
     },
+    "node_modules/@types/prop-types": {
+      "version": "15.7.11",
+      "resolved": "https://registry.npmjs.org/@types/prop-types/-/prop-types-15.7.11.tgz",
+      "integrity": "sha512-ga8y9v9uyeiLdpKddhxYQkxNDrfvuPrlFb0N1qnZZByvcElJaXthF1UhvCh9TLWJBEHeNtdnbysW7Y6Uq8CVng==",
+      "peer": true
+    },
+    "node_modules/@types/react": {
+      "version": "17.0.75",
+      "resolved": "https://registry.npmjs.org/@types/react/-/react-17.0.75.tgz",
+      "integrity": "sha512-MSA+NzEzXnQKrqpO63CYqNstFjsESgvJAdAyyJ1n6ZQq/GLgf6nOfIKwk+Twuz0L1N6xPe+qz5xRCJrbhMaLsw==",
+      "peer": true,
+      "dependencies": {
+        "@types/prop-types": "*",
+        "@types/scheduler": "*",
+        "csstype": "^3.0.2"
+      }
+    },
+    "node_modules/@types/scheduler": {
+      "version": "0.16.8",
+      "resolved": "https://registry.npmjs.org/@types/scheduler/-/scheduler-0.16.8.tgz",
+      "integrity": "sha512-WZLiwShhwLRmeV6zH+GkbOFT6Z6VklCItrDioxUnv+u4Ll+8vKeFySoFyK/0ctcRpOmwAicELfmys1sDc/Rw+A==",
+      "peer": true
+    },
     "node_modules/algoliasearch": {
       "version": "4.10.5",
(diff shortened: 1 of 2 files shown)


Guide/tailwindcss.markdown
```diff
@@ -92,18 +92,48 @@ We also need a CSS entry point for Tailwind. Create a new file at `tailwind/app.
...
+                plugins = [
+                    nodePackages."@tailwindcss/aspect-ratio"
+                    nodePackages."@tailwindcss/forms"
+                    nodePackages."@tailwindcss/language-server"
+                    nodePackages."@tailwindcss/line-clamp"
+                    nodePackages."@tailwindcss/typography"
+                ];
+            })
+        )
+        just
```
amitaibu, line 117: Is the `just` here correct? What is it used for?

amitaibu (commented):
The only drawback I see of this approach is that the tailwind logs are not interwind with other logs. So if there's a compilation error it harder to read it, but I guess it's nicer to have this step executed without having to open a new tab for it.

| `devenv up` | `make ...` |
| - | - |
| image | image |

Guide/tailwindcss.markdown
billksun: Oops, that's what I used instead of `make` before. Thanks for spotting this!

Guide/tailwindcss.markdown
```diff
@@ -92,18 +92,47 @@ We also need a CSS entry point for Tailwind. Create a new file at `tailwind/app.
...
-tailwind-dev:
-	tailwindcss -c tailwind/tailwind.config.js -i ./tailwind/app.css -o static/app.css --watch
+```nix
+...
+ihp = {
+    enable = true;
+    projectPath = ./.;
+    packages = with pkgs; [
+        # Native dependencies, e.g. imagemagick
+        (nodePackages.tailwindcss.overrideAttrs
```
amitaibu, line 106: Maybe worth adding a comment?

```suggestion
        # Use nix to download tailwind and related packages from npm.
        (nodePackages.tailwindcss.overrideAttrs
```

Guide/tailwindcss.markdown
```diff
@@ -92,18 +92,47 @@ We also need a CSS entry point for Tailwind. Create a new file at `tailwind/app.
...
+    haskellPackages = p: with p; [
+        # Haskell dependencies go here
+        ...
+     ];
+};
+
+# Add TailwindCSS build command here
+devenv.shells.default = {
+    processes = {
+        tailwind.exec = "tailwindcss -c tailwind/tailwind.config.js -i ./tailwind/app.css -o static/app.css --watch=always";
```
amitaibu, line 127: Out of curiosity, here we could have arbitrary processes?

```
tailwind.exec = "tailwindcss ..."
foo.exec = "ls -al"
```

billksun: Yep: https://devenv.sh/processes/

You can also add predefined [services](https://devenv.sh/services/) to your devenv as well. I believe IHP has code for using devenv `services` to launch the db, but it is disabled as it breaks some existing `make` commands.

Guide/search/package-lock.json
```diff
@@ -177,6 +177,29 @@
         "react-dom": ">= 16.8.0 < 18.0.0"
       }
     },
+    "node_modules/@types/prop-types": {
```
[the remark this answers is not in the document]
billksun: Good catch. I believe I just ran `make watch` here under the instructions in `Contribute.md`, and it did a bunch of stuff to bring up the dev site for Guides.

Guide/tailwindcss.markdown
billksun: This part was mentioned in the prior section "Adding package to flake.nix", so would the comment be redundant here?

I struggled a bit on the decision to include that snippet here, but in the end I feel including that snippet here makes it easier for users to find where to add the new configuration relative to all the other stuff in `flake.nix`.

amitaibu (approved):
Thanks

mpscholten (approved):
Thanks 👍

PR #1912, opened 2024-02-15
Add loadSchema service

mpscholten:


flake.nix
```diff
@@ -46,6 +46,7 @@
                     services_app = ./NixSupport/nixosModules/services/app.nix;
                     services_worker = ./NixSupport/nixosModules/services/worker.nix;
                     services_migrate = ./NixSupport/nixosModules/services/migrate.nix;
+                    services_loadSchema = ./NixSupport/nixosModules/services/loadSchema.nix;
```
amitaibu, line 49: Should this be before the `services_migrate`?

amitaibu: Oh, it's only declarations, it's not the order of invoking stuff

Issue #1914, opened 2024-02-18
Underscore in ID column leads to compile error in code generation


MonaMayrhofer:
It seems to me that if the primary key of a table contains an underscore, the code generation outputs code that doesn't compile.

This schema
```sql
CREATE TABLE artworks (
    artwork_id UUID DEFAULT uuid_generate_v4() PRIMARY KEY NOT NULL,
    created_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL,
    updated_at TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL
);
```
leads to code generation in `Artwork.hs` looking like this:

```haskell
instance QueryBuilder.FilterPrimaryKey "artworks" where
    filterWhereId artwork_id builder =
        builder |> QueryBuilder.filterWhere (#artworkId, artworkId)
    {-# INLINE filterWhereId #-}
```

Note that the second part of the argument to `filterWhere` is `artworkId`, which is different to the argument of the function, which is named `artwork_id`.
Which results in a compile error
```
    • Couldn't match type: Id' "artworks"
                     with: Artwork' -> Id' "artworks"
 ```

I am not at all well versed with this codebase, but after poking around a bit, I think I might have found the issue.

Within `compileFilterPrimaryKeyInstance`, there seems to be inconsistent naming of the argument of `filterWhereId`, especially line 742 looks out of the ordinary, because it doesn't use `columnNameToFieldName` to convert `artwork_id` to `artworkId`, while everything else does, and line 749 expects the argument to be named `artworkId`.

https://github.com/digitallyinduced/ihp/blob/3ebcd95c9dc21fee4731f17f3968d190ffe5832d/IHP/SchemaCompiler.hs#L732-L749

PR #1915, opened 2024-02-18
fix SchemaCompiler filterPrimaryKey

MonaMayrhofer:
I think this fixes the issue I described in #1914.

I would love to add unit tests for this, however I wasn't able to find the right place to put them - If unit tests are wanted here I'd love a little guidance.

### DIFF
--- a/Test/SchemaCompilerSpec.hs
+++ b/Test/SchemaCompilerSpec.hs
@@ -442,7 +442,55 @@ tests = do
                             builder |> QueryBuilder.filterWhere (#id, id)
                         {-# INLINE filterWhereId #-}
                 |]
-
+        describe "compileFilterPrimaryKeyInstance" do
+            it "should compile FilterPrimaryKey instance when primary key is called id" do
+                let statement = StatementCreateTable $ CreateTable {
+                        name = "things",
+                        columns = [ Column "id" PUUID Nothing False False Nothing ],
+                        primaryKeyConstraint = PrimaryKeyConstraint ["id"],
+                        constraints = [],
+                        unlogged = False
+                    }
+                let compileOutput = compileStatementPreview [statement] statement |> Text.strip
+                
+                getInstanceDecl "QueryBuilder.FilterPrimaryKey" compileOutput `shouldBe` [trimming|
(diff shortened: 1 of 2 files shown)


amitaibu:
> I would love to add unit tests for this

Maybe here? https://github.com/digitallyinduced/ihp/blob/master/Test/IDE/SchemaDesigner/CompilerSpec.hs

MonaMayrhofer:
Im not sure a file in `test/IDE` is the right place tho... the code in question isn't part of the IDE per se, and all the other tests in the file are testing the compiler for Schema.sql and as far as i can tell have nothing to do with the Query Builder, right?

amitaibu:
You are right. Let's get Marc's feedback on that :smile:

mpscholten:
Thanks for the issue and the PR :) (and sorry for the delay in the review)

We could put it into `Test.SchemaCompilerSpec`?

MonaMayrhofer:
That sounds reasonable!👍🏻

Test/SchemaCompilerSpec.hs
```diff
@@ -460,4 +460,4 @@ getInstanceDecl instanceName full =
         takeInstanceDecl (line:rest)
             | isEmpty line = []
             | otherwise = line : takeInstanceDecl rest
-        takeInstanceDecl [] = error "never encountered newline?"
+        takeInstanceDecl [] = [] -- EOF reached
```
MonaMayrhofer, line 511: I would like to draw attention to this change here

In the test cases the instance I am looking for is the last one in the generated code, which resulted in an error, even tho i think it's totally fine. That this results in an error looked like a simple oversight to me so I changed it - the tests still pass... However a second opinion would be great in cause this actually had a deeper reason.

MonaMayrhofer:
I think this is ready for review :blush:

Test/SchemaCompilerSpec.hs
mpscholten: Looks good to me 👍

mpscholten (approved):
Thanks 👍 looks great, will merge once tests have run through

PR #1923, opened 2024-02-27
Allow overriding the logger in the controller context

mpscholten:


IHP/Controller/Context.hs
```diff
@@ -137,5 +138,10 @@ instance HasField "frameworkConfig" ControllerContext FrameworkConfig where
     getField controllerContext = controllerContext.requestContext.frameworkConfig
     {-# INLINABLE getField #-}
 
+-- The following hack is bad, but allows us to override the logger using 'putContext'
+-- The alternative would be https://github.com/digitallyinduced/ihp/pull/1921 which is also not very nice
+--
+-- This design mistake should be fixed in IHP v2
```
amitaibu, line 144: Out of curiosity, what's your idea for fixing on v2?

IHP/Controller/Context.hs
```diff
@@ -137,5 +138,10 @@ instance HasField "frameworkConfig" ControllerContext FrameworkConfig where
...
 instance HasField "logger" ControllerContext Logger where
-    getField controllerContext = controllerContext.frameworkConfig.logger
+    getField context@(FrozenControllerContext { customFields }) = fromMaybe context.frameworkConfig.logger (TypeMap.lookup @Logger)
+    getField context = (unsafePerformIO (freeze context)).logger
```
amitaibu, line 147: I think it's worth adding some comment to what we're doing here -- it's not clear to me

amitaibu (commented):
Looks nicer than the other solution :+1: 

Let's add docs also on where/ how to `put logger`

IHP/Controller/Context.hs
mpscholten: No specific idea yet. We kind of want to make controller context less mutable (e.g. we could  change the type signature from `initContext` to the type signature of `mapContext` in the other branch. That way it's more functional, which would avoid this issue)

IHP/Controller/Context.hs
mpscholten: Just added more comments, hope this clears it a bit

Issue #1924, opened 2024-02-28
Proper support for non "id" primary keys.


MonaMayrhofer:
Currently multiple parts of the `SchemaCompiler` assume that a) the primary key is always called `id` and that b) it is not a composite key.

Some examples for this are:
https://github.com/digitallyinduced/ihp/blob/6971e39134aa10ba1a015cf5e51fd52d7e8c8995/IHP/SchemaCompiler.hs#L585-L590

https://github.com/digitallyinduced/ihp/blob/6971e39134aa10ba1a015cf5e51fd52d7e8c8995/IHP/SchemaCompiler.hs#L617-L623

And possibly others, that are found when investigating further...

Currently using a non-"id" primary key either results in non-compiling code or runtime errors when trying to update (IHP thinks that you forgot a database migration, because "the id column is missing")

After digging in a bit, implementing single-column, non-"id" primary keys seems to be a pretty quick thing, however composite PKs seem to pose a little bit more of a headache, especially for supporting the `genericFetchIds` in `Fetch.hs`, which would probably entail creating a `filterWhereIdIn` in the `FilterPrimaryKey` class in the QueryBuilder (because simply chaining two `QueryBuilder.filterWhereIn` together would yield false positives).
That one should result in generating `WHERE (id_a, id_b) = (?, ?)` predicates in the final sql query... 
That could be done by simply falling back to `sqlQuery`, however as `filterWhereId` is implemented in terms of `QueryBuilder.filterWhere...`, it would probably be better to make `QueryBuilder.filterWhereIn` be able to handle tuples, like
`filterWhereIn (#userId, #thingId) [(1, 2), (3,2)]` or introducing a cousin `QueryBuilder.filterWhereTupleIn` that can do something like that...

I will probably do something along the lines of this anyways in order to incorporate an existing database that i really don't wanna change, the question is whether supporting non-"id" PKs is a goal of ihp, and if a (possibly sizeable) pull request that does that is wanted.

If it is i'd be happy to draft up something proper and collect some more feedback in the form of a PR!

mpscholten:
Happy to have IHP supporting non-id PKs 👍  Ideally we can support that without introducing much new complexity

PR #1925, opened 2024-03-02
Add more support for non-id primary keys

MonaMayrhofer:
These are my efforts to tackle the problems described in #1924.

This PR is still a work in progress.

Goals:
- [X] Create unit tests for relevant code in `SchemaCompiler`
- [x] full support for arbitrary (composite) primaryKeys in SchemaCompiler
- [x] deleteRecord for composite primarykeys
- [x] deleteRecords for composite primarykeys
- [x] genericFetchIds for composite primarykeys
- [x] fix collectionFetchRelated

Other stuff that has to be done to truly support non-id pks, but isn't part of this PR, include (among others?):
- Fix creating migrations with foreign keys to weird primary keys inside the ide
- Create unit test for FromRow, when a composite foreign key is present
- Implement compileFromRow support for composite foreign keys

more goals may get added when i stumble over something that doesnt work with weird primary keys.

Open questions:
~- Is there currently any simple way to test database code, like (`deleteRecord`) either via mocking or using a testing postgres instance?~


mpscholten:
> Is there currently any simple way to test database code, like (deleteRecord) either via mocking or using a testing postgres instance?

The easiest way here might be to create a test IHP app with some CRUD controllers and then load the framework into the app. Check out https://github.com/digitallyinduced/ihp/blob/master/CONTRIBUTING.md#running-an-application-against-a-local-copy-of-the-framework

MonaMayrhofer:
Okay, but there's nothing that i could commit to prevent regressions, right?

IHP/ModelSupport.hs
```diff
@@ -566,21 +566,34 @@ class
...
+--
+-- >>> primaryKeyCondition project
+-- [("id", "d619f3cf-f355-4614-8a4c-e9ea4f301e39")]
+--
+-- If the table has a composite primary key, this returns multiple elements:
+--
+-- >>> primaryKeyCondition postTag
+-- [("post_id", "0ace9270-568f-4188-b237-3789aa520588"), ("tag_id", "0b58fdf5-4bbb-4e57-a5b7-aa1c57148e1c")]
+primaryKeyCondition :: forall record id. (HasField "id" record id, id ~ Id' (GetTableName record), Table record) => record -> [(Text, PG.Action)]
+primaryKeyCondition r = primaryKeyCondition' @record r.id
```
MonaMayrhofer, line 596: `primaryKeyCondition'` is more general than the old `primaryKeyCondition`. Ideally i would swap the naming of the two, however I tried not to introduce too big of a breaking change in the API.

However I am pretty sure this is still a breaking change, due to the changed constraints, and there might be a better way to express this...

IHP/ModelSupport.hs
```diff
@@ -566,21 +566,34 @@ class
...
+--
+-- >>> primaryKeyCondition project
+-- [("id", "d619f3cf-f355-4614-8a4c-e9ea4f301e39")]
+--
+-- If the table has a composite primary key, this returns multiple elements:
+--
+-- >>> primaryKeyCondition postTag
+-- [("post_id", "0ace9270-568f-4188-b237-3789aa520588"), ("tag_id", "0b58fdf5-4bbb-4e57-a5b7-aa1c57148e1c")]
+primaryKeyCondition :: forall record. (HasField "id" record (Id record), Table record) => record -> [(Text, PG.Action)]
+primaryKeyCondition r = primaryKeyCondition' @record r.id
```
MonaMayrhofer, line 596: `primaryKeyCondition'` is more general than the old `primaryKeyCondition`. Ideally i would swap the naming of the two, however I tried not to introduce too big of a breaking change in the API.

However I am pretty sure this is still a breaking change, due to the changed constraints, and there might be a better way to express this...

IHP/ModelSupport.hs
```diff
@@ -566,21 +566,34 @@ class
...
-    -- >>> primaryKeyCondition postTag
+    -- >>> primaryKeyCondition' postTag.id
     -- [("post_id", "0ace9270-568f-4188-b237-3789aa520588"), ("tag_id", "0b58fdf5-4bbb-4e57-a5b7-aa1c57148e1c")]
-    --
-    primaryKeyCondition :: record -> [(Text, PG.Action)]
-    default primaryKeyCondition :: forall id. (HasField "id" record id, ToField id) => record -> [(Text, PG.Action)]
-    primaryKeyCondition record = [("id", toField record.id)]
+    primaryKeyCondition' :: Id record -> [(Text, PG.Action)]
+    default primaryKeyCondition' :: (ToField (Id record)) => Id record -> [(Text, PG.Action)]
+    primaryKeyCondition' id = [("id", toField id)]
```
mpscholten, line 582: Can we give this a better name? E.g. something like `primaryKeyConditionForId`

IHP/ModelSupport.hs
mpscholten: I guess this sShouldn't cause much breakage in most IHP apps

IHP/ModelSupport.hs
MonaMayrhofer: Sure! :blush:

MonaMayrhofer:
I can't find any information about formatting guidelines, and the repository doesn't seem to match ormolu or fourmolu's formatting... is there an official formatter / formatting config, or shall I format stuff by hand to match the rest as good as possible?

mpscholten:
There's no official formatting config, so you can format how it fits best 👍

IHP/ModelSupport.hs
```diff
@@ -566,21 +566,46 @@ class
...
-    -- >>> primaryKeyCondition postTag
+    -- >>> primaryKeyConditionForId postTag.id
     -- [("post_id", "0ace9270-568f-4188-b237-3789aa520588"), ("tag_id", "0b58fdf5-4bbb-4e57-a5b7-aa1c57148e1c")]
-    --
-    primaryKeyCondition :: record -> [(Text, PG.Action)]
-    default primaryKeyCondition :: forall id. (HasField "id" record id, ToField id) => record -> [(Text, PG.Action)]
-    primaryKeyCondition record = [("id", toField record.id)]
+    primaryKeyConditionForId :: Id record -> [(Text, PG.Action)]
+    default primaryKeyConditionForId :: (ToField (Id record)) => Id record -> [(Text, PG.Action)]
+    primaryKeyConditionForId id = [("id", toField id)]
```
MonaMayrhofer, line 594: This is technically not correct, I am assuming here again, that the primary key column is actually called "id". However i don't really see a way around this, if we wanna provide a default implementation. 
Is there already constraint that checks whether "id" is actually a real field within the table?
We could of course generate such a constraint through SchemaCompiler, however that kind of defeats the purpose of the default implementation, because if a user is using SchemaCompiler, they aren't using the default implementation.
I think the cleanest would be to just remove the default implementation, as in most cases it is generated by the SchemaCompiler anyways... Are there any reasons to keep it?

IHP/QueryBuilder.hs
```diff
@@ -826,6 +830,32 @@ filterWhereCaseInsensitive (name, value) queryBuilderProvider = injectQueryBuild
         queryBuilder = getQueryBuilder queryBuilderProvider
 {-# INLINE filterWhereCaseInsensitive #-}
 
+
+filterWhereIdIn :: forall table model queryBuilderProvider (joinRegister :: *). (KnownSymbol table, Table model, model ~ GetModelByTableName table, HasQueryBuilder queryBuilderProvider joinRegister) => [Id model] -> queryBuilderProvider table -> queryBuilderProvider table
+filterWhereIdIn values queryBuilderProvider =
+    -- TODO Null values are ignored here for now, because they need special treatment as in sql they must be compared using "IS NULL"...
+    --  We would a) need to know somehow which values are null (which is not possible with primaryKeyConditionForId returning opaque Actions)
+    --  and b) then decompose the values into something like: (col_a IS NULL AND (col_b, col_c) IN ?) OR (col_b IS NULL AND (col_a, col_c) IN ?)
```
MonaMayrhofer, line 838: To resolve this I would change the `Table` typeclass from 
```haskell
primaryKeyConditionForId :: Id record -> [(Text, PG.Action)]
```
to
```haskell
primaryKeyConditionForId :: Id record -> [(Text, FilterOperator, PG.Action)]
```
and the implementation would get the `FilterOperator` from a call to `toEqOrIsOperator` on the value of the id.

However `Table` is defined in `ModelSupport`, whereas `FilterOperator` and `EqOrIsOperator` are defined in `QueryBuilder` which imports from `ModelSupport`, so in order to avoid cyclic dependencies I would either have to move them and everything they depend on to `ModelSupport` and then reexport from `QueryBuilder` to keep backwards compatibility, but it just seems to me that they don't really belong into `ModelSupport`. We could move them to a whole new `FilterSupport` or something like that...

Or of course I could just leave everything as is and leave this caveat in, as NULLs in primary columns are really a niche thing to do anyways

Whats the best way to progress here?

MonaMayrhofer:
> * [ ] Fix creating migrations with foreign keys to weird primary keys inside the ide
> * [ ]  Create unit test for FromRow, when a composite foreign key is present
> * [ ] Implement compileFromRow support for composite foreign keys

I am tempted to leave these two for a seperate PR in the future, as this one is already getting quite big and convoluted...

The first one should be an easy fix, but it doesn't really fit the spirit of this PR as its in the IDE, so it belongs in a seperate PR imo...

And the last two are a) a rather weird case, which shouldn't come up all too often (a many-to-many relation with unique properties which itself has many of something else...), and b) also doesn't really fit the spirit of this PR as they are more dealing with foreign keys...

IHP/ModelSupport.hs
mpscholten: > Are there any reasons to keep it?

Not really, it can be removed without breaking anything in practise

IHP/QueryBuilder.hs
mpscholten: > We could move them to a whole new FilterSupport or something like that...

Sounds good, let's move it to `IHP.Filter` 👍

mpscholten:
> I am tempted to leave these two for a seperate PR in the future, as this one is already getting quite big and convoluted...

👍

Test/SchemaCompilerSpec.hs
```diff
@@ -148,7 +148,7 @@ tests = do
             it "should compile CanUpdate instance with an array type with an explicit cast" do
                 let statement = StatementCreateTable $ CreateTable {
                     name = "users",
-                    columns = [ Column "id" PUUID Nothing False True Nothing, Column "ids" (PArray PUUID) Nothing False False Nothing],
+                    columns = [ Column "id" PUUID Nothing True True Nothing, Column "ids" (PArray PUUID) Nothing False False Nothing],
```
MonaMayrhofer, line 151: I had to change a whole bunch of existing unit tests, that for some reason specified that the "id" column of their table is nullable.

To me this looked like a simple copy-paste error, but I'd love it if someone confirmed my suspicion.

mpscholten: Yes 👍 looks like a copy paste error to me as well

IHP/QueryBuilder.hs
MonaMayrhofer: I just realized that NULLs in primary keys aren't even a "niche thing to do" but actually not possible, as "Primary Key" implies "Unique" as well as "Not Null"

https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-PRIMARY-KEYS

So this is pretty much all solved already :D

mpscholten: 👍  TIL :D

MonaMayrhofer: I thought I had seen nulls in PKs somewhere, but looking around it seems like its sql standard to not allow that... Must have imagined it :D

MonaMayrhofer:
I think as soon as these tests run through this is pretty much done and ready for review :D

Test/QueryBuilderSpec.hs
```diff
@@ -141,6 +185,46 @@ tests = do
                     (toSQL theQuery) `shouldBe` ("SELECT posts.id, posts.title, posts.external_url, posts.created_at, posts.public, posts.created_by, posts.category_id FROM posts WHERE posts.category_id IS ?", [Plain "null"])
 
 
+        describe "filterWhereIdIn" do
+            it "should produce a SQL with a WHERE condition" do
+                let theValues :: [Id Post] = ["b80e37a8-41d4-4731-b050-a716879ef1d1", "629b7ee0-3675-4b02-ba3e-cdbd7b513553"]
+                let theQuery = query @Post
+                        |> filterWhereIdIn theValues
```
mpscholten, line 192: Would `filterWhereIn (#id, theValues)` also still work here? We use `filterWhereIn` quite a lot in our IHP apps

MonaMayrhofer: Yes, as long as the "id" field isn't virtual and actually corresponds to a column named "id" in the database, this should work exactly as before.
I'll add a unit-test to make sure this is the case.

IHP/ModelSupport.hs
```diff
@@ -645,12 +678,20 @@ deleteRecords records =
...
+deleteRecordByIds ids@(firstId : _) = do
+  let pkCols = case primaryKeyConditionForId @record firstId of
+        [] -> error . cs $ "Impossible happened in deleteRecordById. No primary keys found for table " <> (tableName @record) <> ". At least one primary key is required."
+        [(colName, _)] -> colName
+        ps -> "(" <> intercalate "," (map fst ps) <> ")"
+
+  let theQuery = "DELETE FROM " <> tableName @record <> " WHERE " <> pkCols <> " IN ?"
+
+  let theParameters = PG.Only $ PG.In $ map (ActionTuple . map snd . primaryKeyConditionForId @record) ids
+  sqlExec (PG.Query . cs $! theQuery) theParameters
```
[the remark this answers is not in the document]
MonaMayrhofer: Yup totally! This is fixed together with the abstraction of `ActionTuple . map snd` below.

MonaMayrhofer:
That's looking better^^

Can you take a look if #[dc62630](https://github.com/digitallyinduced/ihp/pull/1925/commits/dc62630c12c5f316c79a8e85409b8c3eb6201e17) has addressed your concerns?

IHP/ModelSupport.hs
```diff
@@ -566,21 +566,66 @@ class
...
     -- If the table has a composite primary key, this returns multiple elements:
     --
-    -- >>> primaryKeyCondition postTag
+    -- >>> primaryKeyConditionForId postTag.id
     -- [("post_id", "0ace9270-568f-4188-b237-3789aa520588"), ("tag_id", "0b58fdf5-4bbb-4e57-a5b7-aa1c57148e1c")]
-    --
-    primaryKeyCondition :: record -> [(Text, PG.Action)]
-    default primaryKeyCondition :: forall id. (HasField "id" record id, ToField id) => record -> [(Text, PG.Action)]
-    primaryKeyCondition record = [("id", toField record.id)]
+    primaryKeyConditionForId :: Id record -> [(Text, PG.Action)]
```
MonaMayrhofer, line 592: I can't find any place within the IHP repo where the first field of the returned tuples of `primaryKeyConditionForId` is actually used, and arguably it really is redundant with the existence of `primaryKeyColumnNames`

I wonder if it would be a good idea to convert `primaryKeyConditionForId` and `primaryKeyCondition` to just return `[PG.Action]`

Or we could retain backward compatibility by just converting `primaryKeyConditionForId` to return just the actions and for `primaryKeyCondition` we zip `primaryKeyColumnNames` together with the actions...

I'm not sure what that would buy us tho, it just feels a bit weird to me to have this redundant information...

mpscholten: > I wonder if it would be a good idea to convert primaryKeyConditionForId and primaryKeyCondition to just return [PG.Action]

👍

mpscholten:
> Can you take a look if #[dc62630](https://github.com/digitallyinduced/ihp/pull/1925/commits/dc62630c12c5f316c79a8e85409b8c3eb6201e17) has addressed your concerns?

Looks great, thanks :)

IHP/ModelSupport.hs
mpscholten: Looking at the git blame on when `primaryKeyCondition` was introduced, it looks it was mainly used by `deleteRecord`. So b.c. isn't much of a concern here.

MonaMayrhofer: Alrighty

IHP/ModelSupport.hs
```diff
@@ -921,6 +973,16 @@ instance ToField value => ToField [value] where
...
+-- | Wraps a list of actions to be used as a Tuple, useful e.g for matching composite keys
+-- >>> toField (ActionTuple [ PG.Escape "myId" ])
+-- Many [Plain "(",Escape "myId",Plain ")"]
+--
+--   Analogous to PGArray from postgres-simple
+newtype ActionTuple = ActionTuple [Action]
+
+instance ToField ActionTuple where
+  toField (ActionTuple actions) = PG.Many ([PG.Plain "("] <> intersperse (PG.Plain ",") actions <> [PG.Plain ")"])
+
```
mpscholten, line 985: Could we simplify this further by getting rid of the newtype and then inlining the definition into `primaryKeyConditionForId` or some other function? This feels like one level of indireciton we might not need (but could be wrong here)

MonaMayrhofer: That would mean that `primaryKeyConditionForId` no longer returns seperate fields for each column...

Which is fine, I guess? Makes it a little bit less flexible, but so far that flexibility isn't needed...

MonaMayrhofer: I'll quickly take a look if this has any other implications

MonaMayrhofer: I think this is a good idea

The primaryKeyCondition would be directly generated by the SchemaCompiler, and the `Table` instances will look like:
```haskell
instance Table CompositeTagging where
    columnNames = ["post_id", "tag_id"]
    primaryKeyColumnNames = ["post_id", "tag_id"]
    primaryKeyConditionForId (Id (postId, tagId)) = PG.Many ([PG.Plain "(", toField postId, PG.Plain "," toField tagId, PG.Plain ")"])

instance Table User where
    columnNames = ["id", "name"]
    primaryKeyColumnNames = ["id"]
    primaryKeyConditionForId (Id id) = PG.Many ([PG.Plain "(", toField id, PG.Plain ")"])

```

I think we could even go so far to make a special case for single-column pks, so that they look like

```haskell
instance Table User where
    columnNames = ["id", "name"]
    primaryKeyColumnNames = ["id"]
    primaryKeyConditionForId (Id id) = toField id
```

which would be rather desirable, but I'll have to look a little bit closer to find out if that would break things

IHP/ModelSupport.hs
```diff
@@ -645,12 +688,15 @@ deleteRecords records =
...
 --
-deleteRecordByIds :: forall record table. (?modelContext :: ModelContext, Show (PrimaryKey table), Table record, ToField (PrimaryKey table), record ~ GetModelByTableName table) => [Id' table] -> IO ()
-deleteRecordByIds ids = do
-    let theQuery = "DELETE FROM " <> tableName @record <> " WHERE id IN ?"
-    let theParameters = (PG.Only (PG.In ids))
-    sqlExec (PG.Query . cs $! theQuery) theParameters
-    pure ()
+deleteRecordByIds :: forall record table. (?modelContext :: ModelContext, Show (PrimaryKey table), Table record, GetTableName record ~ table, record ~ GetModelByTableName table) => [Id' table] -> IO ()
+deleteRecordByIds [] = do
+  pure () -- If there are no ids, we wouldn't even know the pkCols, so we just don't do anything, as nothing happens anyways
```
MonaMayrhofer, line 693: This comment is outdated, we can know the pkCols now that we have primaryKeyConditionColumnSelector. I'll bring this up to date again

MonaMayrhofer:
It's done! Please have a look at the new changes :D

IHP/ModelSupport.hs
```diff
@@ -566,21 +566,62 @@ class
...
+    -- >>> primaryKeyConditionForId project.id
+    -- ["d619f3cf-f355-4614-8a4c-e9ea4f301e39"]
     --
-    primaryKeyCondition :: record -> [(Text, PG.Action)]
-    default primaryKeyCondition :: forall id. (HasField "id" record id, ToField id) => record -> [(Text, PG.Action)]
-    primaryKeyCondition record = [("id", toField record.id)]
+    -- If the table has a composite primary key, this returns multiple elements:
+    --
+    -- >>> primaryKeyConditionForId postTag.id
+    -- ["0ace9270-568f-4188-b237-3789aa520588", "0b58fdf5-4bbb-4e57-a5b7-aa1c57148e1c"]
```
mpscholten, line 591: This comment here is outdated given the output type needs to be of type `PG.Action`

IHP/ModelSupport.hs
```diff
@@ -566,21 +566,62 @@ class
...
+--
+-- >>> primaryKeyCondition project
+-- ["d619f3cf-f355-4614-8a4c-e9ea4f301e39"]
+--
+-- If the table has a composite primary key, this returns multiple elements:
+--
+-- >>> primaryKeyCondition postTag
+-- ["0ace9270-568f-4188-b237-3789aa520588", "0b58fdf5-4bbb-4e57-a5b7-aa1c57148e1c"]
+primaryKeyCondition :: forall record. (HasField "id" record (Id record), Table record) => record -> PG.Action
+primaryKeyCondition r = primaryKeyConditionForId @record r.id
```
mpscholten, line 624: In IHP we try to avoid using single letter variable names:

```suggestion
primaryKeyCondition record = primaryKeyConditionForId @record record.id
```

IHP/ModelSupport.hs
```diff
@@ -645,12 +688,12 @@ deleteRecords records =
...
+deleteRecordByIds :: forall record table. (?modelContext :: ModelContext, Show (PrimaryKey table), Table record, GetTableName record ~ table, record ~ GetModelByTableName table) => [Id' table] -> IO ()
 deleteRecordByIds ids = do
-    let theQuery = "DELETE FROM " <> tableName @record <> " WHERE id IN ?"
-    let theParameters = (PG.Only (PG.In ids))
-    sqlExec (PG.Query . cs $! theQuery) theParameters
-    pure ()
+  let theQuery = "DELETE FROM " <> tableNameByteString @record <> " WHERE " <> (primaryKeyConditionColumnSelector @record) <> " IN ?"
+  let theParameters = PG.Only $ PG.In $ map (primaryKeyConditionForId @record) ids
+  sqlExec (PG.Query $! theQuery) theParameters
+  pure ()
```
mpscholten, line 696: This should be 4 spaces indentation

mpscholten (commented):
Looks like we're almost ready, just found a few small things :)

IHP/ModelSupport.hs
MonaMayrhofer: Ah whoopsie, up in `deleteRecordById` as well i see

mpscholten:
Thanks 👍 great work here!

amitaibu:
@MonaMayrhofer well done :partying_face:

PR #1928, opened 2024-03-05
Make it easier to spot schema + fixture errors

AronNovak:
`make db` right now silently moves on if there's an error in the Schema or in the Fixture. It should not be like that.

### DIFF
--- a/lib/IHP/Makefile.dist
+++ b/lib/IHP/Makefile.dist
@@ -114,10 +114,10 @@ psql: ## Connects to the running postgresql server
 	@psql -h $$PWD/build/db -d app
 
 db: Application/Schema.sql Application/Fixtures.sql ## Creates a new database with the current Schema and imports Fixtures.sql
-	echo "drop schema public cascade; create schema public;" | psql -h $$PWD/build/db -d app
-	psql -h $$PWD/build/db -d app < "${IHP_LIB}/IHPSchema.sql"
-	psql -h $$PWD/build/db -d app < Application/Schema.sql
-	psql -h $$PWD/build/db -d app < Application/Fixtures.sql
+	(echo "drop schema public cascade; create schema public;" | psql -h $${PWD}/build/db -d app) && \
+	psql -v ON_ERROR_STOP=1 -h $${PWD}/build/db -d app < "$${IHP_LIB}/IHPSchema.sql" && \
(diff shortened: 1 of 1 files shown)


amitaibu:
@AronNovak Thanks. Can you share a screenshot of how it will look now

lib/IHP/Makefile.dist
```diff
@@ -114,10 +114,10 @@ psql: ## Connects to the running postgresql server
 	@psql -h $$PWD/build/db -d app
 
 db: Application/Schema.sql Application/Fixtures.sql ## Creates a new database with the current Schema and imports Fixtures.sql
-	echo "drop schema public cascade; create schema public;" | psql -h $$PWD/build/db -d app
-	psql -h $$PWD/build/db -d app < "${IHP_LIB}/IHPSchema.sql"
-	psql -h $$PWD/build/db -d app < Application/Schema.sql
-	psql -h $$PWD/build/db -d app < Application/Fixtures.sql
+	(echo "drop schema public cascade; create schema public;" | psql -h $${PWD}/build/db -d app) && \
```
amitaibu, line 117: Looking at the other make files, we don't use `&& \`

AronNovak: Yep, but this way, we can stop on error. Otherwise it goes on, as each line is executed in a new subshell.

AronNovak:
@amitaibu 
```
INSERT 0 1
INSERT 0 1
ALTER TABLE
ALTER TABLE
ALTER TABLE
ALTER TABLE
INSERT 0 1
ALTER TABLE
ALTER TABLE
INSERT 0 1
ALTER TABLE
ALTER TABLE
ALTER TABLE
ALTER TABLE
ALTER TABLE
ALTER TABLE
INSERT 0 1
ALTER TABLE
ERROR:  relation "public.lm_example_table_foo" does not exist
make: *** [/nix/store/scfcsc0fqpz9gji555wzzzsa0hpnnk9w-source/lib/IHP/Makefile.dist:117: db] Error 3
 aaron   main  ~  gizra  client-app  2  

Most of all, it has exit code `2`, not `0` and the error is immediately visible.
Without this PR, the exit code is `0` and the error is buried in other output lines.

amitaibu:
Thanks, I was missing such a feature :+1:

mpscholten (approved):
Thanks 👍  good addition

PR #1931, opened 2024-03-15
Fix conditional GHC 9.6 compilation

alt-romes:
The CPP macro __GLASGOW_HASKELL__ was using an incorrect integer numbering scheme for GHC versions, making the conditional compilation incorrect (960 means GHC 9.60, whereas 906 means GHC 9.6).

See https://downloads.haskell.org/ghc/latest/docs/users_guide/intro.html#version-numbering


mpscholten:
The existing GHC 9.6 branch was outdated. So I just pushed a new one to https://github.com/digitallyinduced/ihp/tree/ghc906 Right now it's still failing to build because of `hs-connection` (https://discourse.haskell.org/t/fork-of-archived-connection-a-k-a-hs-connection-library/8545/4) which is a dependency of `minio-hs`. I'll try to get this fixed

PR #1932, opened 2024-03-15
GHC 9.6

mpscholten:


mpscholten:
@alt-romes just fixed the broken dependencies via https://github.com/digitallyinduced/ihp/pull/1932/commits/d59d471eda04fd1e82d169799ce4c7343dafb517 It was only two packages depending on the `connection` package. `nix develop --impure` already works on this branch with GHC 9.6 👍 Doing some further tests right now

PR #1934, opened 2024-03-16
Docs about Passing JSON to the View

### LINKED ISSUE #1900: Add to docs/ core the way to encode JSON via hsx
```
import qualified Text.Blaze.Html.Renderer.Text as Renderer
import Text.Blaze.Html (Html)
import qualified Data.Text.Lazy as TL

instance ToJSON Html where
    toJSON html = String (TL.toStrict $ Renderer.renderHtml html)
    
    
<div data-my-encoded-json={encode someJsonValue}></div>
```

amitaibu:
fixes #1900

### DIFF
--- a/Guide/view.markdown
+++ b/Guide/view.markdown
@@ -181,6 +181,64 @@ Now the `company` variable can be used to read the current user's company across
 
 Use [`theRequest`](https://ihp.digitallyinduced.com/api-docs/IHP-ViewSupport.html#v:theRequest) to access the current [WAI request](https://hackage.haskell.org/package/wai-3.2.2.1/docs/Network-Wai.html).
 
+### Passing JSON to the View
+
+You might need to pass JSON values to the view, so later you could have a JS script to read it. You should use Aeson's `toJSON` function to convert your data to JSON and then pass it to the view. Let's say you have a `Posts` controller and `Post`s records that you want to print in the console.
+
+```haskell
+-- Web/Controller/Posts.hs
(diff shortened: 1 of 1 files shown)


Guide/view.markdown
```diff
@@ -181,6 +181,64 @@ Now the `company` variable can be used to read the current user's company across
...
+
+postsToJson :: [Post] -> Value
+postsToJson posts =
+    posts
+        |> fmap (\post -> object
+            [ "id" .= post.id
+            , "title" .= post.title
+            , "body" .= post.body
+            ])
+        |> toJSON
```
mpscholten, line 207: I'd actually move this to the view file as well and then pass the `posts` list directly to the view

amitaibu: I can surely change, but before here's my reasoning: The controller should normally do the heavy lifting. Often it's not as simple as the `postsToJson` function. 

In the beginning my Views would have lots of logic, until I realized it's better to think of them inside the controller and let the view print or filter if needed.

Anyway, either way works, so let me know what you prefer :)

mpscholten: > Often it's not as simple as the postsToJson function

Agree, in those cases I'd also try to keep it in the controller (or even the Helpers)

But in this specific case it IMO really belongs to the view. Especially when the JS code for decoding that json is in the file as well

amitaibu: done

mpscholten (approved):
Thanks 👍

PR #1951, opened 2024-04-18
Fix blank 'Enabling Flakes' section.

### LINKED ISSUE #1897: "Enabling flakes" in the installation guide is empty
This:
https://ihp.digitallyinduced.com/Guide/installation.html#enabling-flakes

Looks like this for me:
Screen Shot 2024-01-21 at 10 15 17

liammcdermott:
Fixes #1897 

The substantive change is on line 49, everything else is either:

1. Comments describing what the code is doing.
2. Changing a few `var` to `const` where appropriate.

This patch just increases the scope of when the section-discovery code stops adding elements to a section. Before, it would keep going until it hit an `<h2>` element, and the section on Nix Flakes is after a `<h3>`, so it would keep munching up elements until it hit _Installing IHP_.

### DIFF
--- a/Guide/spoilify.js
+++ b/Guide/spoilify.js
@@ -1,23 +1,32 @@
 function spoilify () {
     if (document.getElementById("mac") == null) {return;}
-    var mac = document.getElementById("mac");
-    var linux = document.getElementById("linux");
-    var windows = document.getElementById("windows");
+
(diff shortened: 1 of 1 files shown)


mpscholten (approved):
Thanks 👍

PR #1957, opened 2024-04-28
Backoff Strategy docs

amitaibu:
Follow up to https://github.com/digitallyinduced/ihp/pull/1956

### DIFF
--- a/Guide/jobs.markdown
+++ b/Guide/jobs.markdown
@@ -91,6 +91,23 @@ instance Job EmailCustomersJob where
     maxAttempts = 3
 ```
 
+#### Backoff Strategy
+By default, retries will occur every 30 seconds, until maxAttempts is reached. To change the timing so that retries happen every 5 seconds, you can set:
+
+```haskell
+backoffStrategy :: BackoffStrategy
+backoffStrategy = LinearBackoff { delayInSeconds = 5 }
+```
+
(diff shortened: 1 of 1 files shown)


Guide/jobs.markdown
```diff
@@ -91,6 +91,23 @@ instance Job EmailCustomersJob where
     maxAttempts = 3
 ```
 
+#### Backoff Strategy
+By default, retries will occur every 30 seconds, until maxAttempts is reached. To change the timing so that retries happen every 5 seconds, you can set:
```
amitaibu, line 95:
```suggestion
#### Backoff Strategy

By default, retries will occur every 30 seconds, until maxAttempts is reached. To change the timing so that retries happen every 5 seconds, you can set:
```

Guide/jobs.markdown
```diff
@@ -91,6 +91,24 @@ instance Job EmailCustomersJob where
     maxAttempts = 3
 ```
 
+#### Backoff Strategy
+
+By default, retries will occur every 30 seconds, until maxAttempts is reached. To change the timing so that retries happen every 5 seconds, you can set:
```
amitaibu, line 96:
```suggestion
By default, retries will occur every 30 seconds, until `maxAttempts` is reached. To change the timing so that retries happen every 5 seconds, you can set:
```

mpscholten (approved):
Thanks 👍

PR #1969, opened 2024-06-12
Some small improvements

mpscholten:
I've split up the generated haskell modules into even more files for faster builds on large projects. And added some strictness annotations to the logger


amitaibu:
> And added some strictness annotations to the logger

What is the benefit of that?

mpscholten:
Should make it minimally faster. I initially worked on some larger refactorings of the logger to make possibly extract it out into it's own `ihp-logger` package. But that failed as it would break too much b.c., so that was the only commit I could use from that :)

PR #1973, opened 2024-06-16
ihp-troubleshoot should look for `flake.nix` instead of `start` script

amitaibu:
As it's no longer required, since we're using `devenv up`


mpscholten:
👍  maybe then we can change it to look for a `flake.nix`?

amitaibu:
Good idea, changed

mpscholten (approved):

Issue #1867, opened 2023-11-21
Add `queryOrList` to `IHP.QueryBuilder`


amitaibu:
Use would be something like this:

```haskell
  action ProjectsAction = do
      let values :: [(ProjectType, Int)] = [(ProjectTypeOngoing, 3), (ProjectTypeNotStarted, 2)]

          valuePairToCondition :: (ProjectType, Int) -> QueryBuilder "projects"
          valuePairToCondition (projectType, participants) =
              query @Project
                  |> filterWhere (#projectType, projectType)
                  |> filterWhere (#participants, participants)

          theQuery = queryOrList (map valuePairToCondition values)

      projects <- fetch theQuery
      render IndexView { .. }
```

amitaibu:
@mpscholten can you please give some pointers. Maybe I can give it a first try

amitaibu:
I keep hacking my way since there's no `queryOrList` using this code:

```haskell
{-| @todo: Remove when https://github.com/digitallyinduced/ihp/issues/1867 is done.
-}
buildOrQueryStandards :: [QueryBuilder "standards"] -> NoJoinQueryBuilderWrapper "standards"
buildOrQueryStandards (query:rest) = queryUnion query (buildOrQueryStandards rest)
buildOrQueryStandards [] =
    let
        inner =
            query @Standard
            |> filterWhereSql (#id, "IS NULL") -- Always false condition
    in
        queryUnion inner inner -- Terrible hack to make the types work out
```

Would be nice to get a proper solution to IHP core :smile:

PR #1975, opened 2024-06-22
queryUnionList

mpscholten:
Fixes https://github.com/digitallyinduced/ihp/issues/1867

### DIFF
--- a/IHP/QueryBuilder.hs
+++ b/IHP/QueryBuilder.hs
@@ -163,7 +163,8 @@ class HasQueryBuilder queryBuilderProvider joinRegister | queryBuilderProvider -
     getQueryBuilder :: queryBuilderProvider table -> QueryBuilder table
     injectQueryBuilder :: QueryBuilder table -> queryBuilderProvider table
     getQueryIndex :: queryBuilderProvider table -> Maybe ByteString
-    getQueryIndex _ = Nothing 
+    getQueryIndex _ = Nothing
+    {-# INLINE getQueryIndex #-}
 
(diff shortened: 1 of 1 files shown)


amitaibu:
Thank you!

IHP/QueryBuilder.hs
```diff
@@ -1020,6 +1030,26 @@ queryUnion firstQueryBuilderProvider secondQueryBuilderProvider = NoJoinQueryBui
...
+-- >              query @Project
+-- >                  |> filterWhere (#projectType, projectType)
+-- >                  |> filterWhere (#participants, participants)
+-- >
+-- >          theQuery = queryOrList (map valuePairToCondition values)
+-- >
+-- >      projects <- fetch theQuery
+-- >      render IndexView { .. }
+queryUnionList :: forall table. (Table (GetModelByTableName table), KnownSymbol table, GetTableName (GetModelByTableName table) ~ table) => [QueryBuilder table] -> QueryBuilder table
+queryUnionList [] = FilterByQueryBuilder { queryBuilder = query @(GetModelByTableName table) @table, queryFilter = ("id", NotEqOp, Plain "id"), applyLeft = Nothing, applyRight = Nothing }
```
amitaibu, line 1049: With https://github.com/digitallyinduced/ihp/pull/1925, is `"id"` correct?  (That is, are we guarnteed to havd an `record.id` ?)

mpscholten: Good point. Likely we should use `head (primaryKeyColumnNames @(GetModelByTableName table))` here. Or even better support free form SQL snippets here (we just want `WHERE false` here, but there's no way to write this with the current query builder)

PR #1980, opened 2024-06-25
Make overmind the default process

amitaibu:


amitaibu:
I couldn't see `overmind` working locally with this change. Maybe it's the wrong location?

Issue #1824, opened 2023-09-23
helper.js does not submit multiple checkbox values via query parameter


billksun:
When submitting a form with multiple checkbox values of the same group using the GET method, the browser is suppose to repeat the checkbox name with each value in the query parameter:

```
?name=value1&name=value2&name=value3
```

However, with `helper.js` enabled, only the last value gets appended to the query parameters:

```
?name=value3
```

This interferes with the server-side parameters parsing and will result in a `NoConstructorMatched` error.

Removing `helper.js` from `Web/View/Layout.hs` fixes this issue.

mpscholten:
This was already completed

PR #1986, opened 2024-07-11
Fix multiple checkboxes submit

andreytroeglazov:
#1824 

This change will add all submitted checkboxes as query params on form submit


mpscholten (approved):
Good catch, thanks!

PR #1988, opened 2024-07-13
Fix docs about ImageStyle

amitaibu:
With current fix, the image style is created with the correct name, as the original file.

### DIFF
--- a/Guide/file-storage.markdown
+++ b/Guide/file-storage.markdown
@@ -869,6 +869,7 @@ import Web.Controller.Prelude
 import IHP.ControllerSupport
 import System.Directory (doesFileExist)
 import qualified Data.Text as Text
+import qualified Data.UUID as UUID (fromString)
 
 instance Controller ImageStyleController where
     action RenderImageStyleAction { width, height, originalImagePath } = do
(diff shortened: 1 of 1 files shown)


Guide/file-storage.markdown
```diff
@@ -891,6 +892,8 @@ instance Controller ImageStyleController where
                 let options :: StoreFileOptions = def
                         { directory = imageStylePathDirectory
                         , preprocess = applyImageMagick "jpg" ["-resize", cs size <> "^", "-gravity", "center", "-extent", cs size, "-quality", "85%", "-strip"]
+                        -- Keep the original filename.
+                        , fileName = UUID.fromString (cs uuid)
```
mpscholten, line 896: We can simplify this by using `fromText` directly:

```suggestion
                        , fileName = UUID.fromText uuid
```

Guide/file-storage.markdown
```diff
@@ -869,6 +869,7 @@ import Web.Controller.Prelude
 import IHP.ControllerSupport
 import System.Directory (doesFileExist)
 import qualified Data.Text as Text
+import qualified Data.UUID as UUID (fromString)
```
amitaibu, line 872:
```suggestion
import qualified Data.UUID as UUID (fromText)
```

mpscholten (approved):

Issue #1989, opened 2024-07-13
Save only the relative path of the filename in `storeFileWithOptions`


amitaibu:
Right now, on  `StaticDirStorage ` we save the filename with the full file path `https://example.com/static/image/foo.jpg`

https://github.com/digitallyinduced/ihp/blob/4526f58b29741f351b66b070f165111d83a0efc1/IHP/FileStorage/ControllerFunctions.hs#L106-L115

I think it's better to save it as `/image/foo.jpg`; it takes less space and is more flexible if the base path is changed.

mpscholten:
Agree. I think it was done to be consistent with the S3 storage method (as that also generates a full absolute url). But I see the point in changing that, so in case it's not breaking anything let's do this 👍

PR #1990, opened 2024-07-13
Save only the relative path of the filename for `StaticDirStorage`

amitaibu:
fixes https://github.com/digitallyinduced/ihp/issues/1989


amitaibu:
I still need to verify it locally, but looking here:

https://github.com/digitallyinduced/ihp/blob/a58f7a0ee23bb541c9e0f2c90c42195d613cf2cb/IHP/FileStorage/ControllerFunctions.hs#L223-L229

It looks like it's expecting `objectPath` to be relative. So Likely this is also a bug fix.

Test/FileStorage/ControllerFunctionsSpec.hs
```diff
@@ -0,0 +1,60 @@
...
+            let objectPath = "https://example.com/static/test.txt"
+            temporaryDownloadUrl <- createTemporaryDownloadUrlFromPath objectPath
+
+            temporaryDownloadUrl.url `shouldBe` "https://example.com/static/test.txt"
+
+createControllerContext = do
+    let
+        requestBody = FormBody { params = [], files = [] }
+        request = Wai.defaultRequest
+        requestContext = RequestContext { request, respond = error "respond", requestBody, frameworkConfig = error "frameworkConfig" }
```
amitaibu, line 58: @mpscholten :wave: I suspect this line, with the `error`, causes the error I get when running the test

image

How can I make it work? (Copied from [here](https://github.com/digitallyinduced/ihp/blob/7da31c2d02a7c8eb41a63ebc4d30bbf6aacbeafe/Test/Controller/CookieSpec.hs#L36-L42))

mpscholten: That GHC error is unrelated to the error call. I just pushed a workaround via https://github.com/digitallyinduced/ihp/commit/66518dee183d71bc5a1e1df59c1e8019298a0c0c to master (appartently the INLINE triggers the bug in GHC, but chaging that to an optional inline via INLINABLE not triggers the bug). Can you sync with master and then check again?

amitaibu: Thanks. I think now it hits that `error`

image

mpscholten: You can use `IHP.FrameworkConfig.withFrameworkConfig` to get a framework config object, like this:

```haskell
    let withFrameworkConfig = IHP.FrameworkConfig.withFrameworkConfig (pure ())
    describe "createTemporaryDownloadUrlFromPathWithExpiredAt" $ do
        it "returns baseUrl concatenated with objectPath when objectPath does not start with http:// or https://" $ do
            withFrameworkConfig \frameworkConfig -> do
                context <- createControllerContext frameworkConfig
                let ?context = context
                let objectPath = "static/test.txt"
                temporaryDownloadUrl <- createTemporaryDownloadUrlFromPath objectPath

                temporaryDownloadUrl.url `shouldBe` "http://localhost:8000/static/test.txt"
```

And then use the frameworkConfig passed to `createControllerContext` instead of the `error` call

amitaibu: With this I get 

```
  ./IHP/Prelude.hs:92:17: 
  1) IHP.FileStorage.ControllerFunctions.storeFileWithOptions returns the objectPath without the baseUrl
       uncaught exception: ErrorCall
       Could not find FileStorage in config. Did you call initS3Storage from your Config.hs?
         error, called at ./IHP/Prelude.hs:92:17 in main:IHP.Prelude
```

While trying to figure how to pass a config with the `initStaticDirStorage` (didn't figure yet), I found `aroundAll (withIHPApp WebApplication config) do`

Maybe we should use that? (If so, not sure how :sweat_smile: )

amitaibu: I was able to inti the storage [here](https://github.com/digitallyinduced/ihp/pull/1990/commits/600d9feae61ff75157d6eec80735b894c1a8aff1)

Test/FileStorage/ControllerFunctionsSpec.hs
```diff
@@ -0,0 +1,70 @@
...
+
+                -- We pass the UUID that will be used as the filename, so we can easily assert the objectPath.
+                let options :: StoreFileOptions = def
+                        { fileName = Just "4c55dac2-e411-45ac-aa10-b957b01221df"
+                        , directory = "Test.FileStorage.ControllerFunctionsSpec"
+                        }
+
+                result <- storeFileWithOptions fileInfo options
+
+                result.url `shouldBe` ("Test.FileStorage.ControllerFunctionsSpec/4c55dac2-e411-45ac-aa10-b957b01221df")
```
mpscholten, line 43: Wouldn't it be better to prepend a `/` to the path? Otherwise you cannot really use the url e.g. in an `` tag or for something else.

```suggestion
                result.url `shouldBe` ("/Test.FileStorage.ControllerFunctionsSpec/4c55dac2-e411-45ac-aa10-b957b01221df")
```

amitaibu: Right, I thought I've done it, maybe I've reverted. Let me check :smile:

amitaibu: Fixed in https://github.com/digitallyinduced/ihp/pull/1990/commits/adfaf3c28ae352df0e7cbd9ae3e8e65e7279dbd4

PR #1993, opened 2024-07-26
Add appName To options for flake.

Montmorency:
When hosting multiple ihp apps on a particular server it is helpful for debugging/maintenance/tooling to be able to inspect which binaries are in the production servers /nix/store by their name (not just hash). This is consistent with all other nixpkgs that contain the hash and the package name in the store. 

For managed production clusters it is also  nice for roll back to be able to give semver names to app packages so that it is easy to instantly roll back to a prev version of app binary on a prod server if required. This PR adds option of appName to ihp flake-modules so users can specify the appName. The appName option gets passed all the way to ihp's (un)optimized-prod-server package `mkDerivation` (where it is called `name` and is a standard attribute for nix's mkDerivation.). We could call it `name` at the top level so it matches specifically but in the context of the app level flake it is probably less ambiguous to call it `appName`.

Furthermore having an appName option at top level flake has been found useful for a number of other IHP production helpers i.e. scripts to synchronize databases and deploy to servers etc. that can leverage the appName variable as well.

### DIFF
--- a/NixSupport/default.nix
+++ b/NixSupport/default.nix
@@ -11,6 +11,7 @@
 , optimized ? false
 , includeDevTools ? !optimized # Include Postgres?
 , rtsFlags ? ""
+, appName ? "app"
 , optimizationLevel ? "2"
 }:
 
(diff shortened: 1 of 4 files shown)


Montmorency:
Fixed comments, indentation and whitespace.

mpscholten (approved):
Thanks 👍

PR #1995, opened 2024-08-02
Drop blaze

mpscholten:


amitaibu:
Is the idea that since we only support `Text`, we are able to simplify the implementation? Is the difference significant?

mpscholten:
Yep, and with the simple implementation GHC can optimize hsx expressions better. This results in less memory usage and faster runtime performance (we should do a benchmark to figure out how significant the difference is).

Ideally we switch from Text to https://hackage.haskell.org/package/mason which provides even faster string concatenation.

PR #1996, opened 2024-08-06
Add empty option value to styledSelectFormField for required html validation

andreytroeglazov:


amitaibu:
@andreytroeglazov Thanks. Errors seem unrelated. Can you please check if removing those imports fixes the issue

image

andreytroeglazov:
If `required == True` we need `value=""` on placeholder so the html validation will work correctly.
If `required == False` it does not matter if we have `value=""` or not. But to keep things consistent I think it's better to always have `value=""` on placeholder option.

mpscholten:
Thanks, got it now 👍

mpscholten:
@andreytroeglazov the tests need to be updated as well https://github.com/digitallyinduced/ihp/actions/runs/10289623933/job/28477778824 can you take a look?

andreytroeglazov:
@mpscholten sure, will check

PR #1998, opened 2024-08-13
Implement `inherits` for Schema.sql

amitaibu:
I won't lie, so far ChatGPT has done the work :smile:

### DIFF
--- a/Test/IDE/SchemaDesigner/CompilerSpec.hs
+++ b/Test/IDE/SchemaDesigner/CompilerSpec.hs
@@ -628,7 +628,7 @@ tests = do
                     , indexType = Nothing
                     }
             compileSql [statement] `shouldBe` sql
-        
+
         it "should escape an index name inside a 'CREATE INDEX' statement" do
             let sql = cs [plain|CREATE INDEX "Some Index" ON "Some Table" ("Some Col");\n|]
             let statement = CreateIndex
@@ -841,7 +841,7 @@ tests = do
                         )
                     }
             compileSql [policy] `shouldBe` sql
-        
+
         it "should compile 'CREATE POLICY' statements with a 'ihp_user_id() IS NOT NULL' expression" do
             -- https://github.com/digitallyinduced/ihp/issues/1412
             let sql = "CREATE POLICY \"Users can manage tasks if logged in\" ON tasks USING (ihp_user_id() IS NOT NULL) WITH CHECK (ihp_user_id() IS NOT NULL);\n"
@@ -1040,7 +1040,7 @@ tests = do
         it "should compile 'CREATE UNLOGGED TABLE' statements" do
             let sql = [trimming|
                 CREATE UNLOGGED TABLE pg_large_notifications (
-                
+
                 );
             |] <> "\n"
             let statements = [
@@ -1068,3 +1068,22 @@ tests = do
                             }
                         ]
             compileSql statements `shouldBe` sql
+
+        it "should compile a CREATE TABLE statement with INHERITS" do
+            let sql = "CREATE TABLE child_table (\n    id UUID PRIMARY KEY\n) INHERITS (parent_table);\n"
+            let statement = StatementCreateTable CreateTable
+                    { name = "child_table"
+                    , columns = [Column
+                        { name = "id"
+                        , columnType = PUUID
+                        , defaultValue = Nothing
+                        , notNull = False
+                        , isUnique = False
+                        , generator = Nothing
+                        }]
+                    , primaryKeyConstraint = PrimaryKeyConstraint ["id"]
+                    , constraints = []
+                    , unlogged = False
+                    , inherits = Just "parent_table"
+                    }
+            compileSql [statement] `shouldBe` sql
--- a/ihp-ide/IHP/IDE/SchemaDesigner/Compiler.hs
+++ b/ihp-ide/IHP/IDE/SchemaDesigner/Compiler.hs
@@ -22,7 +22,7 @@ compileSql statements = statements
     |> unlines
 
 compileStatement :: Statement -> Text
-compileStatement (StatementCreateTable CreateTable { name, columns, primaryKeyConstraint, constraints, unlogged }) = "CREATE" <> (if unlogged then " UNLOGGED" else "") <> " TABLE " <> compileIdentifier name <> " (\n" <> intercalate ",\n" (map (\col -> "    " <> compileColumn primaryKeyConstraint col) columns <> maybe [] ((:[]) . indent) (compilePrimaryKeyConstraint primaryKeyConstraint) <> map (indent . compileConstraint) constraints) <> "\n);"
+compileStatement (StatementCreateTable CreateTable { name, columns, primaryKeyConstraint, constraints, unlogged, inherits }) = "CREATE" <> (if unlogged then " UNLOGGED" else "") <> " TABLE " <> compileIdentifier name <> " (\n" <> intercalate ",\n" (map (\col -> "    " <> compileColumn primaryKeyConstraint col) columns <> maybe [] ((:[]) . indent) (compilePrimaryKeyConstraint primaryKeyConstraint) <> map (indent . compileConstraint) constraints) <> "\n)" <> maybe "" (\parentTable -> " INHERITS (" <> compileIdentifier parentTable <> ")") inherits <> ";"
 compileStatement CreateEnumType { name, values } = "CREATE TYPE " <> compileIdentifier name <> " AS ENUM (" <> intercalate ", " (values |> map TextExpression |> map compileExpression) <> ");"
 compileStatement CreateExtension { name, ifNotExists } = "CREATE EXTENSION " <> (if ifNotExists then "IF NOT EXISTS " else "") <> compileIdentifier name <> ";"
 compileStatement AddConstraint { tableName, constraint = UniqueConstraint { name = Nothing, columnNames } } = "ALTER TABLE " <> compileIdentifier tableName <> " ADD UNIQUE (" <> intercalate ", " columnNames <> ")" <> ";"
--- a/ihp-ide/IHP/IDE/SchemaDesigner/Parser.hs
+++ b/ihp-ide/IHP/IDE/SchemaDesigner/Parser.hs
@@ -98,6 +98,8 @@ createTable = do
         columnsAndConstraints <- ((Right <$> parseTableConstraint) <|> (Left <$> parseColumn)) `sepBy` (char ',' >> space)
         pure (lefts columnsAndConstraints, rights columnsAndConstraints)
 
+    inherits <- optional parseInheritsClause
+
     char ';'
 
     -- Check that either there is a single column with a PRIMARY KEY constraint,
@@ -221,6 +223,12 @@ parseOnDelete = choice
         , (lexeme "SET" >> ((lexeme "NULL" >> pure SetNull) <|> (lexeme "DEFAULT" >> pure SetDefault)))
         , (lexeme "CASCADE" >> pure Cascade)
         ]
+
+parseInheritsClause :: Parser Text
+parseInheritsClause = do
+    lexeme "INHERITS"
+    parentTable <- between (char '(' >> space) (char ')' >> space) qualifiedIdentifier
+    pure parentTable
 
 parseColumn :: Parser (Bool, Column)
 parseColumn = do
--- a/ihp-ide/IHP/IDE/SchemaDesigner/Types.hs
+++ b/ihp-ide/IHP/IDE/SchemaDesigner/Types.hs
@@ -85,6 +85,7 @@ data CreateTable
       , primaryKeyConstraint :: PrimaryKeyConstraint
       , constraints :: [Constraint]
       , unlogged :: !Bool
+      , inherits :: Maybe Text
       }
   deriving (Eq, Show)
 


CONTRIBUTING.md
```diff
@@ -82,7 +82,7 @@ use `make console` to load your application together with the framework located
 
 ```
 ghci
-:l IHP/exe/IHP/IDE/DevServer.hs
+:l IHP/ihp-ide/exe/IHP/IDE/DevServer.hs
```
amitaibu, line 85: @mpscholten I think the doc is outdated.

1. Is `make console`, mentioned above , needed?
2. I'm getting an error
image

mpscholten: It could just be `ghci`. `make console` is just an alias for `ghci`.

Can you try `:set -iIHP/ihp-ide` in the ghci? This will add the `IHP/ihp-ide` directory to the load path. After that it should load without the error.

amitaibu: Thanks, it helped (adapted docs)

But now I get

image

mpscholten: I can reproduce that. Just tried it locally and here's an approach how I got it working with latest IHP:

```
# Assuming workingDirectory is the project

cd IHP # switch to IHP
direnv allow # load direnv things if this is freshly cloned

ghci # Load ghci will all IHP dependencies
:cd .. # switch back to project directory
:set -iIHP
:set -iIHP/ihp-ide
 :l IHP/ihp-ide/exe/IHP/IDE/DevServer.hs
```

That should work now

locally i run into a ghc bug while doing that, I hope it works for you:

```
ghci> :l IHP/ihp-ide/exe/IHP/IDE/DevServer.hs
Ok, 164 modules loaded.
ghci> main
panic! (the 'impossible' happened)
  GHC version 9.8.2:
	nameModule
  internal $fHasQueryBuilderkJoinQueryBuilderWrapperjoinRegister1_iefl
  Call stack:
        callStackDoc, called at compiler/GHC/Utils/Panic.hs:191:37 in ghc-9.8.2-inplace:GHC.Utils.Panic
        pprPanic, called at compiler/GHC/Types/Name.hs:341:3 in ghc-9.8.2-inplace:GHC.Types.Name
    panic, called at compiler/GHC/Utils/Error.hs:503:29 in ghc-9.8.2-inplace:GHC.Utils.Error


Please report this as a GHC bug:  https://www.haskell.org/ghc/reportabug
```

mpscholten: As a workaround for the GHC bug I reverted the commit that introduced the bug by running `git revert 4fe8b56e5a8c5dcb394a16b0dfb5eb2d6f77b18f` from within the `IHP` directory

mpscholten: I've ran into more issues with the dev server right now and added a new helper called [`mainInParentDirectory` to fix that](https://github.com/digitallyinduced/ihp/commit/27a696af90efcd0f84260d84805dd2dc753e818c).

After you've synced this branch with latest master it should now work like this:
```
# Assuming workingDirectory is the project

cd IHP # switch to IHP
direnv allow # load direnv things if this is freshly cloned

ghci # Load ghci will all IHP dependencies
 :l ihp-ide/exe/IHP/IDE/DevServer.hs
mainInParentDirectory # Will run the IHP project in the parent project directory
```

amitaibu: Thank you for looking into it :pray: 

I'm confused about which URL should be in my `flake.nix`

https://github.com/digitallyinduced/ihp/blob/27a696af90efcd0f84260d84805dd2dc753e818c/CONTRIBUTING.md?plain=1#L43-L67

I'm guessing we should always tell folks to point to the local IHP, or is there a different use case?

amitaibu:
Failing tests belong to https://github.com/digitallyinduced/ihp/blob/27a696af90efcd0f84260d84805dd2dc753e818c/Test/SchemaCompilerSpec.hs that require some trailing spaces. Let's see if we can remove those :)

amitaibu:
Next, I want to see if we can have revisions created either manually or by using triggers.

Manual will be something like this:

```sql
CREATE TABLE post_revisions (
    revision_id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
    post_id UUID NOT NULL REFERENCES posts(id) ON DELETE CASCADE,
    revised_at TIMESTAMPTZ DEFAULT now(),
    -- Inherit all fields from the posts table
) INHERITS (posts);
```

```haskell
action CreatePostRevisionAction { postId } = do
    post <- fetch postId
    let revision = newRecord @PostRevision
            |> set #postId post.id
            |> set #title post.title
            |> set #content post.content
            |> set #revisedAt (getCurrentTime)
    createRecord revision
    -- Now update the original post if needed
    updateRecord post
    redirectTo ShowPostAction { .. }
```

amitaibu:
The `INHERTIS` keyword is now recognized in Schema.sql. Next is figuring how to auto-generate types/getter/setter based on the columns from the parent table

image

amitaibu:
@mpscholten I've started working on the `SchemaCompiler`, but I've hit a panic

image

What are you doing in such cases?

mpscholten:
try `git revert 4fe8b56e5a8c5dcb394a16b0dfb5eb2d6f77b18f` (that commit causes the GHC panic). After that restart the ghc and it should work

amitaibu:
> try git revert 4fe8b56e5a8c5dcb394a16b0dfb5eb2d6f77b18f

Thanks. How did you know it was that commit? So I'll know for next time :smile:

mpscholten:
When this specific bug happend the first time I was looking through the git history and just tried out various commits until I found the one that reproduced it. This was like a week or two weeks after that commit was created, so it was only a few commits I had to try out

ihp-ide/IHP/SchemaCompiler.hs
```diff
@@ -336,20 +336,44 @@ primaryKeyTypeName :: Text -> Text
...
+
+        compileParentFields parentTable =
+            let parentTableDef = findTableByName parentTable
+            in parentTableDef
+                |> maybe [] (dataFields . (.unsafeGetCreateTable))
+                -- Remove the MetaBag field from the parent table.
+                -- @todo: Avoid clashing of field names.
+                |> filter (\(fieldName, _) -> fieldName /= "meta")
+                |> map (\(fieldName, fieldType) -> fieldName <> " :: " <> fieldType)
+                |> commaSep
```
amitaibu, line 369: @mpscholten With this code I'm getting inside `Generated/ActualTypes.hs`

```
data PostRevision' postId = PostRevision {id :: (Id' "posts"), title :: Text, body :: Text, userId :: userId, postRevisions :: postRevisions, revisionId :: (Id' "post_revisions"), postId :: postId, meta :: MetaBag} deriving (Eq, Show)
```

`postRevisions :: postRevisions` is wrong in this case. Not sure how to know when to filter it out

mpscholten: I think you need to add a type argument for `postRevisions`:

```
-- current:
data PostRevision' postId = 

-- fixed:
data PostRevision' postId postRevisions = 
```

ihp-ide/IHP/SchemaCompiler.hs
```diff
@@ -857,7 +906,30 @@ compileDataTypePattern :: (?schema :: Schema) => CreateTable -> Text
...
+            else currentTypeArguments <> " " <> parentTypeArguments
+
+        currentTypeArguments = dataTypeArguments table |> unwords
+
+        modelName = tableNameToModelName name
+
+        -- @todo: Find a better way.
+        colName = modelName |> pluralize |> Text.toLower
+
+        parentTypeArguments :: Text
```
amitaibu, line 922: @mpscholten I continue to hack through this. The latest error is this

image

Here's the generated code. Are you able to spot what I'm doing wrong in terms of types?

```haskell
data PostRevision' postId userId = PostRevision {title :: Text, body :: Text, userId :: userId, id :: (Id' "post_revisions"), postId :: postId, meta :: MetaBag} deriving (Eq, Show)

type instance PrimaryKey "post_revisions" = UUID

type PostRevision = PostRevision' (Id' "posts")

type instance GetTableName (PostRevision' _) = "post_revisions"
type instance GetModelByTableName "post_revisions" = PostRevision

instance Default (Id' "post_revisions") where def = Id def

instance () => Table (PostRevision' postId userId) where
    tableName = "post_revisions"
    tableNameByteString = Data.Text.Encoding.encodeUtf8 "post_revisions"
    columnNames = ["id","post_id","id","title","body","userId","postRevisions"]
    primaryKeyColumnNames = ["id"]
    primaryKeyConditionForId (Id (id)) = toField id
    {-# INLINABLE primaryKeyConditionForId #-}
```

mpscholten: It looks like the `type PostRevision` is wrong

```haskell
-- Current
type PostRevision = PostRevision' (Id' "posts")

-- Should be:
type PostRevision = PostRevision' (Id' "posts") (Id' "users")
```

ihp-ide/IHP/SchemaCompiler.hs
```diff
@@ -903,14 +1043,48 @@ compileSetFieldInstances table@(CreateTable { name, columns }) = unlines (map co
...
+                in parentTableDef
+                        |> maybe [] (dataFields . (.unsafeGetCreateTable))
+                        -- We remove ref to own table (e.g. `post_revisions` table should not have postRevisions)
+                        -- @todo: Check name of `id` column.
+                        |> filter (\(fieldName, _) -> fieldName /= "meta" && Text.toLower fieldName /= colName && fieldName /= "id")
+
+
+        allDateFields = currentDataFields <> parentDataFields <> [("meta", "MetaBag")]
+
+        compileSetField (name, fieldType) = "instance UpdateField " <> tshow name <> " (" <> compileTypePattern table <>  ") (" <> compileTypePattern' name  <> ") " <> valueTypeA <> " " <> valueTypeB <> " where\n    {-# INLINE updateField #-}\n    updateField newValue (" <> compileDataTypePattern table <> ") = " <> modelName <> " " <> (unwords (map compileAttribute (allDateFields |> map fst)))
```
amitaibu, line 1084: @mpscholten At this point I'm just trying to get things to compile. It would require code cleanup (but open for early review, if you see some evident things).

Right now I have this error - and I admit those types are confusing me :sweat_smile: 

image

```haskell
instance UpdateField "postId" (PostRevision' postId userId) (PostRevision' postId' userId) postId postId' where
    {-# INLINE updateField #-}
    updateField newValue (PostRevision id postId title body userId meta) = PostRevision id newValue title body userId (meta { touchedFields = "postId" : touchedFields meta })
```

amitaibu: Ok, I think I know why, because the order of fields is not consistent, as previously defined:

```
data PostRevision' postId userId = PostRevision {title :: Text, body :: Text, userId :: userId, id :: (Id' "post_revisions"), postId :: postId, meta :: MetaBag} deriving (Eq, Show)
```

I'll fix that

amitaibu:
Got it to compile with [this](https://github.com/digitallyinduced/ihp/pull/1998#issuecomment-2288427645) code! :balloon:

amitaibu:
I've updated the OP with a checklist. I'm not sure about some of the stuff, as I think it's not a small feature :smile:

mpscholten:
Happy to help if there's more questions :)

amitaibu:
Thanks. 

> NEW.user_id

Right. I guess I wanted to have another column called `revision_user_id` and have that populated with the `ihp_user_id()`.

Is there a technical reason not to inject the user ID also if row level security isn't used?

mpscholten:
> Is there a technical reason not to inject the user ID also if row level security isn't used?

Yes, right now the `ihp_user_id()` requires a second database role to be available. You can see some code here https://github.com/digitallyinduced/ihp/blob/master/IHP/ModelSupport.hs#L437 In theory we could also support `ihp_user_id()` without that second database role, but that would require some rework of the RLS code.

amitaibu:
## Indicate inherits on the UI

image

amitaibu:
I'm trying to see how migration works. I'm assuming `inherits` isn't enough to update columns, in case the parent columns have changed.

So I've added a new col to the parent table.

image

But on `migrate db` I get this error

image

Server is running via the new `mainInParentDirectory` on `ghci`

mpscholten:
Could it be that the postgres server is not running? If that's the case, you could manually start one by running `make postgres` in a different tab

amitaibu:
No, it's running. The app and ide run well. Just the migrate fails

mpscholten:
If you run the pg_dump command from the error message in a terminal yourself, does it output some SQL?

amitaibu:
```
➜  IHP git:(core-inherits) ✗ pg_dump
pg_dump: [archiver (db)] connection to database "amitaibu" failed: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: FATAL:  role "amitaibu" does not exist
```

But it's running on `ghci`, and the IDE & app are running

image

image

amitaibu:
To workaround the error, I'm checking it by running `devenv up`, and then the migration doesn't error due to missing connection

mpscholten:
This has been superseded by #2505 which implements the same feature against the current codebase structure (monorepo with ihp-postgres-parser/, ihp-schema-compiler/, ihp-ide/ packages) and also covers the hasql-based statement generators that were added since this PR was opened.

PR #2001, opened 2024-09-03
Fix grepping for direnv, add profile files to messages

kukimik:
I'm on Linux Mint (an Ubuntu derivative). I'm using the bash shell. The file `~/.bash_profile` doesn't exist on my machine. My `~/.bashrc` file contains the `eval "$(direnv hook bash)"` line. 

The script displayed a `grep` error saying that the `~/.bash_profile` file does not exist and instructed me to add the line above to `~/.bashrc`, which I already did.

It looks like the fix works in my config. I didn't try it on `zsh`. Unfortunately I'm not sure if this is a POSIX-compliant solution.


mpscholten (approved):
Sorry for slow merge 👍 and thanks! According to o1-preview it's posix compliant 👍

PR #2004, opened 2024-10-09
Add capitalisation to view generator

### LINKED ISSUE #1950: IHP codegen does not enforce capitalised View names
The steps I took, as an IHP beginner, just working with the web console:

1. Created a table named `users`, using CodeGen
2. Clicked CodeGen in the console, then _Controller_
3. Entered `users` in the _Controller name_ text box (notice the console corrects `users` → `Users` so it's capitalised)
4. This all seemed to work, so I went to _CodeGen_ again, and selected _Users_ from the drop-down. I was unsure what it meant _precisely_ to create a View in this context, so in the _View name_ text box I entered `test`.
5. Clicked _Preview_ then _Generate_

The CodeGen failed to generate valid Haskell code (since type and module names cannot be lowercase). The attached screenshot shows the _Preview_ step, including the invalid code.

image

liammcdermott:
Fixes #1950

### DIFF
--- a/IHP/NameSupport.hs
+++ b/IHP/NameSupport.hs
@@ -13,6 +13,7 @@ module IHP.NameSupport
 , fieldNameToColumnName
 , escapeHaskellKeyword
 , tableNameToControllerName
+, tableNameToViewName
 , enumValueToControllerName
 , toSlug
 , module IHP.NameSupport.Inflections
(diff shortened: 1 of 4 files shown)


liammcdermott:
This should be ready for review.

Some notes:

- If the user enters a lowercase or snake_case name for their view, the generator updates it to proper case (like the controller generator)
- The view generator was outputting Haskell types like `data Name = {Name}` due to a typo, this is fixed in the generator and the tests
- Added tests for view names like `edit_view` and `editView` - similar to tests for the controller generator
- Added a `tableNameToViewName` function in `NameSupport`. It's just a synonym of `tableNameToControllerName` right now, but it seems like a good idea:
    1) to avoid confusion (_why is the view generator calling a function related to controllers?_)
    2) in case the ways the names are transformed should change in future, it will be very easy to just alter the implementation of `tableNameToViewName`

mpscholten (approved):
Looks great 👍  thanks for fixing this!

liammcdermott:
@mpscholten You're welcome. Thanks for reviewing this so quickly!

Issue #1987, opened 2024-07-11
Update jQuery to latest version


amitaibu:
Currently, we have 3.6.0, and the latest release is 3.7.1

https://github.com/digitallyinduced/ihp/tree/a125f9ef9aa617d0a95349f082a1849afb9d8e17/lib/IHP/static/vendor

<img width="1433" alt="image" src="https://github.com/digitallyinduced/ihp/assets/125707/ec7cd09a-9e65-4ea8-a278-a669e21fa6d9">

amitaibu:
@andreytroeglazov care to tackle? 😄

PR #2007, opened 2024-10-17
Update jquery

andreytroeglazov:
#1987


lib/IHP/static/vendor/jquery-3.5.0.min.js
mpscholten, line 1: Deleting the old JS files will break existing IHP apps, so for now we sadly need to keep them :/

UPGRADE.md
```diff
@@ -942,9 +942,9 @@ Now you can start your project as usual with `./start`.
 
 ## Updating jQuery
 
-The jQuery version used by IHP has changed. We're switching from `3.2.1` to `3.6.0`. jQuery `3.2.1` has some known security vulnerabilities, so it's recommended that you follow these update steps.
```
mpscholten, line 945: This refers to an older version of IHP. Ideally you could add a new entry for the next IHP release and add a note on the jquery upgrade

andreytroeglazov: Thanks, sorry, this line confused me, will fix it 🙂

mpscholten:
This was solved yesterday

PR #2008, opened 2024-10-22
Use GHC 9.8

mpscholten:
We can remove our custom GHC patches now as they've been upstreamed.

According to https://github.com/NixOS/nixpkgs/pull/346720#issuecomment-2427628892 we need to wait until ghc-lib-parser has been fixed for 9.8.3, otherwise haskell language server won't work.


CharlesEkkel:
@mpscholten is this still blocked? Looks like a version of `ghc-lib-parser` for 9.8.3 is on hackage now as best I can tell: https://hackage.haskell.org/package/ghc-lib-parser-9.8.3.20241022.

mpscholten:
The package update of `ghc-lib-parser` is done in https://github.com/NixOS/nixpkgs/pull/351154 but that's not been merged yet. Also it looks like haskell language server doesn't work with 9.8.3 yet (see https://github.com/NixOS/nixpkgs/pull/350904#issuecomment-2439028953)

So seems to be still blocked

Issue #1985, opened 2024-07-09
NixOS Deployment: Support sessionSecretFile


mpscholten:
**Feedback by @nomeata **

> The docs suggest to include the session secret right in the nix config, as
>                       sessionSecret = "xxx";
> but this isn’t particularly great, secrets shouldn’t be in the repository, and not where they might end up in the nix store. Maybe the default setup should create a secret on the server if it isn't there and store it in a file, and users who want more control can copy it there?

PR #2012, opened 2024-11-12
nixos: add sessionSecretFile option

mpscholten:
Closes https://github.com/digitallyinduced/ihp/issues/1985


nomeata:
Looks good so far (didn’t test, though)

NixSupport/nixosModules/options.nix
```diff
@@ -75,6 +75,21 @@ with lib;
 
         sessionSecret = mkOption {
             type = types.str;
+            descriptiom = ''
```
amitaibu, line 78: => `description` ?

mpscholten: good catch, will fix

PR #2015, opened 2024-11-19
Migration path Env Var.

Montmorency:
Adds an environment variable so user can run migration systemd service from any directory on FileSystem. Useful in certain production situations. Defaults to the previous value. Docs updated.

### DIFF
--- a/Guide/database-migrations.markdown
+++ b/Guide/database-migrations.markdown
@@ -73,6 +73,20 @@ migrate
 
 A good value for `MINIMUM_REVISION` is typically the unix timestamp of the time when the database was initially created.
 
+
+### IHP_MIGRATIONS_DIR
+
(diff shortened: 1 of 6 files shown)


mpscholten (approved):
Looks good to me 👍

mpscholten:
With this we can now simplify the migration service here https://github.com/digitallyinduced/ihp/blob/master/NixSupport/nixosModules/services/migrate.nix#L10 a little bit

Montmorency:
> With this we can now simplify the migration service here https://github.com/digitallyinduced/ihp/blob/master/NixSupport/nixosModules/services/migrate.nix#L10 a little bit

Yes that line confuses me 😄 . What can it be switched to now?

mpscholten:
simplified version is here https://github.com/digitallyinduced/ihp/pull/2016/files

PR #2021, opened 2024-12-14
Refactor pagination SQL queries to use generic paramaters

rvarun11:
Changes examples to use [] to match the expected [Action] parameter type.

### DIFF
--- a/IHP/Pagination/ControllerFunctions.hs
+++ b/IHP/Pagination/ControllerFunctions.hs
@@ -15,18 +15,12 @@ module IHP.Pagination.ControllerFunctions
 import IHP.Prelude
 import IHP.Controller.Context
 import IHP.Controller.Param ( paramOrDefault, paramOrNothing )
-
-import IHP.Pagination.Types
-    ( Options(..), Pagination(..) )
-
-import IHP.QueryBuilder
-    ( HasQueryBuilder, filterWhereILike, limit, offset )
+import IHP.Pagination.Types ( Options(..), Pagination(..) )
+import IHP.QueryBuilder ( HasQueryBuilder, filterWhereILike, limit, offset )
 import IHP.Fetch (fetchCount)
-
 import IHP.ModelSupport (GetModelByTableName, sqlQuery, sqlQueryScalar, Table)
 
-import Database.PostgreSQL.Simple.ToField (toField, Action)
-import Database.PostgreSQL.Simple.Types (Query(Query))
+import Database.PostgreSQL.Simple (FromRow, ToRow, Query(..), Only(Only), (:.)(..))
 
 -- | Paginate a query, with the following default options:
 --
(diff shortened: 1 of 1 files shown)


mpscholten:
Will the empty list not generate a type error?

rvarun11:
No, using the  `()` you get
```
 Couldn't match expected type ‘[Database.PostgreSQL.Simple.ToField.Action]’
```

It will get rid of that error but you may need to annotate it with the type of `model` if you get the following:
```
 The type variable ‘model0’ is ambiguous
 ```
 to fix that, you can:
 ```haskell
 (users, pagination) <- paginatedSqlQuery @User "SELECT * FROM users" []
```

Should I add this as well?

amitaibu:
Yes, I think it's best to have a fully working example.

rvarun11:
@amitaibu I've updated the examples as discussed :)

mpscholten:
The type of `paginatedSqlQuery` is wrong. It needs to be a type parameter. E.g. `sqlExec` or `sqlQuery` in IHP use a generic `query` type argument with a `ToRow q` constraint:

https://github.com/digitallyinduced/ihp/blob/17f17b5f1bcc3475df2ba8e90f1c7dffd5ae9773/IHP/ModelSupport.hs#L416

rvarun11:
Hey @mpscholten, I've updated the PR to what we discussed.

CSchank:
@rvarun11 We played with this exact problem recently and ran into lots of weird stuff with the constraints. I believe the `FromRow` constraint is unnecessary and will constrain things more than needed. For instance, `sqlQuery` doesn't require it: https://ihp.digitallyinduced.com/api-docs/IHP-ModelSupport.html#v:sqlQuery. It would mean you can't use certain types of values as inputs, if they don't have FromRow instances, at least that's what we ran into. 

I think it seems to be needed here because of the subquery to do the count query. We actually implemented this exact function in our project a few months ago and were able to avoid the need for `FromRow`. Our code isn't open source, but I'll provide you a link to our version in case it's useful.

CSchank:
Update: sorry, I got confused. `FromRow` is on the output result, not the input. So I think this matches what we came up with. I think for a while our type signature was more complex and I remember us running into some other weird things but we both arrived at the same type signature in the end. 😊

rvarun11:
@mpscholten It is ready for review

IHP/Pagination/ControllerFunctions.hs
```diff
@@ -179,12 +173,12 @@ defaultPaginationOptions =
 --
 -- *AutoRefresh:* When using 'paginatedSqlQuery' with AutoRefresh, you need to use 'trackTableRead' to let AutoRefresh know that you have accessed a certain table. Otherwise AutoRefresh will not watch table of your custom sql query.
 paginatedSqlQuery
-  :: forall model
-   . ( FromRow model
+  :: ( FromRow r
```
mpscholten, line 176: Why was the type variable name changed from `model` to `r`?

IHP/Pagination/ControllerFunctions.hs
```diff
@@ -202,14 +196,14 @@ paginatedSqlQuery = paginatedSqlQueryWithOptions defaultPaginationOptions
 --
 -- *AutoRefresh:* When using 'paginatedSqlQuery' with AutoRefresh, you need to use 'trackTableRead' to let AutoRefresh know that you have accessed a certain table. Otherwise AutoRefresh will not watch table of your custom sql query.
 paginatedSqlQueryWithOptions
-  :: forall model
-   . ( FromRow model
+  :: ( FromRow r
```
mpscholten, line 199: same here

IHP/Pagination/ControllerFunctions.hs
rvarun11: To maintain consistency with similar functions: 

https://github.com/digitallyinduced/ihp/blob/17f17b5f1bcc3475df2ba8e90f1c7dffd5ae9773/IHP/ModelSupport.hs#L379

mpscholten: Makes sense, but let's keep the old name. Single variable names are not good. (We should eventually patch `sqlQuery` to fix this there as well)

rvarun11: I see. That makes sense. I've updated the PR accordingly.

mpscholten (approved):

PR #2044, opened 2025-02-21
Force evaluation of blaze HTML also on `ProdMode`

amitaibu:
Even when running on the latest master  that has [print exceptions to log](https://github.com/digitallyinduced/ihp/pull/2038), looking at logs - connected to AWS CloudWatch - we don't see that extra info.

image (7)

Marc wrote:

```haskell
let devHandlers =
            [ postgresHandler
            , paramNotFoundExceptionHandler
            , patternMatchFailureHandler
            , recordNotFoundExceptionHandlerDev
            ]

    -- Prod handlers should not leak any information about the system
    let prodHandlers =
            [ recordNotFoundExceptionHandlerProd
            ]

    let allHandlers = if ?context.frameworkConfig.environment == Environment.Development
            then devHandlers
            else prodHandlers

```
so in prodMode (which your app likely is running in), only the recordNotFoundExceptionHandlerProd handler is available. And that handler always returns a 404 status code. Your logs indicate a 500 status code, so it's either in the genericHandler or not in the displayException at all.

So this PR will force eval, the same as in the dev handlers.

See https://github.com/digitallyinduced/ihp/issues/1028

### DIFF
--- a/IHP/Controller/Render.hs
+++ b/IHP/Controller/Render.hs
@@ -26,25 +26,18 @@ renderPlain text = respondAndExit $ responseLBS status200 [(hContentType, "text/
 
 respondHtml :: (?context :: ControllerContext) => Html -> IO ()
 respondHtml html =
-        -- The seq is required to force evaluation of `maybeEvaluatedBuilder` before returning the IO action. See below for details
-        maybeEvaluatedBuilder `seq` (respondAndExit $ responseBuilder status200 [(hContentType, "text/html; charset=utf-8"), (hConnection, "keep-alive")] maybeEvaluatedBuilder)
+        -- The seq is required to force evaluation of `evaluatedBuilder` before returning the IO action. See below for details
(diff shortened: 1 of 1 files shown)


amitaibu:
If we do this change, maybe there's no meaning anymore to devHandlers vs prodHandlers.

So we can remove that code as well

mpscholten (approved):

mpscholten:
> If we do this change, maybe there's no meaning anymore to devHandlers vs prodHandlers.

This difference is primarly because in dev mode we can show more detailed error messags (e.g. including the specific sql query). In prod we don't want to show these details to avoid accidentally leaking secrets etc.

amitaibu:
Indeed. I think the PR isn't not ready yet, as I want to confirm that on `ProdMode` we [log](https://github.com/digitallyinduced/ihp/blob/b69611ba7401a915bcf2cc3785060ffd9010567f/IHP/ErrorController.hs#L103) the same error as on Dev, but send the browser to limited error message

amitaibu:
Actually, let's start with the PR as is, and if needed we can have a follow up

PR #2046, opened 2025-02-21
Delete `run-tests`

amitaibu:
The file appears to be outdated and isn't functioning properly.

We already have https://github.com/digitallyinduced/ihp/blob/master/CONTRIBUTING.md#running-tests on how to run the tests


mpscholten (approved):
agree 👍

PR #2047, opened 2025-02-21
Be more strict and prevent incomplete patterns

### LINKED ISSUE #1681: Be more strict and prevent incomplete patterns
Currently, we have the `-Wincomplete-patterns` flag, which results in easily overlooked warnings.
Can we convert this to a compiler error instead?

amitaibu:
fixes #1681

### DIFF
--- a/.ghci
+++ b/.ghci
@@ -44,6 +44,7 @@
 :set -XDeepSubsumption
 :set -XOverloadedRecordDot
 :set -Werror=missing-fields
+:set -Werror=incomplete-patterns
 :set -fwarn-incomplete-patterns
 :set -Wno-ambiguous-fields
 :set -O0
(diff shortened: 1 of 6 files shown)


amitaibu:
I didn't get any error when running `devenv up` locally—it's as if it doesn't have any effect. Let's see if GH actions will get an error.

amitaibu:
@mpscholten Errors on GH actions don't seem to be related.

image


I believe it's coming from `master`, but I don't understand why it sometimes works on `master` and sometimes doesn't.

mpscholten:
For it to work inside the app itself we need to add it to https://github.com/digitallyinduced/ihp/blob/master/ihp-ide/lib/IHP/applicationGhciConfig aswell 👍

mpscholten:
I tried to reproduce the tests error but it works locally when checking out this branch. So let's ignore that for now

mpscholten:
@georgefst agree 👍

mpscholten:
the world has changed a lot since this PR was created. With agents doing most of the work now we should have this enabled by default in IHP apps. So I created a new a PR for this

PR #2048, opened 2025-02-22
Revert "Force evaluation of blaze HTML also on `ProdMode`"

amitaibu:
Reverts digitallyinduced/ihp#2044

I believe this PR can be reverted. I've originally reported not seeing the exceptions being caught on logs.

However, my mistake was that I expected it to be shown as part of the 500 error. In reality, it's logged as a _separate_ log item.

In the screenshot, we can see that my app has sent those logs on 19 Feb -- two days before #2044 was merged. 

image


amitaibu:
On the other hand, we may decide to keep the PR un-reverted, since the code currently on `master` is simpler, and might be more predictable. @mpscholten your call :smile:

mpscholten:
Let's keep it in master for now and see what happens 👍

PR #2052, opened 2025-03-05
Show routing error only on DEV mode

### LINKED ISSUE #2051: Don't leak IHP related info in Production error messages
Image

I think we can keep the upper and lower part, but we shouldn't show the `Routing failed with: NoConstructorMatched {expectedType = "IHP.ModelSupport.Id'", value = Nothing, field = "lmActivityId"}`, since:

1. It's to technical for any end user
2. We shouldn't expose internal structures

Code is coming from here

https://github.com/digitallyinduced/ihp/blob/033293fd9dcea635e2af40ab86986c822b296ac6/IHP/ErrorController.hs#L330

amitaibu:
fixes #2051

| Dev | Production |
| -- | -- |
| image | image |

### DIFF
--- a/IHP/ErrorController.hs
+++ b/IHP/ErrorController.hs
@@ -326,16 +326,16 @@ handleRouterException exception request respond =
     let ?context = ?applicationContext
     in case fromException exception of
         Just Router.NoConstructorMatched { expectedType, value, field } -> do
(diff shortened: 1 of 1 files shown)


mpscholten (approved):

mpscholten:
Thanks 👍

PR #2055, opened 2025-03-22
Add Lucid2 backend to ihp-hsx

iteratee:
This PR does 2 things, and I'll understand if you only want to merge the first one.

1. It splits out `ihp-hsx` into two libraries: `ihp-hsx-core` that does all the parsing, without needing `blaze-html`. And `ihp-hsx`, that includes all of the blaze-specific code. It keeps the same name for compatibility, and it re-exports all the modules in `ihp-hsx-core` for the same reason.
2. It adds `ihp-hsx-lucid`: A library that uses `ihp-hsx-core` to parse HSX, and then has Lucid2 specific template haskell. It doesn't depend on `blaze-html` or `blaze-markup`.

If you don't want the lucid bits in tree, please consider the first half. I'd be happy to publish `ihp-hsx-lucid` on hackage.

Consider this PR a compliment on a very nicely written library. I was able to add Lucid2 support in an afternoon.


mpscholten (commented):
Thanks 👍 looks good to me.

The tests right now have some redundancies. Maybe we can have a single function where can pass in different `shouldBeHtml` functions (one for blaze, one for lucid2). Then we could have one `ihp-hsx-tests-core` and two `ihp-hsx-tests-blaze` `ihp-hsx-tests-lucid` test suites

mpscholten:
One more idea, instead of `ihp-hsx-core` we could name it `ihp-hsx-parser`, that would be more descriptive. wdyt?

iteratee:
> Thanks 👍 looks good to me.
> 
> The tests right now have some redundancies. Maybe we can have a single function where can pass in different `shouldBeHtml` functions (one for blaze, one for lucid2). Then we could have one `ihp-hsx-tests-core` and two `ihp-hsx-tests-blaze` `ihp-hsx-tests-lucid` test suites

They use different quasi quoters as well. I thought about this, but I couldn't see a good way to re-use the infrastructure.

Let me sleep on this bit.

iteratee:
> One more idea, instead of `ihp-hsx-core` we could name it `ihp-hsx-parser`, that would be more descriptive. wdyt?

I will do that.

I also fixed the failing tests. I missed adding lucid2 to the nix package list.

iteratee:
> > Thanks 👍 looks good to me.
> > The tests right now have some redundancies. Maybe we can have a single function where can pass in different `shouldBeHtml` functions (one for blaze, one for lucid2). Then we could have one `ihp-hsx-tests-core` and two `ihp-hsx-tests-blaze` `ihp-hsx-tests-lucid` test suites
> 
> They use different quasi quoters as well. I thought about this, but I couldn't see a good way to re-use the infrastructure.
> 
> Let me sleep on this bit.

There are also 2 examples that give slightly different answers:

The doctype doesn't insert a newline, and the comment test doesn't pad with spaces.

I double checked the specifications, and the Lucid answer is fine in both cases.


Although, even with those 2 cases, it would be better to share the remainder of the tests.

mpscholten:
> Although, even with those 2 cases, it would be better to share the remainder of the tests.

👍

iteratee:
OK, I did the rename to `ihp-hsx-parser`. That was easy.

I got the tests to share most of the test cases. I did it by creating a different `QuasiQuoter`, one that produces a result for both Lucid and Blaze. (Just for testing of course). To make gluing pieces together work, I made the data type an instance of both `ToHtml` classes.

Then I made a `shouldBeSameHtml` function, that calls the correct serializer for both backends, and compares it against a single expected output. It fails if either doesn't match.

This makes it easy to see which test cases are byte-for-byte the same, and which ones differ. There are 3. Two are whitespace differences, and I checked against the specifications. Lucid's answers seem to be inline.

The third is an attribute that is unescaped. The particular data type is specific to Blaze, so there is a mirror test with the corresponding Lucid Datatype.

I'm pleased with it at the moment. I only worry that the shared testing is a little too clever.

iteratee:
Actually, I just ran into an issue that showed up in usage, but not in the current test suite.

I'll see if I can't write a test for it, and then fix it up.

Don't merge this for the moment.

iteratee:
OK, I'm much happier with this.

The wrapper type family ends up being necessary to help GHC know that the monad type of any intermediate results ought to be the same monad as the outer result.

Fortunately, users will only see this if they need to define `EmbedAsHtml` for a new type, which is straightforward. The `QuasiQuoter` unwraps the entire result expression making it more transparent.

iteratee:
I think the version with effects is useful, and worth keeping.

But I'm thinking about bringing back the version that forced `m` to `Identity`.

It's more approachable, and if you don't need the extra complexity, it's probably better to avoid it.

I think the best way is to just export two `QuasiQuoters`: `hsx` and `hsxM`. `hsxM` lets you embed effects, but is slightly more complicated, and you shouldn't reach for it by default. Haddock can help point the way to the simpler one.

iteratee:
> I think the version with effects is useful, and worth keeping.
> 
> But I'm thinking about bringing back the version that forced `m` to `Identity`.
> 
> It's more approachable, and if you don't need the extra complexity, it's probably better to avoid it.
> 
> I think the best way is to just export two `QuasiQuoters`: `hsx` and `hsxM`. `hsxM` lets you embed effects, but is slightly more complicated, and you shouldn't reach for it by default. Haddock can help point the way to the simpler one.

What do you think? The simpler version is there in the commit history.

iteratee:
OK, I just went ahead and made 2 versions:

`hsx` allows pure values to be embedded. This should be easier to use, and less likely to run into a type error.
`hsxM` allows you to embed monadic values, including values that produce effects and html. This makes it easier to put a header or a footer in a library, with effects that look things up, and then to just include it as `[hsxM|<body>{header} ... {footer}</body>|]`

I made all of the shared tests run with the both Lucid versions. There's a test specific to Lucid with effects.

Sorry for the noise as I duck-code-review. I think this version is a keeper.

iteratee:
Most recent push fixes a minor issue if the module that is spliced into has `OverloadedLists` enabled. GHC can no longer typecheck the call to `sequence_` because the list expression is now ambiguous. The solution was straightforward, add a type application to the two `sequence_` calls that could be spliced in, specifying that it should be over lists.

mpscholten (approved):
Thanks 👍  looks good to me

PR #2057, opened 2025-04-04
Improve docs on adding speicifc package versions

amitaibu:


Guide/package-management.markdown
```diff
@@ -282,6 +282,8 @@ mkDerivation {
 
 Save this package definition code to a new file in `Config/nix/haskell-packages/google-oauth2.nix`. IHP projects are configured to automatically pick up any Haskell package definitions in the `Config/nix/haskell-packages` directory. So this package definition will be used automatically.
 
+Ensure you add this new nix package to git `git add Config/nix/haskell-packages`
```
mpscholten, line 285: we should explain why this is needed 👍

mpscholten: something like this 🗡️ 
```
Ensure you add this new nix package to git `git add Config/nix/haskell-packages`, as nix will only pick up files that are part of the git repository. If you forget to add it, Nix may silently ignore your changes, and your build won’t reflect the new package.
```

amitaibu:
```suggestion
Ensure you add this new nix package to git `git add Config/nix/haskell-packages`, as nix will only pick up files that are part of the git repository. If you forget to add it, Nix may silently ignore your changes, and your build won’t reflect the new package.
```

mpscholten (approved):
Thanks :)

PR #2059, opened 2025-04-05
Fix createTemporaryDownloadUrlFromPath missing a slash

### LINKED ISSUE #2058: Missing slash when saving signedUrl
I followed the docs from https://ihp.digitallyinduced.com/Guide/file-storage.html#required-uploads-with-a-file-record

and I got `http://localhost:8000uploaded_files/0a9a5590-9a85-40f3-b84f-9bdccc728e9e` in the DB

I'm quite sure I'm the one to blame, as I remember touching that area, so self assigning 😄

amitaibu:
fixes #2058

### DIFF
--- a/IHP/FileStorage/ControllerFunctions.hs
+++ b/IHP/FileStorage/ControllerFunctions.hs
@@ -231,8 +231,8 @@ createTemporaryDownloadUrlFromPathWithExpiredAt validInSeconds objectPath = do
             let url = if any (`isPrefixOf` objectPath) urlSchemes
                     -- BC, before we saved only the relative path of a file, we saved the full URL. So use it as is.
                     then objectPath
-                    -- We have the relative path (prefixed with slash), so add the baseUrl.
-                    else frameworkConfig.baseUrl <> objectPath
+                    -- We have the relative path (without with slash prefix), so add the baseUrl.
+                    else frameworkConfig.baseUrl <> "/" <> objectPath
 
             pure TemporaryDownloadUrl { url = cs url, expiredAt = publicUrlExpiredAt }
         S3Storage { connectInfo, bucket} -> do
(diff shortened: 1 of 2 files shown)


amitaibu:
@mpscholten good to go?

IHP/FileStorage/ControllerFunctions.hs
```diff
@@ -231,8 +231,8 @@ createTemporaryDownloadUrlFromPathWithExpiredAt validInSeconds objectPath = do
             let url = if any (`isPrefixOf` objectPath) urlSchemes
                     -- BC, before we saved only the relative path of a file, we saved the full URL. So use it as is.
                     then objectPath
-                    -- We have the relative path (prefixed with slash), so add the baseUrl.
-                    else frameworkConfig.baseUrl <> objectPath
+                    -- We have the relative path (without with slash prefix), so add the baseUrl.
+                    else frameworkConfig.baseUrl <> "/" <> objectPath
```
mpscholten, line 235: Won't this break if existing objectPath's that are already in the database already have a `/` in front?

amitaibu: We didn't have a release with that, it was only on `master` for a while

amitaibu: But if you think, I can check if `objectPath` already has `/` or not, before crafting the URL

amitaibu: I'll adapt PR, to check for the "/", so this function is more robust

amitaibu: Improved logic, with test

https://github.com/digitallyinduced/ihp/pull/2059/commits/53baf82a9b2f103760c891380a8d19590a38a33c

mpscholten (approved):
Thanks 👍

PR #2061, opened 2025-04-11
filterWhereInCaseInsensitive

andreytroeglazov:


IHP/QueryBuilder.hs
```diff
@@ -78,6 +79,7 @@ import qualified Data.ByteString.Char8 as ByteString
 import qualified Data.ByteString.Lazy as LByteString
 import qualified Control.DeepSeq as DeepSeq
 import qualified Data.Text.Encoding as Text
+import Data.Text (toLower)
```
andreytroeglazov, line 82: Not sure if it's correct to import it here

mpscholten (approved):
Thanks 👍

PR #2062, opened 2025-04-28
Add docs for using validateFieldIO

stephenbenedict:
First time doing a pull request so hope I'm doing this the right way. One part that needs clarification, but which I do not understand sufficiently to explain, is why only 1 argument (`post`) is included when calling the custom validator function `belongsToCategory`:

```
|> (\post -> validateFieldIO #subCategoryId (belongsToCategory post) post) 
```

### DIFF
--- a/Guide/validation.markdown
+++ b/Guide/validation.markdown
@@ -213,6 +213,57 @@ Then just call it like:
 user |> validateField #age isAge`
 ```
 
+### Creating a custom validator that uses IO
+
+Example scenario: you have 3 tables: `posts`, `categories`, and `sub_categories`.
+
+The `posts` table has the fields `category_id` and `sub_category_id` (which reference the `categories` and `sub_categories` tables).
+
+The `sub_categories` table has a `category_id` field to determine which category a sub-category belongs to.
+
+So when creating a new post, you want to validate that its `sub_category_id` is for a `sub_category` record that includes a `category_id` that matches the post’s `category_id`.
+
+Since the validation will require querying the database, add a type signature to the `buildPost` function specifying that it requires the `?modelContext` to be available and specify that the return value will be `IO Post` rather than `Post`.
+
+``` 
+buildPost :: (?modelContext :: ModelContext, ?context :: ControllerContext) => Post -> IO Post
+buildPost post = post
+    |> fill @'["title","categoryId", "subCategoryId"]
+```
+
+Next, add the custom validator:
+
+```
+buildPost :: (?modelContext :: ModelContext) => Post -> IO Post
+buildPost post = post
+    |> fill @'["title","categoryId", "subCategoryId"]
+    -- Add the following custom validator.
+    |> (\post -> validateFieldIO #subCategoryId (belongsToCategory post) post) 
+        where
+            
+            belongsToCategory :: (?modelContext :: ModelContext) => Post -> Id SubCategory -> IO ValidatorResult
+            belongsToCategory post subCategoryId = do
+                let categoryId = post.categoryId
+                subCategory <- fetch subCategoryId
+                if subCategory.categoryId == categoryId
+                    then pure $ Success
+                    else pure $ Failure "Subcategory does not belong to the selected category"
+```
+
+Finally, in your controller, wherever you use the `buildPost` function, since it is now inside IO, use the >>= (bind) operator rather than |> (pipe) operator after it. 
+
+```
+ post
+    |> buildPost
+    >>= ifValid \case -- Changed |> to >>=
+        Left post -> do
+            render NewView { .. } 
+        Right post -> do
+            post <- post |> createRecord
+            setSuccessMessage "Post created"
+            redirectTo PostsAction
+```
+
 ### Checking If A Record Is Valid
 
 Use [`ifValid`](https://ihp.digitallyinduced.com/api-docs/IHP-Controller-Param.html#v:ifValid) to check for validity of a record:


stephenbenedict:
@amitaibu I forgot to add `?context :: ControllerContext` but it seems to work without it. Why is that?

Guide/validation.markdown
```diff
@@ -213,6 +213,57 @@ Then just call it like:
...
 
+### Creating a custom validator that uses IO
+
+Example scenario: you have 3 tables: `posts`, `categories`, and `sub_categories`.
+
+The `posts` table has the fields `category_id` and `sub_category_id` (which reference the `categories` and `sub_categories` tables).
+
+The `sub_categories` table has a `category_id` field to determine which category a sub-category belongs to.
+
+So when creating a new post, you want to validate that its `sub_category_id` is for a `sub_category` record that includes a `category_id` that matches the post’s `category_id`.
```
amitaibu, line 224: Thanks for creating the PR!

I believe we should simplify this example. Following the logic has become more challenging than the `validateFieldIO`. :smile: 

How about Checking if the title already existing in another post.


```
    pinCodeIsUnique :: (?modelContext :: ModelContext) => Post -> Text -> IO ValidatorResult
    pinCodeIsUnique post title = do
        exists <-
            query @Post
                |> filterWhere (#title, ...)
                |> filterWhereNot (#id, ...)
                |> fetchExists


        if exists
            then pure $ Failure ...
            else pure Success
```

stephenbenedict: Hah, you are absolutely right. Just committed a simplified version.

Guide/validation.markdown
```diff
@@ -213,6 +213,46 @@ Then just call it like:
...
+Modify the `buildPost` function like:
+
+```haskell
+-- Add type signature specifying that it requires the Model Context and Controller Context
+-- to be available and that the function’s return value will be IO Post rather than Post.
+buildPost :: (?modelContext :: ModelContext, ?context :: ControllerContext) => Post -> IO Post
+buildPost post = post
+    |> fill @'["title", "body"]
+    -- Add custom validator. The below example checks if the post’s title is unique.
+    |> (\post -> validateFieldIO #title (titleIsUnique post) post)
```
mpscholten, line 227:
```suggestion
    |> validateFieldIO #title (titleIsUnique post)
```

Guide/validation.markdown
```diff
@@ -213,6 +213,46 @@ Then just call it like:
...
+            titleIsUnique post title = do
+                exists <-
+                    query @Post
+                        |> filterWhere (#title, post.title)
+                        |> filterWhereNot (#id, post.id)
+                        |> fetchExists
+        
+                if exists
+                    then pure $ Failure "Title is not unique"
+                    else pure Success
```
mpscholten, line 239: Maybe we can extract this into a top level function instead of the `where` clause

Guide/validation.markdown
amitaibu: Is `post`, at this point that one that was piped and filled?

mpscholten: oh, true, i missed that, sorry. Then it needs to stay how it is right now 👍

Guide/validation.markdown
```diff
@@ -213,6 +213,46 @@ Then just call it like:
...
+```
+
+In your controller, wherever you use the `buildPost` function, since it is now inside IO, use the `>>=` (bind) operator rather than `|>` (pipe) operator after it. 
+
+```haskell
+ post
+    |> buildPost
+    >>= ifValid \case -- Changed |> to >>=
```
amitaibu, line 247: :+1:

Guide/validation.markdown
amitaibu: @stephenbenedict are you up for the challenge? :smile_cat:

stephenbenedict: When you say a "top level function" outside of the where clause, do you mean just like:

```haskell
-- Controller/Posts.hs
buildPost :: (?modelContext :: ModelContext, ?context :: ControllerContext) => Post -> IO Post
buildPost post = post
    |> fill @'["title", "body"]
    -- Add custom validator. The below example checks if the post’s title is unique.
    |> (\post -> validateFieldIO #title (titleIsUnique post) post) 
    -- Remove where clause

-- In Controller/Posts.hs or Helper/Controller.hs
titleIsUnique :: (?modelContext :: ModelContext) => Post -> Text -> IO ValidatorResult
titleIsUnique post title = do
   ...
```

amitaibu: The latter  😊 

AI could help you with that, if you'd like to try out. PRs are the best way to learn 💪

mpscholten: I meant `titleIsUnique` like in your example :+1:

amitaibu:
Oh 😅 think a generic unique would be a handy validation function

Gizra
https://www.gizra.com  | +972-54-4444986

Guide/validation.markdown
amitaibu: I think a generic unique would be a handy validation function.

But in order not to block @stephenbenedict first PR(!), let's move the `titleIsUnique` to be a top function. And I'll suggest the new function in a separate PR

Guide/validation.markdown
```diff
@@ -213,6 +213,46 @@ Then just call it like:
 user |> validateField #age isAge`
 ```
 
+### Creating a custom validator that uses IO
+
+Modify the `buildPost` function like:
```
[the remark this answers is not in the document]
stephenbenedict: Just committed edits to do that.

amitaibu (approved):
:ok_hand:

mpscholten (approved):
Thanks 👍

PR #2064, opened 2025-05-01
Updater docs on developing the DevServer

amitaibu:
@mpscholten I want to debug what's the value set on [mainInProjectDirectory](https://github.com/digitallyinduced/ihp/blob/012c7a3c03437109426c64b7c98189b6d639ca44/ihp-ide/exe/IHP/IDE/DevServer.hs#L49-L51)

```diff
mainInProjectDirectory :: FilePath -> IO ()
mainInProjectDirectory projectDir = do
    cwd <- Directory.getCurrentDirectory

+    liftIO $ putStrLn $ "mainInProjectDirectory: " <> cwd
+    liftIO $ putStrLn $ "projectDir: " <> projectDir
```

So I figured I need to run the DevServer via ghci. PR tries to fix the docs about it, but I get this error.

image

I tried to `nix flake lock --update-input ihp` from the host project, but it doesn't help

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -87,6 +87,7 @@ When making changes to the development tooling, we have to start the server diff
 use `make ghci` to load your application together with the framework located in `IHP`.
 
 ```
+make ghci
 :l IHP/ihp-ide/exe/IHP/IDE/DevServer.hs
 main
 ```
@@ -99,8 +100,8 @@ You can enable additional debug logging for the development server by setting th
 
 ```
 export DEBUG=1
-ghci
-:l ihp/ihp-ide/exe/IHP/IDE/DevServer.hs
+make ghci
+:l IHP/ihp-ide/exe/IHP/IDE/DevServer.hs
 main
 ```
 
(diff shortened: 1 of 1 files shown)


mpscholten:
The `mainInProjectDirectory` function assumes that you are starting the ghci from within the IHP directory itself. In your example above you are starting the ghci inside the project directory instead.

```
# Go to project with checked out IHP
cd my-project

# Go to IHP
cd IHP

ghci
:l ihp-ide/exe/IHP/IDE/DevServer.hs

# For this example it makes sense to use mainInParentDirectory, in a generic case it makes sense to use mainInProjectDirectory
mainInParentDirectory
```

amitaibu:
Go it working, and serving the static files from my local installation, a not from nix store.

I'll update docs here to use `mainInParentDirectory`. And seems we don't need `make ghci` -- `ghci` is enough.

@mpscholten Maybe it makes more sense in this doc to always point to the `path:///...`  URL, for contributors?

mpscholten:
yes we could do that. but it will increase the build time a bit as it will then need to recompile IHP locally. But likely the simplification is worth it 👍

mpscholten (approved):
Thanks 👍

PR #1386, opened 2022-03-08
Update Browser Location With Form Response URL

s0kil:
Fixes https://github.com/digitallyinduced/ihp/issues/1385

Issue #1385, opened 2022-03-08
When POST Submitting A Form, URL Is Not Updated


s0kil:
When submitting a form as POST, and the `create` action redirects to a new URL, the URL is not updated in the browser, so when refreshing, the form is rendered again.

Example:
GET /NewPost
Submit Form
POST /CreatePost
Action redirects to "/NewMessage"
/NewMessage is Rendered
Browser URL is still /CreatePost
So refresh causes `/CreatePost` to be rendered

s0kil:
Easiest way to see this in action is at "https://ihp.digitallyinduced.com/community/" do to a thread, create a new comment, it will take you back to thread, but the browser location is not updated, so refreshing cause new comment page to render.

PR #2066, opened 2025-05-05
Fix back button after form submit

### LINKED ISSUE #2063: Back button does not change contents after form submission + link click
This demonstrates the problem:

https://github.com/user-attachments/assets/666270b5-4213-4d13-9785-edf51fe8502f

ie. we do a submit, click a link, then try to go back. 

expected: contents of Users index (url being /Users)

got: same contents for /ShowUser and /Users

Code is at https://github.com/unhammer/ihp-back-button-repro – I just did `ihp-new`, created a users table with a name column, a users controller and ran it. Nothing out of the ordinary there.
That repo is using newest ihp master, but the problem was there at least as far back as 1.0.1 and I think earlier versions too.
Happens in both firefox and chrome.

A workaround is to use `formForWithoutJavascript`, but that loses the "interactive" feel (downgrades to web 1.0 🖳). I'd love to be able to have both the js form submission and a working back button.

Possibly related:
* https://github.com/digitallyinduced/ihp/issues/1893
* https://github.com/digitallyinduced/ihp/issues/451

amitaibu:
fixes #2063 

- Remove `initBack` which seems no longer used (?)
- Modernize `initDisableButtonsOnSubmit` to use a better way to prevent double binding

@unhammer could you try this PR locally to confirm it works for you?

### DIFF
--- a/lib/IHP/static/helpers.js
+++ b/lib/IHP/static/helpers.js
@@ -116,10 +116,11 @@ function initDelete() {
 }
 
 function initDisableButtonsOnSubmit() {
-    if (window.initDisableButtonsOnSubmitRun) {
+    if (Boolean(document.initDisableButtonsOnSubmit) === true) {
         return;
     }
-    window.initDisableButtonsOnSubmitRun = true;
+
+    document.initDisableButtonsOnSubmit = true;
 
     var lastClicked = null;
     document.addEventListener('submit', function (event) {
--- a/lib/IHP/static/helpers.js
+++ b/lib/IHP/static/helpers.js
@@ -8,7 +8,6 @@ window.addEventListener('beforeunload', function () {
 document.addEventListener('DOMContentLoaded', function () {
     initDelete();
     initDisableButtonsOnSubmit();
-    initBack();
     initToggle();
     initTime();
     initDatePicker();
@@ -19,7 +18,6 @@ document.addEventListener('DOMContentLoaded', function () {
 
 document.addEventListener('turbolinks:load', function () {
     initDelete();
-    initBack();
     initToggle();
     initTime();
     initFileUploadPreview();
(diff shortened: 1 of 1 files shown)


unhammer:
AFAICT this change has no effect on the issue.

From how I read the code, `initBack` only does something if you've manually added a `<button class="js-back">`. I can't find this `js-back` class documented anywhere (`git log --stat -S js-back` only shows changes to helpers.js) and I can find no open source usages on github https://github.com/search?q=js-back+hsx+language%3Ahaskell&type=code so it seems ripe for deprecation, but I also don't see how it can be related to my issue.

I tried adding some logging to see whether `initDisableButtonsOnSubmit` was being run twice, it doesn't seem like it was being run twice before either. In any case, the issue remains on my end.

Did the issue disappear in your testing with this patch?

amitaibu:
(My assumption is that after the form redirect, the new page isn't registered on the `history`)

amitaibu:
##  With Bug

https://github.com/user-attachments/assets/51f02fe3-a88e-475d-b202-6e67ca68c914 


## With PR 

https://github.com/user-attachments/assets/ace13b92-4aff-401b-8459-8b754825a3af

unhammer:
OMG this seems to actually work :-D

amitaibu:
> I wonder what the reason was for replaceState instead of pushState there in the first place

Hopefully @mpscholten will remember :smile:

lib/IHP/static/helpers.js
```diff
@@ -211,11 +186,11 @@ window.submitForm = function (form, possibleClickedButton) {
             );
             transitionToNewPage(request.response);
             Turbolinks.clearCache();
-            if (urlPathnameWithQuery !== formAction) {
+
+            if (urlPathnameWithQuery !== formAction || urlPathnameWithQuery !== responseUrlPath) {
```
amitaibu, line 190: This is the fix for the back button

amitaibu:
Ready for review

mpscholten:
> > I wonder what the reason was for replaceState instead of pushState there in the first place
> 
> Hopefully @mpscholten will remember 😄

Sadly I don't remember anymore :D but i did some research and the change was introduced via this PR: https://github.com/digitallyinduced/ihp/pull/1386

The related issue is https://github.com/digitallyinduced/ihp/issues/1385 Luckily the issue specifies a way to reproduce it:

```
GET /NewPost
Submit Form
POST /CreatePost
Action redirects to "/NewMessage"
/NewMessage is Rendered
Browser URL is still /CreatePost
So refresh causes /CreatePost to be rendered
```

@amitaibu could you try this patch against these reproduction steps?

unhammer:
https://github.com/digitallyinduced/ihp/pull/1386/files added replaceState two places, while this PR only removes it one place, so hopefully that's the right amount :-)

mpscholten:
Good point :D

amitaibu:
The PR fixed the code running when this is TRUE

https://github.com/digitallyinduced/ihp/blob/5903d2899c1abe751a3c7003ba8997ee63e57f21/lib/IHP/static/helpers.js#L182

When is it not TRUE, so I'll try to reproduce.

amitaibu:
Might be replaced by https://github.com/digitallyinduced/ihp/pull/2087

PR #1872, opened 2023-12-05
Emacs direnv

Montmorency:
Added a useful emacs config with direnv and lsp/lsp-ui. Addresses issue #1846.

PR #2072, opened 2025-05-14
Clean up emacs docs a bit

unhammer:
make it clearer that lsp(-ui) is an *alternative* to eglot

drop commented-out "flycheck but it does not work" bit and keep configs in the use-package blocks they belong in, and don't fail if smartparens/ligature/rainbow packages are not installed (they are not builtin)

drop reference to someone's long init-file
(though would like to know why it was there cf.
https://github.com/digitallyinduced/ihp/pull/1872#issuecomment-2878942116 )

note difference between M-x eglot and eglot-ensure hook


unhammer:
(though we'd be better off with vim =P)

mpscholten (approved):
👍 thanks!

Issue #2067, opened 2025-05-07
devenv up FileWatcher fails if there is a broken symlink in build/


unhammer:
Upgrading an IHP project from pre-flakes to 1.3 I had a broken symlink `build/ihp-lib` which blocked the auto-reload file watcher from picking up any changes. It seems reproducible still in 1.3 by just doing 

```
$ ( cd build && ln -s borken-symlink )
$ nix develop --impure
$ devenv up
```

and trying to change a file

Related: https://github.com/digitallyinduced/ihp/issues/1714

mpscholten:
I guess (but have not verified) it might be crashing here https://github.com/digitallyinduced/ihp/blob/master/ihp-ide/IHP/IDE/FileWatcher.hs#L87 The file watcher tries to recursively scan all directories and set up file watchers for each directory. Likely the "is directory" check crashes if the symlink is invalid.

unhammer:
I looked into this and the bug is actually in fsnotify. 

Some debug-prints on `startWatchingSubDirectory` shows it manages to watch `lib` and then tries to watch `build` and then hangs.

When I add a catch-all to the `FS.watchTree` call like

```diff
diff --git a/IHP/IDE/FileWatcher.hs b/IHP/IDE/FileWatcher.hs
index d380eb9d..8c60a050 100644
--- a/IHP/IDE/FileWatcher.hs
+++ b/IHP/IDE/FileWatcher.hs
@@ -12,6 +12,9 @@
 import qualified Data.List as List
 import IHP.IDE.LiveReloadNotificationServer (notifyAssetChange)
 import qualified Control.Debounce as Debounce
+import qualified Prelude
+import Control.Exception (catch)
+
 
 withFileWatcher :: (?context :: Context) => IO () -> IO ()
 withFileWatcher inner =
@@ -57,9 +60,10 @@ startWatchingSubDirectory
     watchedDirectories <- readMVar state
     case Map.lookup path watchedDirectories of
         Just _ -> pure ()
-        Nothing -> do
+        Nothing -> (do
             stop <- FS.watchTree manager path shouldActOnFileChange handleFileChange
             modifyMVar_ state (\map -> pure (Map.insert path stop map))
+            ) `catch` (\(e :: SomeException) -> Prelude.print e)
 
 stopWatchingSubDirectory :: FileWatcherState -> FilePath -> IO ()
 stopWatchingSubDirectory state path = do
@@ -90,6 +94,7 @@ fileWatcherConfig
[...]

PR #2076, opened 2025-05-20
use fsnotify 0.4.3.0

mpscholten:
relates to https://github.com/digitallyinduced/ihp/issues/2067

Initially I only overwrote the fsnotify version. But got interested in refactoring this part of the nix and and now replaced the whole mkGhcCompiler thing we had with normal nixpkgs overlays. This seems to be more idiomatic nix.


unhammer:
Seems to fix it :-)

PR #2077, opened 2025-05-22
psql-remote command

AronNovak:


lib/IHP/Makefile.dist
```diff
@@ -100,6 +100,11 @@ postgres: ## Starts the postgresql server
 psql: ## Connects to the running postgresql server
 	@psql -h $$PWD/build/db -d app
 
+psql-remote: ## Connects to the postgresql server of a deployed instance.
```
mpscholten, line 103: IMO we should try to stick this into the nix configuration as a devenv script, similarly how `deploy-to-nixos` is also a devenv script: https://github.com/digitallyinduced/ihp/blob/master/flake-module.nix#L273

Then it could be called by running `psql-remote ihp-app`

amitaibu: What's the advantage of this way over `make`?

mpscholten: As that command depends directly on the nix configuration (by calling nix directly) it feels like it should belong there

PR #2087, opened 2025-07-08
Migrate from Turbolinks 5 to Hotwire Turbo 8

amitaibu:
Migrates IHP from legacy Turbolinks 5.1.1 to modern [Hotwire Turbo](https://turbo.hotwired.dev/) 8.0.13, replacing morphdom with Turbo's built-in morphing.

  ## Changes
  • Replace Turbolinks with Turbo 8.0.13 bundle
  • Remove morphdom dependency - use Turbo's Idiomorph instead
  • Enable morphing with `<meta name="turbo-refresh-method" content="morph">`
  • Update auto-refresh to use Turbo Streams
  • Migrate `transitionToNewPage()` to `Turbo.renderPage()` API
  • Update events: `turbolinks:load` → `turbo:load`

  ## Benefits
  🚀 Better performance with Idiomorph vs morphdom
  📦 Single library instead of multiple dependencies
  🔧 Actively maintained vs discontinued Turbolinks

  ## Test plan
  - [x] IDE navigation works
  - [x] Auto-refresh preserves form state
  - [x] Live reload
  - [x] Handle `js-delete`
  - [ ] Server-side components functional
  - [x] IDE close modal dialogs
  - [x] IDE Delete button

  🤖 Generated with [Claude Code](https://claude.ai/code)


amitaibu:
This will require more testing, and likely would need a new release with upgrade notes (to update the JS files projects call on `Layout.hs`)

https://github.com/user-attachments/assets/2724336e-0250-4848-a386-8213d87e880d

But so far it's working :exploding_head: 

Seems to also fix https://github.com/digitallyinduced/ihp/issues/2063 that triggered this experiment :smile:

Guide/deployment.markdown
```diff
@@ -594,7 +594,6 @@ JS_FILES += ${IHP}/static/vendor/popper-2.11.6.min.js
 JS_FILES += ${IHP}/static/vendor/bootstrap.min.js
 JS_FILES += ${IHP}/static/vendor/flatpickr.js
 JS_FILES += ${IHP}/static/helpers.js
-JS_FILES += ${IHP}/static/vendor/morphdom-umd.min.js
 JS_FILES += ${IHP}/static/vendor/turbolinks.js
 JS_FILES += ${IHP}/static/vendor/turbolinksInstantClick.js
 JS_FILES += ${IHP}/static/vendor/turbolinksMorphdom.js
```
amitaibu:
```suggestion
JS_FILES += ${IHP}/static/vendor/turbo.js
```

amitaibu:
Auto refresh is also working.


https://github.com/user-attachments/assets/2b890bef-42f7-4035-93e6-5fda94cbe97f

amitaibu:
I'm trying to check the SSC following https://ihp.digitallyinduced.com/Guide/server-side-components.html#serverside-components

But I'm getting this error.

image

My `flake.nix` has `ihp.url = "path:///home/amitaibu/Sites/ihp-landing-page/IHP";`, and I've `nix flake update` and `direnv allow`

What else might I be missing?

p.s. Is anyone using SSC? Is this a feature we might want to drop at a certain point?

amitaibu:
@mpscholten this can already benefit from a review/ testing locally, as it's a big change.

amitaibu:
I've fixed ihp-scc not working, maybe it's not needed if running non-local IHP :shrug: 

```diff
haskellPackages = p: with p; [
    # Haskell dependencies go here
    p.ihp
+   p.ihp-ssc
```

amitaibu:
SSC is now included in the compile, but it complains about the code from the guide. So maybe someone that has ever used it, should give it a try :smile:

amitaibu:
@unhammer care to check your scenario with the [back button](https://github.com/digitallyinduced/ihp/issues/2063)? You can do it by changing in your `flake.nix`

1. `ihp.url = "github:digitallyinduced/ihp/pull/2087/head";` and `direnv reload`
2. And in `Layout.hs`

```diff
- <script src={assetPath "/vendor/morphdom-umd.min.js"}></script>
+ <script src={assetPath "/vendor/turbo.js"}></script>
- <script src={assetPath "/vendor/turbolinks.js"}></script>
- <script src={assetPath "/vendor/turbolinksInstantClick.js"}></script>
- <script src={assetPath "/vendor/turbolinksMorphdom.js"}></script>
```

Guide/recipes.markdown
```diff
@@ -436,15 +436,15 @@ Let's say you have a page with comments, and you link to them with `<a href="#co
...
 
 The solution is to assign your own class to the targeted element:
 
 ```javascript
 // app.js
 
-$(document).on('ready turbolinks:load', function () {
-    // Highlight the #hash target
+document.addEventListener('ihp:load', function () {
+    // Highlight the #hash target - works with or without Turbo
```
amitaibu:
```suggestion
    // Highlight the #hash target
```

unhammer:
I haven't noticed any big issues yet, so this seems very promising :-)

I did find one minor regression. I have a form that submits when I hit enter/tab to the next field. This worked fine in turbolinks, but in turbo it loses focus:

turbolinks, here I'm typing numbers and hitting enter, no mouse needed:
turbolinks

hotwire turbo, after I hit enter it does switch to the next field but then when the POST is complete and console says ihp:load runs, it loses focus:
turbo

unhammer:
I'm pretty sure it's because of this change to `ihp/data/static/helpers.js`, which to me seems like it sidesteps any morphing and just reverts to plain document.body.innerHTML = newHtml.body.innerHTML:


```diff
     var isModalOpen = document.body.classList.contains('modal-open');
-    morphdom(                                                                             
-        document.body,
-        newHtml.tagName === 'BODY' ? newHtml : newHtml.body,
-        {                                                                                                                                                                            
-            childrenOnly: false,                                                          
-            onBeforeElUpdated: function (from, to) {      
-                if (                  
-                    newHtml.body &&                                                       
-                    newHtml.body.classList.contains('modal-open') &&
-                    from.id === 'main-row'                                                
-                ) {
-                    return false;
-                } else if (isModalOpen && from.id === 'main-row') {
-                    return false;                                                         
-                } else if (              
-                    from.classList.contains('flatpickr-input') &&
-                    from._flatpickr
-                ) {                                                                       
-                    unsafeSetTimeout(
-                        function (from, to) {
-                            console.log(
-                                'FROM',
-                                from,
-                                to.getAttribute('value'),
-                                to.value                                                  
-                            );
-                            from.value = to.value;
-                            // from.setAttribute('value', to.getAttribute('value'));
-                        },                                                                
-                        0,                                                                
-                        from,
-                        to
-                    );         
-                }             
-            },                                                                            
-            getNodeKey: function (el) {                                                   
-                var key = el.id;  
-                if (el.id) {
-                    key = el.id;
-                } else if (el instanceof HTMLScriptElement) {
-                    key = el.src;
-                }
-                return key;
-            },                                                                                                                                                                       
-        }               
-    );                                                                                    
+    var newBodyHasModal = newHtml.body && newHtml.body.classList.contains('modal-open');
  
+    if (isModalOpen && !newBodyHasModal) {
+        // Modal is currently open but new content doesn't have modal - preserve modal state
+        // Only update non-modal content areas
+        var mainRow = document.getElementById('main-row');
+        if (mainRow && newHtml.body) {                                                    
+            var newMainRow = newHtml.body.querySelector('#main-row');                                                                                                                
+            if (newMainRow) {
+                mainRow.innerHTML = newMainRow.innerHTML;
+            }
+        }
+    } else {
+        // Normal page update - replace entire body content
+        if (newHtml.tagName === 'BODY') {
+            document.body.innerHTML = newHtml.innerHTML;
+        } else if (newHtml.body) {
+            document.body.innerHTML = newHtml.body.innerHTML;
+        }
+    }
```

So turbo doesn't get a chance to run any morphing here.

unhammer:
If I simply disable the `window.submitForm` override on `submit` events like 

```diff
--- ../ihp/ihp/data/static/helpers.js   2025-09-22 11:25:01.897260252 +0200
+++ static/helpers.js   2025-09-22 12:40:17.399831586 +0200
@@ -149,28 +149,20 @@
 function initDisableButtonsOnSubmit() {
     if (window.initDisableButtonsOnSubmitRun) {
         return;
     }
     window.initDisableButtonsOnSubmitRun = true;

-    var lastClicked = null;
-    document.addEventListener('submit', function (event) {
-        event.preventDefault();
-
-        var form = event.target;
-        window.submitForm(form, lastClicked);
-    });
-
     document.addEventListener('mouseup', function (event) {
         lastClicked = event.target;
     });
 }
```

then turbo.js seems to pick up the POST results and morph them on its own, no need for explicit calls to `morphdom`. I think *most* of what `window.submitForm` does is handled by the browser (and should be automatically handled by turbo morph), but it also does stuff like:
* pause/unpause live reload
* dismiss alerts
* handle `disableJavascriptSubmission` 
* in transitionToNewPage:
   *  dispatchEvent(ihpUnloadEvent) – I don't know what this is
   * only updating  `main-row`  if `.modal-open` (seems very special-cased? haven't seen any reference to those classes in IHP before now, but it's been there since 2020).

So one way forward would be to let turbo do the form submissions, and we hook into turbo's form submission events: https://turbo.hotwired.dev/handbook/drive#form-submissions where I guess we could do most of the stuff that window.submitForm currently does, and `data-turbo="false"` should be equivalent to data-disable-javascript-submission (though for backwards compat maybe helpers.js could go through and add data-turbo=false to all  data-disable-javascript-submission).

-----

Also, turbo.js seems to require an `id` attribute on forms to be able to morph them (maybe something to put in `formFor`, doesn't seem to matter what the id is or even that it's the same in the response, so maybe a random one would work)

amitaibu:
@unhammer it's very possible stuff are wrong. It was a session of me and Claude AI, and it's for sure not perfect.

I wouldn't mind if you want to take it from here :smile:

unhammer:
:+1: I don't think there's that much missing for it to be functional, apart from that window.submitForm (which I think is mostly about *removing* code).  I'll see if I find some time to make an attempt at it after Potato Holiday.

unhammer:
It took a bit longer to find a free moment than I expected. But I started working on this in https://github.com/unhammer/ihp/commits/hotwire-turbo/

In https://github.com/unhammer/ihp/commit/f4893c1b7fd03a645c43b478afdd457a9708a29d I made the changes to use turbo.js's form submission instead of `window.submitForm`, while keeping the functionality of `window.submitForm`:
* keeping backwards-compatibility with `data-disable-javascript-submission`
* dispatching the `ihp:load` event
* clearing intervals/timeouts
* dismissing `.alert`, and
* pausing autorefresh.

I also removed the `window.reload` on `popstate`, and from my testing so far it seems that whatever turbo.js does to handle forms/history is enough to fix #2063 .

----

There are some remaining challenges unfortunately:

**1: js-delete**

 `window.submitForm` was used both for regular form submissions and for turning `<a href="/Delete…" class="js-delete">`  into method=DELETE requests. For now I have *not* used turbo.js for the `js-delete` case, since turbo.js moves the search/query params into body params before submitting links as forms. So `/DeletePost?id=1&foo=bar` would be sent as a DELETE with `/DeletePost` and an empty query string and `id=1&foo=bar` in the request body. Their reasoning for this is given in https://github.com/hotwired/turbo/commit/f8b0e17a0c605080de0097112a5cf408abef9bb1 They refer to the HTML spec which of course does not mention DELETE since that's not a part of HTML forms at all, so I'm not 100% sure search params need to be moved in order for DELETE to follow "the html way". Maybe someone more knowledgeable about hypertext can chime in?  

In any case, I [stripped down](https://github.com/unhammer/ihp/commit/9d628c7cd7a8c147589c5e83099e1953a5d9d1ad) `window.submitForm` to what was actually needed for supporting `<a href="/Delete…" class="js-delete">`, though it would be nicer if we could drop it all. 

Possible ways forward:
- Keep using our submit code for `js-delete`  – no changes needed, it works, just feels a bit unsatisfying :)
- Switch to turbo for `js-delete`, just accept that search params are put in body, and say in `UPGRADE.md` that people need to use body params instead of query params for Delete actions – a breaking change
- Ask turbo.js if they could make it possible to have exceptions to their search param handling (though I don't feel confident that what they're doing is wrong)

**2: render 200 vs redirect on form submission**

A worse problem is this:
turbo.js Error: Form responses must redirect to another location
The basic IHP controller does 
```haskell
ifValid \case 
  Right post -> redirectTo PostsAction
  Left post -> render NewView{..}
```
but [Turbo requires either a 303 redirect or 4xx/5xx](https://turbo.hotwired.dev/handbook/drive#redirecting-after-a-form-submission) when rendering form submission responses. If we render a 200 to the same url, turbo.js fails with the above error (so the request gets sent, but response gets ignored). 

I guess it might be possible for `ifValid` to do something like `putContext (Status 422)`, and then `render` could inspect that (defaulting to 200). But there could be projects out there that use `render` on form submission actions without any `ifValid` – they would break, and require the same `putContext (Status 422)` (or switch to redirect). I have no idea how much people do this (though I guess it's kind of an unsafe thing to do anyway, if it can break reload?).

Anyone got any better ideas? I haven't yet looked into whether it's possible to coax turbo.js to be less strict here.

unhammer:
would have to check how this is affected by the https://github.com/digitallyinduced/ihp/pull/2196 fix (does turbo already handle it?)

unhammer:
@mpscholten I see there's been a lot of changes recently :-) I haven't been able to follow everythiing, but I'm guessing "challenge 2" above still remains? Ie. turbo requires form submissions to 303 redirect, while the default IHP way is to just 200 render.

mpscholten:
@unhammer I think we are pretty close to get this in a mergable state :) I just made 303 the default as that's the better default for our use cases anyways. 422 also works (we can just detect whether the current render is in response to a POST request)

unhammer:
Oh nice :-D

mpscholten:
I found a few migration regressions in this PR:

1. `window.morphPage()` appears to regress live reload for head changes.

Before this change, `livereload.js` morphed both `document.head` and `document.body`. After the Turbo migration, `livereload.js` now calls `window.morphPage(dom)`, but `morphPage()` only copies over the `ihp-auto-refresh-id` meta tag from `newHtml.head`.

That means changes to `<title>`, other meta tags, and head-managed assets no longer get applied during live reload, so layout/head edits won't fully show up without a full page reload.

Relevant spots:
- `ihp/data/static/livereload.js`
- `ihp/data/static/helpers.js`

2. The IDE layout no longer loads any script that defines `submitForm`, but the SQL REPL and migration editor still call it.

`helpers.js` no longer defines `window.submitForm`, and the new compatibility shim lives in `ihp/data/static/ihp-bc.js`. But `ihp-ide/IHP/IDE/ToolServer/Layout.hs` does not include that file, while these IDE scripts still call `submitForm()`:
- `ihp-ide/data/static/IDE/query-editor.js`
- `ihp-ide/data/static/IDE/migration-editor.js`

That means `Shift-Enter` / `Command-Enter` in those editors can now fail with `ReferenceError: submitForm is not defined`.

3. Generated app layouts also drop the documented `window.submitForm()` API.

`ihp-ide/IHP/IDE/CodeGen/ApplicationGenerator.hs` now emits `turbo.js`, `helpers.js`, `ihp-auto-refresh.js`, and `app.js`, but not the new compatibility shim. The branch still documents `window.submitForm(formElement)` in `Guide/form.markdown`, so upgraded apps that still rely on that helper can break at runtime.

4. There are still leftover `turbolinks:*` event listeners in the Turbo migration.

I found at least these:
- `ihp-ide/data/static/IDE/ihp-help.js` still listens for `turbolinks:load`
- `ihp-ide/IHP/IDE/Logs/View/Logs.hs` still listens for `turbolinks:load`

Turbo does not fire those events, so those initializers no longer run unless they are moved to `ihp:load` or `turbo:load`.

PR #2088, opened 2025-07-08
Fix Troubleshoot script

amitaibu:
- It didn't parse the nix output correctly
- It was looking for the `start` script
- It was looknig at `default.nix` instead of `flake.nix`

I'm still seeing two errors

image

1.  `direnv`  not allowed, but I did run `direnv allow`
image

2. `Symlink build/ihp-lib target directory does not exist. Try `nix-shell --run "make -B build/ihp-lib"` to fix this` -- maybe because I'm running with local IHP?


mpscholten:
The check for build/ihp-lib can be removed as this symlink is not used anymore in recent IHP versions

amitaibu:
Updated, and removed also the `hie.yml` check

image

mpscholten (approved):
Thanks 👍

PR #2090, opened 2025-07-10
Document Environment and the default differences between Prod/Dev

unhammer:
I was wondering what differences switching between the environments lead to, and ended up grepping the ihp repo. Maybe this helps others.

### DIFF
--- a/IHP/Environment.hs
+++ b/IHP/Environment.hs
@@ -3,7 +3,8 @@ import IHP.Prelude
 
 -- | The 'Environment' type is used to switch between development and production configurations.
 --
--- You can set it with 'option Development' or 'option Production' in your Config.hs, or via the 'IHP_ENV' environment variable.
+-- When running 'devenv up', this will be set to 'Development', while 'deploy-to-nixos' will set it to 'Production'.
+-- You can also manually set it with 'option Development' or 'option Production' in your Config.hs, or via the 'IHP_ENV' environment variable.
 --
 -- You can check the current environment using 'isDevelopment', 'isProduction', or 'isEnvironment' if you want to change behaviour based on the environment.
 -- IHP by default implements the following differences:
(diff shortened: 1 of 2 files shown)


IHP/Environment.hs
```diff
@@ -1,4 +1,16 @@
 module IHP.Environment where
 import IHP.Prelude
 
+-- | The 'Environment' type is used to switch between development and production configurations.
+--
+-- You can set it with 'option Development' or 'option Production' in your Config.hs, or via the 'IHP_ENV' environment variable.
```
amitaibu: You _can_, but you don't need to. When running `devenv up` it will set to DEV, and when you `deploy-to-nixos` it sets to Production.

I think we should clarify that, or maybe link to the docs

IHP/Environment.hs
```diff
@@ -1,4 +1,16 @@
...
+--
+-- You can check the current environment using 'isDevelopment', 'isProduction', or 'isEnvironment' if you want to change behaviour based on the environment.
+-- IHP by default implements the following differences:
+--
+-- - Static file caching: In 'Development', browser cache is disabled (max-age=0). In 'Production', max-age is forever, with asset hashes for invalidation. See 'initStaticApp'.
+-- - Logger: In 'Development', logging uses the default format. In 'Production', it uses an Apache-style logger and higher log level. See 'ihpDefaultConfig'.
+-- - Background workers: In 'Development', a development job worker server is started automatically. In 'Production', you need to manually start a separate RunJobs process. See the Guide for details.
```
amitaibu: Maybe link to docs

amitaibu (commented):
Thank, coupld of nitpicks

IHP/Environment.hs
unhammer: added clarification

IHP/Environment.hs
unhammer: added link

mpscholten:
Thanks! Looks good to me 👍 @amitaibu is this ready to merge from your side?

amitaibu (approved):
Thanks 🙏

PR #2091, opened 2025-07-16
Use Approot middleware for redirects

mpscholten:
Makes redirects more flexible when an IHP app is running on multiple domains

### DIFF
--- a/IHP/Controller/Redirect.hs
+++ b/IHP/Controller/Redirect.hs
@@ -18,6 +18,7 @@ import IHP.Controller.RequestContext
 import IHP.RouterSupport (HasPath (pathTo))
 import IHP.FrameworkConfig
 import Network.HTTP.Types.Status
+import qualified Network.Wai.Middleware.Approot as Approot
 
 import IHP.Controller.Context
 import IHP.ControllerSupport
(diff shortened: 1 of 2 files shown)


amitaibu:
Example of how this is used?

mpscholten:
Assuming we have an IHP blog app deployed on a server with an nginx load balancer in front of it. The IHP app and nginx is configured to serve two domains: `a.com` and `b.com`. The `IHP_BASEURL` env var is set to `https://a.com`.

Before this PR:
When you visit `https://b.com/NewBlogPost` and submit the form, it would redirect to `https://a.com/ShowBlogPost` instead of staying on the `b.com` domain. Because `IHP_BASEURL` is configured only to `a.com`.

Basically with `IHP_BASEURL` we don't really support multi-domain deployments of IHP apps.

After this PR:
When you visit `https://b.com/NewBlogPost` and submit the form, it will redirect to `https://b.com/ShowBlogPost`. `IHP_BASEURL` is not used for the url construction. Instead the current hostname is used directly.

(Other parts of IHP still rely on IHP_BASEURL, so multi-domain deployments are still non trivial, but atleast redirects work well now)

PR #2096, opened 2025-07-24
enable Approot middleware for the IDE

mpscholten:


amitaibu:
Do you know why the existing tests didn't catch the error?

mpscholten:
we don't have any tests for the IDE / devtools right now

mpscholten (commented):
Just finished my local IHP build and this seems to fix the error for tpp-app 👍

PR #2097, opened 2025-07-25
testing: call approotMiddleware

mpscholten:
@andreytroeglazov this should fix the error with the test suite 👍


andreytroeglazov:
@mpscholten thanks, can be merged? :slightly_smiling_face:

PR #2101, opened 2025-08-09
ihp-hsx: version 1.5.0

mpscholten:
@iteratee I'll push the lucid2 changes to hackage via version 1.5.0 in a few minutes.

Maybe we should refactor the package structure such that we have two hackage packages. Otherwise all users of ihp-hsx will always depend on lucid2 (e.g. splitting into ihp-hsx-parser, ihp-hsx-blaze and ihp-hsx-lucid2)

### DIFF
--- a/ihp-hsx/changelog.md
+++ b/ihp-hsx/changelog.md
@@ -1,5 +1,11 @@
 # Changelog for `ihp-hsx`
 
+## Version 1.5.0
+
+- Suport for GHC 9.10.x and 9.12.x
+- Add backend for lucid2 via dedidcated ihp-hsx-lucid2 library
+- Add missing attributes muted, onkeypress
+
 ## Versions < 0.18:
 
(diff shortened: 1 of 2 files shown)


iteratee:
> @iteratee I'll push the lucid2 changes to hackage via version 1.5.0 in a few minutes.
> 
> Maybe we should refactor the package structure such that we have two hackage packages. Otherwise all users of ihp-hsx will always depend on lucid2 (e.g. splitting into ihp-hsx-parser, ihp-hsx-blaze and ihp-hsx-lucid2)

It depends on the build tool. For cabal if you depend on `ihp-hsx:ihp-hsx-blaze` then it won't pull in `lucid2`. I occasionally submit patches to `rules_haskell` for `bazel`. It currently pulls in dependencies based on packages and not on particular components. That's a build tool limitation, it could be made to do the correct thing.

IHP uses nix, I don't know if the haskell tooling in nixpkgs currently supports this. It might be more expedient to split up the packages, but it's worth at least checking if nix supports this before doing it.

PR #2118, opened 2025-09-06
Doc: Change ihp.url back to default in postgres example

unhammer:
very minor documentation fix to avoid possible confusion (though could be useful on its own as an example of how to set to local ihp checkout?)


amitaibu (approved):
Thanks

PR #2120, opened 2025-09-08
Doc: 1.4 upgrade notes on ihp-migrate and nixpkgs package management

unhammer:


Guide/database-migrations.markdown
```diff
@@ -59,6 +68,17 @@ In a production context you need to specify the correct database URL via the `DA
 DATABASE_URL=postgresql://... migrate
 ```
 
+For running it when deploying from Github Actions, it may be faster to add a devshell containing just that package. Put
+
+```nix
+                devShells.ihp-migrate = pkgs.mkShell {
```
mpscholten: There's already a default package defined for the app flake called `migrate` https://github.com/digitallyinduced/ihp/blob/master/flake-module.nix#L183

So you can just do `nix build migrate` and then `./result/bin/migrate` 👍

mpscholten: Even better, with nix run we can do `nix run .#migrate`.

While testing this I found out that the binary is broken right now. The `${ghcCompiler.ihp}/bin/migrate` should be `${ghcCompiler.ihp-migrate}/bin/migrate`. Will do a PR to fix this

unhammer: I guess that requires an ihp 1.4.1 to be available?

mpscholten: Yes 👍 true

unhammer: How about in the guide we just change it to 

```diff
-For running it when deploying from Github Actions, it may be faster to add a devshell containing just that package. Put
-
-```nix
-                devShells.ihp-migrate = pkgs.mkShell {
-                  buildInputs = [
-                    pkgs.ghc.ihp-migrate
-                  ];
-                };
-
-```
-in `flake.nix` in the `perSystem` block (next to the `ihp` block). Then you can run it as `nix develop .#ihp-migrate --command migrate`.
+For running migrations when deploying from Github Actions, you can use `nix run .#migrate`.
```
since this is what will end up in 1.4.1, but maybe leave the UPGRADE.md as-is? (Or I could remove the devshell part from UPGRADE.md as well to keep it simple, since it just means slightly longer deploy-time until 1.4.1 comes.)

mpscholten: > since this is what will end up in 1.4.1, but maybe leave the UPGRADE.md as-is?

👍

unhammer: *pushed*

mpscholten (approved):

PR #2126, opened 2025-09-17
Doc: `nix flake metadata` needs `inputs-from .`

unhammer:
It seems like if you just do `nix flake metadata nixpkgs` it takes your global default nixpkgs or something, not the one from the local flake.nix. 

(My previous attempt seems like it gave the wrong hash, I didn't notice at first since it was cached I guess.)


mpscholten:
Confusing behaviour from nix :D Thanks!

mpscholten (approved):

PR #2130, opened 2025-09-19
Allow withIHPApp to return a value

unhammer:
handy for e.g. re-using Web calls in Admin tests

### DIFF
--- a/ihp-ide/IHP/Test/Database.hs
+++ b/ihp-ide/IHP/Test/Database.hs
@@ -33,7 +33,7 @@ import IHP.Test.Mocking
 withConnection databaseUrl = Exception.bracket (PG.connectPostgreSQL databaseUrl) PG.close
 
 -- | Create contexts that can be used for mocking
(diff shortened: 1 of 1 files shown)


amitaibu:
Can you give a code example of what you mean by Admin tests, and how this PR helps it?

unhammer:
I have two IHP apps, Web and Admin (as in https://ihp.digitallyinduced.com/Guide/architecture.html#next-to-my-main-web-application-im-also-building-an-admin-backend-application-where-to-place-it ); I've got an endpoint in Web that would be nice to call from the Admin test set since it sets up some stuff in the db. I was hoping I could do something like 
```haskell
tests = aroundAll (withIHPApp AdminApplication config) do
   ... 
   order <- withIHPApp WebApplication config $ withContext $ callActionWithParams ...
```

but on testing, it seems this actually doesn't make sense. I guess `withIHPApp` sets up a new temporary db and so the changes are gone outside that call, so I wouldn't be able to *use* the stuff it set up (in the example above, `orders` no longer refers to something that exists in the AdminApplication db context). And so it's probably safer to keep the `()` return value as-is; I'll close this.

PR #2141, opened 2025-10-16
fix run-script not working anymore

mpscholten:


Montmorency:
Great this sorts the environment issue by re-exporting the run-script :
Is this no longer required: 
```
echo ":def loadFromIHP \file -> (System.Environment.getEnv "IHP_LIB") >>= (\ihpLib -> readFile (ihpLib <> "/" <> file))"
echo ":loadFromIHP applicationGhciConfig"

```

and it looks like Config.hs loaded twice.

mpscholten:
It's not longer required as ghc implicitly loads the `.ghci` of the project which contains this line already 👍

mpscholten:
fixed the multiple loadings of the app now

PR #2143, opened 2025-10-17
change log and removed deprecated setting.

Montmorency:
Updated change log and remove a setting from app deployment config no longer required. i.e. WORKING_DIRECTORY does not need to change on system deployment.


ihp/CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 ## Minor Changes
 
 ### Nix / Deployment
+- [Removed option: 'WORKING_DIRECTORY' no longer required on deployment static assets wrapped automatically](https://github.com/digitallyinduced/ihp/commit/9c54588d983d74bdc28b063d1132952deb57dc16)
```
mpscholten: This should be added to a new `# unreleased` headline at the top as this was changed with a recent commit on master that is not yet released

PR #2145, opened 2025-10-22
Add --match option description to testing guide

andreytroeglazov:


Guide/testing.markdown
```diff
@@ -135,6 +135,14 @@ Another way of executing the tests, that we'll use on [CI](https://github.com/di
 runghc $(make print-ghc-extensions) -i. -ibuild -iConfig Test/Main.hs
 ```
 
+To execute a specific bunch of tests, you can use `--match`:
+
+```
+runghc $(make print-ghc-extensions) -i. -ibuild -iConfig Test/Main.hs --match "PostsController"
+```
+
+This command will execute all tests which are described under `describe "PostsController"`.
```
amitaibu: I'm assuming the match doesn't need to be exact? So something like this will work?

```suggestion
runghc $(make print-ghc-extensions) -i. -ibuild -iConfig Test/Main.hs --match "Posts"
```

This command will execute all tests which are described under `describe "Posts controller functionality"`.
```

File: Guide/testing.markdown [hunk 2]
```diff
@@ -135,6 +135,14 @@ Another way of executing the tests, that we'll use on [CI](https://github.com/di
 runghc $(make print-ghc-extensions) -i. -ibuild -iConfig Test/Main.hs
 ```
 
+To execute a specific bunch of tests, you can use `--match`:
```
amitaibu:
```suggestion
To run a particular set of tests, use `--match`.
```

Guide/testing.markdown
andreytroeglazov: yes, adjusted description, thanks

Guide/testing.markdown
andreytroeglazov: fixed

PR #2147, opened 2025-10-24
Contribute doc fix

Montmorency:
A couple tweaks to the contributing doc. I added a comment as when I went to patch the flake path using sed script the mac system sed fell over. A new contributor might not want to parse the sed script themselves so added a comment explaining it was patching the ihp.url input to refer to the path version.

Also added note saying suggesting to run `nix flake update` which I needed to do to get the app to ref the local ihp version (maybe it was still fetching according to the inputs in the lock file?).


mpscholten (approved):
Thanks 👍

PR #2149, opened 2025-10-31
Use vault for autorefresh

mpscholten:


amitaibu:
I've requested ChatGPT to provide a high-level explanation of the change. Usually, I skip reading its lengthy output, but this time, it's quite interesting.

@mpscholten did the AI explain it well? :smile: 

# Overview

Here’s the gist, in plain IHP terms.

```haskell
-- New shape: wire core services into the Request.vault and read them from there
run configBuilder = do
  withFrameworkConfig configBuilder \frameworkConfig ->
    withModelContext frameworkConfig \modelContext ->
      PGListener.withPGListener modelContext \pgListener -> do
        autoRefreshMw <- AutoRefresh.initAutoRefreshMiddleware pgListener
        Warp.runSettings settings $
             requestLogger
           . methodOverridePost
           . sessionMiddleware
           . approotMiddleware
           . autoRefreshMw
           . modelContextMiddleware modelContext
           . frameworkConfigMiddleware frameworkConfig
           . pgListenerMiddleware pgListener
           $ frontControllerToWAIApp @RootApplication @AutoRefresh.AutoRefreshWSApp id RootApplication staticApp
```

# High level summary

* **Bye `ApplicationContext`, hello vault.**
  The global `ApplicationContext` implicit parameter is removed. Instead, three core values now live in the WAI request vault and are retrieved via helpers in `IHP.RequestVault`:

  * `request.modelContext`
  * `request.frameworkConfig`
  * `request.pgListener`

* **AutoRefresh moves to per-request storage via vault.**
  `AutoRefreshServer` is created by a new middleware `AutoRefresh.initAutoRefreshMiddleware pgListener`. The server’s `IORef` is stored in the request vault. All AutoRefresh APIs now fetch it from the current `Request` instead of `?applicationContext`.

* **Controllers and websockets no longer require `?applicationContext`.**
  `ControllerSupport`, `RouterSupport`, `WebSocket`, and friends read what they need from `?context.requestContext.request`. You’ll see signatures drop `, ?applicationContext :: ApplicationContext`.

* **Error pages depend only on environment, not the removed context.**
  `ErrorController.renderError` and routing error handlers take an explicit `Environment` and pull other bits from the request.

* **IDE server and tests simplified.**
  `ToolServer` no longer spins up its own `PGListener` or `ApplicationContext`. Tests got lighter since many helpers don’t need to fake `ApplicationContext` or AutoRefresh state. Flash messages init now safely no-ops if session vault is missing (useful in tests).

* **New utilities for lifecycle.**
  `FrameworkConfig.withModelContext` and `ModelSupport.releaseModelContext` were added for clean acquire/release.
  `Controller.Session.lookupSessionVault` exposed, used to guard flash init.

# Why this is better

* **Stronger isolation.** Each request carries the exact resources it needs. No hidden globals, fewer spooky implicit params.
* **Easier to compose.** Mounting multiple apps or custom stacks is cleaner. Everything is standard WAI middleware.
* **More testable.** You can build minimal apps in tests without faking a giant context. Things that aren’t present just aren’t in the vault.
* **Clearer ownership.** `AutoRefreshServer` lifecycle is explicit and tied to the app middleware, not a global.

# What you need to change in your app or libs

1. **Remove uses of `?applicationContext`.**

   * Replace `?applicationContext.modelContext` with `request.modelContext`.
   * Replace `?applicationContext.frameworkConfig` with `request.frameworkConfig`.
   * Replace `?applicationContext.pgListener` with `request.pgListener`.

2. **Add the new middlewares, in this order (roughly):**

   * `frameworkConfigMiddleware frameworkConfig`
   * `modelContextMiddleware modelContext`
   * `pgListenerMiddleware pgListener`
   * `AutoRefresh.initAutoRefreshMiddleware pgListener`
   * Keep your session, CORS, logger, approot, and custom middleware as before.

3. **Adapt AutoRefresh calls.**

   * Drop `?applicationContext` constraints.
   * Where code previously did `fromContext @(IORef AutoRefreshServer)`, now call `autoRefreshServerFromRequest request`.

4. **Update tests and custom servers.**

   * If a test calls the WAI app directly, ensure you create the app via the same middleware stack or skip bits that aren’t needed.
   * Flash messages in tests work only if the session vault is present.

# Risks and breaking points to scan for

* Any module that **imports `IHP.ApplicationContext`** or expects `?applicationContext` will break. Migrate it to vault accessors.
* Any custom websocket app or `WSApp` implementation that referenced `?applicationContext` needs a tiny refactor to use `request`-backed access.
* If you forget to install `AutoRefresh.initAutoRefreshMiddleware`, you’ll hit the explicit runtime error about missing middleware. Good guardrail.

# Mental model

Think of IHP now as a standard WAI app that **sticks shared services into `Request.vault`** at the edge, then everything inside reads from `request`. It keeps the nice implicit `?context` ergonomics, but eliminates a big, app-wide implicit that was hard to reason about.

Opinionated take: this is the right direction. Less magic, more WAI. You’ll pay a small migration cost, but you get simpler composition and cleaner tests.

mpscholten:
The AI explains it very well 👍

PR #2155, opened 2025-11-18
Change IDE approot name

### LINKED ISSUE #2144: IHP v1.4.x adds port 8000 to redirect URL for POST
In previous versions of IHP (v1.3 and earlier), setting the URL using the environment variable worked properly, which for instance, allowed IHP to use Codespaces (`https://*.github.dev/Posts`) and allowed you to use a reverse proxy like Nginx. Which still mostly works, except now when we submit a POST request, by means of a `Create*` action for instance, the redirect URL is `https://*.github.dev:8000/Posts`. This happens even if we switch to `option Production`. This causes the pageload to fail and the site to appear to stall. But the `Create` action still works, and the post is inserted into the database in this case. This doesn't seem to affect `GET` requests using `pathTo` either, only when you use `redirectTo Action {..}`, and `render` also works. 

This feels like it would break for anyone who wants to run behind a reverse proxy of any kind, like we do in prod with Nginx. We're sticking with v1.3 for now because of this. I assume setting `PORT=443` may fix this, but this feels like an incorrect workaround. My understanding was that the `PORT` env var was supposed to specify the internal port only, and not affect what the user/browser sees.

CSchank:
This is a simple PR to hopefully fix #2144. I tried testing this on my macOS Apple Silicon laptop and it didn't build properly, complaining about a library being available in x86 instead of ARM, so worth looking into whether this is a problem more broadly with IHP. It worked on the x86-based Codespace I ran it on. 

This APPROOT bug is preventing us from being able to update to IHP v1.4.x, so I hope we can get this fix deployed soon :)

### DIFF
--- a/ihp-ide/IHP/IDE/ToolServer.hs
+++ b/ihp-ide/IHP/IDE/ToolServer.hs
@@ -73,7 +73,7 @@ startToolServer' toolServerApplication port isDebugMode liveReloadClients = do
     store <- fmap clientsessionStore (ClientSession.getKey "Config/client_session_key.aes")
     let sessionMiddleware :: Wai.Middleware = withSession store "SESSION" (frameworkConfig.sessionCookie) sessionVaultKey
 
-    approotMiddleware <- Approot.envFallback
+    approotMiddleware <- Approot.envFallbackNamed "IDE_APPROOT"
     
     staticApp <- initStaticApp
 


mpscholten (approved):
👍

mpscholten:
> I tried testing this on my macOS Apple Silicon laptop and it didn't build properly, complaining about a library being available in x86 instead of ARM, so worth looking into whether this is a problem more broadly with IHP.

This sounds like https://github.com/NixOS/nixpkgs/issues/462046 to me.

It was introduced a couple days ago via https://github.com/digitallyinduced/ihp/commit/c2b8a7ca321c7bdb606966c9f843365739428dc1

On my Mac it was caused by an old homebrew installation. I guess it was automatically copied over from my old Intel Mac when I set up the arm mac. I fixed this by deleting the library file that was mentioned in the nix error, and later deleted homebrew entirely. After that it worked great :)

Issue #2161, opened 2025-11-24
Running tests has changed with latest main


amitaibu:
<img width="748" height="209" alt="Image" src="https://github.com/user-attachments/assets/0ef1f67d-08a6-4660-9480-36fe1acfe0f1" />

I need to figure out how to execute the tests now and update the docs here
https://github.com/digitallyinduced/ihp/blob/37753b9f5bf1484715911284b18c14d0de49678e/Guide/testing.markdown?plain=1#L65

amitaibu:
## What changed

  - `withIHPApp` was removed from the core ihp package (`ihp/IHP/Test/Mocking.hs` no longer defines it). This happened in commit
    e681e9322ed4283d1af6a06f70e8eb995aa1bad5 (“Remove IHP.Libdir, moved IHP.Test.Database from ihp to ihp-ide”).
  - The helper now lives in the IDE package at `ihp-ide/IHP/Test/Database.hs` (still named `withIHPApp`).

@mpscholten I think we should bring it back?

amitaibu:
Fixed with https://github.com/digitallyinduced/ihp/pull/2163

PR #2162, opened 2025-11-24
Restore withIHPApp in core test helpers

amitaibu:
Fixes https://github.com/digitallyinduced/ihp/issues/2161

## Summary
- bring `withIHPApp` back into core `IHP.Test.Mocking` so tests don’t rely on `ihp-ide`
- bundle `IHPSchema.sql` with core and use it for test DB provisioning
- make the IDE helper delegate to the core implementation

### DIFF
--- a/ihp/IHP/Test/Mocking.hs
+++ b/ihp/IHP/Test/Mocking.hs
@@ -26,12 +26,14 @@ import           IHP.ModelSupport                          (createModelContext, 
 import           IHP.Prelude
 import           IHP.Log.Types
 import           IHP.Job.Types
+import qualified IHP.RequestVault                          as RequestVault
 import Test.Hspec
 import qualified Data.Text as Text
 import qualified Network.Wai as Wai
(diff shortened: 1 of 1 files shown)


amitaibu:
Tried one with AI, but it's not working, so need to look myself :sweat_smile:

amitaibu:
AI did it :smile: 
Confirmed with private project it's working

amitaibu:
@mpscholten ready for review

mpscholten:
I think it would be cleaner to refactor this function into a dedicated `ihp-tests` package (or a temporary workaround is to make your app depend on `ihp-ide` in `flake.nix`)

mpscholten:
@amitaibu what do you think about this https://github.com/digitallyinduced/ihp/pull/2163 ? With that PR you only need to add `ihp-hspec` to your project `flake.nix` as a dependency. This way we can also add more hspec specific helpers in the future to that module :)

amitaibu:
> what do you think about this https://github.com/digitallyinduced/ihp/pull/2163 ?

Great!

PR #2163, opened 2025-12-01
Move IHP.Test.Database to ihp-hspec package

mpscholten:


amitaibu:
Need to update docs accordingly

mpscholten:
Good point, just updated docs 👍

ihp-hspec/IHP/Test/Database.hs
```diff
@@ -0,0 +1,92 @@
+module IHP.Test.Database (withIHPApp) where
```
amitaibu: I'm not sure about the name. We're providing more than just a DB connection here. Maybe call `IHP.Test` or `IHP.Test.Mock`

Guide/testing.markdown
```diff
@@ -10,13 +10,15 @@ This section provides some guidelines for testing your IHP applications. It is h
 
 The following setup and tests can be viewed in the [Blog example](https://github.com/digitallyinduced/ihp-blog-example-app).
 
-1. Add `hspec` in `flake.nix`
+1. Add `hspec` and `ihp-hspec` in `flake.nix`
```
amitaibu, line 13: Maybe `ihp-hspec` can have a dependency on `hspec`, so we don't have to add both?

ihp-hspec/IHP/Test/Database.hs
mpscholten: Agree, initially I didn't want to rename this to avoid breaking existing apps, but we are breaking them anyways now, so it's a good time for changing it then :) I'd pick `IHP.Hspec`

Guide/testing.markdown
mpscholten: `hspec` would still be needed explicitly, otherwise ghc would complain that the package is a hidden package because it's not directly a dependency of the current app

ihp-hspec/IHP/Test/Database.hs
amitaibu: `IHP.Hspec` sounds right :+1:

ihp-hspec/ihp-hspec.cabal
```diff
@@ -0,0 +1,32 @@
...
+        TypeApplications
+        TemplateHaskell
+        QuasiQuotes
+        PackageImports
+        ScopedTypeVariables
+        BlockArguments
+        OverloadedStrings
+        ImplicitParams
+        RecordWildCards
+    build-depends: base >= 4.17.0 && < 4.22, ihp, wai, process, text, ihp-ide, vault, uuid, postgresql-simple
```
amitaibu, line 30: Not sure, but maybe:

```suggestion
    build-depends: base >= 4.17.0 && < 4.22, ihp, wai, process, text, hspec, vault, uuid, postgresql-simple
```

ihp-hspec/ihp-hspec.cabal
```diff
@@ -0,0 +1,32 @@
+cabal-version:       2.2
+name:                ihp-hspec
+version:             1.4.0
+synopsis:            Test helpers for IHP apps
+description:         Provides the withIHPApp function for hspec tests
+license:             MIT
+author:              digitally induced GmbH
+maintainer:          support@digitallyinduced.com
+bug-reports:         https://github.com/digitallyinduced/ihp/issues
+category:            Database
```
amitaibu: Should be `Test` category?

ihp-hspec/ihp-hspec.cabal
mpscholten: The `ihp-ide` dependency is required as it provides the `findIHPSchemaSql` function which returns the path to `IHPSchema.sql`. Right now we don't use anything from `hspec`, so a `hspec` dependency is not needed at the moment

ihp-hspec/ihp-hspec.cabal
mpscholten: Good catch 👍

PR #2169, opened 2025-12-07
Remove makefile targets

mpscholten:


amitaibu:
@mpscholten Meanwhile, also co-pilot created a draft [here](https://github.com/digitallyinduced/ihp/pull/2168) :smile:

PR #2197, opened 2025-12-17
Making log window colorful

ZichaoNickFox:
For making the log window easy to read or distinguishing log types. I added a functionality to give users the chance to print logs in color. Such as.

image

The call from users would like
```haskell
logDebug :: (?context :: context, LoggingProvider context) => Text -> IO ()
logDebug = Log.debug
logError :: (?context :: context, LoggingProvider context) => Text -> IO ()
logError = Log.error
logInfo :: (?context :: context, LoggingProvider context) => Text -> IO ()
logInfo = Log.info
logWarn :: (?context :: context, LoggingProvider context) => Text -> IO ()
logWarn = Log.warn

logController :: (?context :: context, LoggingProvider context) => LogLevel -> Text -> IO ()
logController level log = do
  logInfo $ "\\ESC[31m" <> log <> "\\ESC[0m"
```

Compiling `hsx` will discard `\ESC`. So I use fake `\\ESC` to keep `\ESC` to the frontend. Then, before `ansiUp.ansi_to_html` converts the string into HTML, I replace `\\ESC` back to `\ESC` in JavaScript.
```javascript
    const ESC = "\u001b";
    const text = pre.textContent.replace(/\\ESC\[/g, ESC + "[");
    const html = ansiUp.ansi_to_html(text);
```

`ansiUp` will insert `<span>` by `\ESC[` keywords, which is the principle of this implementation.
image
 
`type="module"` of `<script type="module" src={assetPath "/vendor/ansi-up.min.js"}></script>` is a trade-off. I cannot figure out another method for `ansi-up`, because it has `export` even in `ansi-up.min.js`. AI says it is a `ES6` standard.

### DIFF
--- a/Guide/database.markdown
+++ b/Guide/database.markdown
@@ -853,3 +853,21 @@ CREATE TABLE users (
     UNIQUE (email, username)
 );
 ```
+
+PostgreSQL constraint names have a **63-byte** limit.
+
+For a multi-column unique constraint, Postgres auto-generates a name from table + column names.  
+For `strategy_factor_regime_online` + `(symbol_id, timeframe, ts, version)`, this becomes a long name (truncated form: `strategy_factor_regime_online_symbol_id_timeframe_ts_version_ke`) and can trigger noisy migrations like:
+
+```sql
+ALTER TABLE strategy_factor_regime_online DROP CONSTRAINT strategy_factor_regime_online_symbol_id_timeframe_ts_version_ke;
+```
+
+Solution: use a short explicit constraint name after `CREATE TABLE strategy_factor_regime_online`:
+
+```sql
+ALTER TABLE strategy_factor_regime_online
+    ADD CONSTRAINT uq_sfr_on_main UNIQUE (symbol_id, timeframe, ts, version);
+```
+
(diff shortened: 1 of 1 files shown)


unhammer:
shouldn't all those [37m…[0m in your screenshots be turned into span color's too then?

ZichaoNickFox:
Other `[37m…[0m` was printed in other places, which use `\ESC` instead of `\\ESC`. Unlike mine `logInfo $ "\\ESC[31m" <> log <> "\\ESC[0m"`.

So `\ESC`s are discarded when converting from `hsx` to `html` in https://github.com/digitallyinduced/ihp/blob/b6c60a58eabe53017ee719d2783fc0d4c2812e4d/ihp-ide/IHP/IDE/Logs/View/Logs.hs#L7

To solve the problem, we should not discard `\ESC` in the `<pre>` element of `<pre>{standardOutput}</pre>` when compiling `hsx`.

I will confirm if the converting discards `\ESC`s or `ByteString` already discards `\ESC`.
`\\ESC` is the fake one to keep raw text into the frontend.

PR #2199, opened 2026-01-09
redirect see other for auth

vcombey:
## What / Why

This PR adds `redirectTo*SeeOther` helpers (HTTP **303**) and uses them in the auth Sessions controller.

With **302** after a non-GET (ex: `DELETE` logout), some clients (notably in HTMX flows) can follow the redirect using the **same method**, meaning the redirect target gets hit with **DELETE** instead of **GET**. With **303 See Other**, the client must follow the redirect with a **GET**, which fixes logout/login redirects.

## HTMX impact

Before, using `hx-delete` for logout could cause the redirect step to also be a DELETE, so you needed a form + `_method=DELETE` workaround.

Now this works cleanly:

```diff
- <form method="POST" action={pathTo DeleteSessionAction} hx-boost="false">
-   <input type="hidden" name="_method" value="DELETE" />
-   <button type="submit" class="flex items-center gap-3 text-error">
-     {iconPower}
-     <span>Log out</span>
-   </button>
- </form>
+ <button class="flex items-center gap-3 text-error"
+         hx-delete={pathTo DeleteSessionAction}>
+   {iconPower}
+   <span>Log out</span>
+ </button>


mpscholten (approved):

mpscholten:
Thanks 👍

PR #2202, opened 2026-01-18
Add PostgreSQL 18 support with UUIDv7 as default

mpscholten:
PostgreSQL 18 introduces native uuidv7() and uuidv4() functions, eliminating the need for the uuid-ossp extension. UUIDv7 provides time-ordered UUIDs that are better for database indexing and reduce B-tree fragmentation compared to random UUIDv4.

Changes:
- Remove uuid-ossp extension from IHPSchema.sql
- Use uuidv7() as default for new tables in Schema Designer
- Update code generators to use uuidv7()
- Add uuidv7() and uuidv4() to recognized SQL functions
- Maintain backward compatibility for parsing uuid_generate_v4()
- Update all documentation and tests

### DIFF
--- a/ihp-ide/IHP/IDE/CodeGen/DefaultUuidFunction.hs
+++ b/ihp-ide/IHP/IDE/CodeGen/DefaultUuidFunction.hs
@@ -1,14 +1,3 @@
 module IHP.IDE.CodeGen.DefaultUuidFunction (defaultUuidFunction) where
 
-import IHP.Prelude
-import qualified System.Environment as Env
-import Text.Read (readMaybe)
-
--- | Returns the default UUID function based on the @IHP_POSTGRES_VERSION@ env var.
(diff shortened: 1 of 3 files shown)


avitkauskas:
`export IHP_POSTGRES_VERSION=18` in `.envrc`
`direnv allow`
`nix flake update ihp`
`direnv reload`
`devenv up`
But still running postgres 17.7 and uuidv7() is not found.
Still too early? Something else is needed?

mpscholten:
we still need to update nixpkgs or more specially the postgres package to use v18

avitkauskas:
Is that doable in the project flake.nix or should that be done on the IHP framework side?

mpscholten:
it can be done locally by setting `services.postgres.package = pkgs.postgresql_18;` in the devenv section of flake.nix

Issue #2148, opened 2025-10-29
Make IHP more haskell-y


mpscholten:
A while ago Tom Ellis brought this up via email to the haskell foundation:

```
Dear board,

A few months I came across a post on Reddit that presented an
interesting perspective. It said that

the Haskell community feels boring, full of "why so serious" people,
Clojure/Kotlin/Elixir communities or many other PL communities have
a totally different vibe, more "vibrant"

https://old.reddit.com/r/haskell/comments/1kygyf3/a_break_from_programming_languages/mv6csf6/

I asked for more details and got a very interesting, detailed, response here:

https://old.reddit.com/r/haskell/comments/1kygyf3/a_break_from_programming_languages/mvjjt10/

I think the response is worth reading in full, but to give a taster,
it says that there seem to be fewer "new stuff" announcements for
Haskell than Clojure, the Haskell Interlude is more "bland" (that word
is my interpretatation) than the Clojure podcast, and that "Servant's
website is meh, stuck in time" whereas IHP "doesn't align with what
the haskell community values".

I don't have a particular proposed response, nor necessarily a wish to
start a discussion, but I thought it was an interesting point and I
can see where it's coming from, so I wanted to share.

Tom
```

We should fix this feedback and make IHP more aligned with normal Haskell - while still keeping the IHP vibes.

The original mentioned reddit comment is very good  https://old.reddit.com/r/haskell/comments/1kygyf3/a_break_from_programming_languages/mvjjt10/

amitaibu:
> We should fix this feedback and make IHP more aligned with normal Haskell - while still keeping the IHP vibes.

Do you have examples of this? A significant advantage for IHP is that I was able to get non-Haskell Gizra devs (but with some Elm knowledge) to start working productively on IHP from day one, so we should be careful with introducing too advanced concepts 😄

unhammer:
On the positive side, IHP is mentioned as Haskell's answer to Phoenix :-) 

Also if the community "feels boring", do we really want to align with that? 😛 I find IHP quite fun, and love the focus on features and developer experience over "just" safety/abstraction. So IHP certainly aligns with my values, and I think it would be good to play to IHP's strengths, perhaps taking some inspiration from how Phoenix is marketed and developed.

[...]

PR #2205, opened 2026-01-20
Refactor response handling to avoid exceptions on normal requests

mpscholten:
Previously, every render/redirect threw a ResponseException which was caught by runAction. Now render/redirect call WAI's respond directly and return IO ResponseReceived.

For early exits (accessDeniedUnless, notFoundWhen, basicAuth), a new earlyReturn helper throws EarlyReturnException after the response is sent.

This should improve performance since exceptions are no longer thrown on every request - only for guard-based early exits.

Relates to https://github.com/digitallyinduced/ihp/issues/2148

### DIFF
--- a/Guide/controller.markdown
+++ b/Guide/controller.markdown
@@ -408,7 +408,7 @@ action ExampleAction = do
 
 ### Early Return
 
-When you need to exit an action early (e.g. for access control), use `earlyReturn`. This sends the response and stops execution of the rest of the action:
+When you need to exit an action early, use `earlyReturn`. This sends the response and stops execution of the rest of the action:
 
 ```haskell
 action ExampleAction = do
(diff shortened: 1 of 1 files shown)


mpscholten:
## Local Benchmark Results

5000 requests, 10 concurrent connections, persistent connections (`hey`), ihp-forum app.

### HTTP Latency

| Metric | Baseline (master) | PR | Change |
|--------|-------------------|-----|--------|
| Average | 2.8 ms | 2.9 ms | +3.6% |
| p50 | 3.0 ms | 3.0 ms | 0% |
| p99 | 4.9 ms | 4.6 ms | **-6.1%** |
| Max | 17.2 ms | 10.2 ms | **-40.7%** |
| Requests/sec | 3,442 | 3,412 | -0.9% |

### Core Size

| Metric | Baseline (master) | PR | Change |
|--------|-------------------|-----|--------|
| Core size | 16,414,170 bytes | 14,733,504 bytes | **-10.2%** |

### Summary

No throughput or median latency regression. Tail latency (p99, max) improved — the old exception-based response handling occasionally caused GC pressure spikes from stack unwinding on every request. The new direct-return approach is more predictable. Compiled code is 10% smaller.

Guide/controller.markdown
```diff
@@ -398,19 +398,28 @@ action ExampleAction = do
...
 ```
 
-The [`putStrLn`](https://ihp.digitallyinduced.com/api-docs/IHP-Prelude.html#v:putStrLn) will never be called because the [`redirectTo`](https://ihp.digitallyinduced.com/api-docs/IHP-Controller-Redirect.html#v:redirectTo) already stops execution.
+### Early Return
+
+When you need to exit an action early (e.g. for access control), use `earlyReturn`. This sends the response and stops execution of the rest of the action:
+
+```haskell
+action ExampleAction = do
+    when (not loggedIn) do
```
amitaibu, line 415: Maybe we should mention `accessDeniedUnless` function here?

PR #2206, opened 2026-01-20
Disable profiling and haddock for faster local builds

mpscholten:
nixpkgs enables profiling by default for Haskell packages. This disables both profiling and haddock generation for IHP packages to speed up local development builds.

- Add fastBuild helper in overlay.nix that applies disableLibraryProfiling and dontHaddock
- Add ihp-with-docs for reference documentation builds that need haddock
- Apply same optimizations to models package in default.nix

### DIFF
--- a/NixSupport/default.nix
+++ b/NixSupport/default.nix
@@ -109,7 +109,7 @@ CABAL_EOF
     };
 
     # Build the models package as a proper Haskell package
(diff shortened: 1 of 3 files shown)


amitaibu:
PR is failing on `master`

https://github.com/digitallyinduced/ihp/actions/runs/21171950854/job/60890653634

image

PR #2219, opened 2026-01-24
Add integration test for ToolServer requestBodyMiddleware

mpscholten:
## Summary

Adds an integration test that verifies the ToolServer middleware stack correctly includes `requestBodyMiddleware`, which is required for controllers to read form params.

This test will catch regressions like the one fixed in #2217 / 1029fe34 where removing `requestBodyMiddleware` caused `lookupRequestVault: Could not find RequestBody`.

## Test approach

The test:
1. Builds a middleware stack mirroring `ToolServer.hs`
2. Makes real HTTP POST requests with form data via `Network.Wai.Test`
3. Verifies the parsed body is available in the request vault
4. Includes a negative test showing what happens when the middleware is missing

## Test plan

- [x] All 347 tests pass including the 3 new middleware tests

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/ihp-ide/IHP/IDE/ToolServer.hs
+++ b/ihp-ide/IHP/IDE/ToolServer.hs
@@ -1,4 +1,4 @@
-module IHP.IDE.ToolServer (runToolServer) where
+module IHP.IDE.ToolServer (runToolServer, buildToolServerApplication, ToolServerApplicationWithConfig(..)) where
 
 import IHP.Prelude
 import qualified Network.Wai as Wai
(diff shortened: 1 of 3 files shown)


ihp-ide/IHP/IDE/ToolServer/Types.hs
```diff
@@ -15,6 +15,16 @@ data ToolServerApplication = ToolServerApplication
         , databaseNeedsMigration :: !(IORef Bool)
         }
 
+-- | Create a new ToolServerApplication with empty IORefs
+newToolServerApplication :: PortNumber -> IO ToolServerApplication
```
amitaibu: Maybe move this to ihp-ide/Test/IDE/ToolServer/MiddlewareSpec.hs

amitaibu:
@mpscholten Gave this a look, but I admit I don't fully understand everything - so your call :smile:

PR #2230, opened 2026-01-25
Update and rename CLAUDE.md to AGENTS.md

amitaibu:
Makes this work also for non-claude AI agents


mpscholten:
Claude doesn't work with AGENTS.md. Maybe we can use a symlink to workaround this...

I also would not recommend using anything else other than Claude Code right now 😄

PR #2232, opened 2026-01-26
Fix flash messages and improve test middleware stack

mpscholten:
## Summary
- Fix "consumeFlashMessagesMiddleware was not called" error by passing the current request (with `flashVaultKey` in vault) to `renderHtml`
- Improve test mocking to run the full middleware stack for each request, matching production behavior
- Preserve mock sessions from `withUser` when running through middleware stack

## Problem
`renderHtml` was using `frozenContext.request` which didn't have `flashVaultKey` in the vault. The flash messages middleware adds `flashVaultKey` to the request it passes to `next`, but this modified request wasn't being used when rendering views.

## Test plan
- [x] Production app no longer shows "consumeFlashMessagesMiddleware was not called" error
- [x] Tests run with full middleware stack like production
- [x] `withUser` mock sessions work correctly

### DIFF
--- a/ihp/IHP/ControllerSupport.hs
+++ b/ihp/IHP/ControllerSupport.hs
@@ -56,6 +56,7 @@ import qualified Network.WebSockets as WebSockets
 import qualified IHP.WebSocket as WebSockets
 import qualified Data.TMap as TypeMap
 import IHP.RequestVault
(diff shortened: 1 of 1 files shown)


mpscholten:
@amitaibu this should fix the tests in tpp-app

PR #2236, opened 2026-01-29
Add customRoutes and customPathTo to AutoRoute

mpscholten:
## Summary

- Adds two new methods to the `AutoRoute` typeclass: `customRoutes` and `customPathTo`, allowing users to override the URL for individual actions without reimplementing the entire `CanRoute` and `HasPath` instances
- The default `CanRoute` instance now tries `customRoutes` before falling back to `autoRoute`
- The default `HasPath` instance now checks `customPathTo` before falling back to auto-generated paths
- Fully backwards compatible: defaults (`empty` and `Nothing`) inline away to zero runtime cost when not used

## Test plan

- [x] All 37 existing routing tests pass (no regressions)
- [x] 6 new tests verify custom route parsing, auto-generated fallback, and custom path generation
- [x] Module compiles cleanly

### DIFF
--- a/Guide/routing.markdown
+++ b/Guide/routing.markdown
@@ -172,6 +172,42 @@ When using multiple applications in your IHP project, e.g. having an admin back-
 
 This prefixing has special handling for the `Web` module so that all controllers in the default `Web` module don't have a prefix.
 
+## Overriding Individual AutoRoute Actions
+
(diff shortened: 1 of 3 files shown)


amitaibu (approved):
Very handy!

PR #2255, opened 2026-01-29
Add compileRelationSupport flag to SchemaCompiler

mpscholten:
## Summary

- Add `compileRelationSupport` flag to `CompilerOptions` that, when disabled via `IHP_RELATION_SUPPORT=0`, generates simpler types without `Include`/`fetchRelated` type machinery
- Removes type parameters from record types, omits has-many `QueryBuilder` fields, and skips generating `Include` modules for large-schema compile time improvements
- Add documentation in the Relationships guide and Troubleshooting guide

## What changes for the user

Set `IHP_RELATION_SUPPORT=0` before building to opt into simpler generated types:

```bash
export IHP_RELATION_SUPPORT=0
```

**Default output** (for a `posts` table with `user_id` FK and `comments` referencing it):
```haskell
data Post' userId comments = Post {
    id :: (Id' "posts"), title :: Text, userId :: userId,
    comments :: comments, meta :: MetaBag
}
type Post = Post' (Id' "users") (QueryBuilder.QueryBuilder "comments")
```

**Simple mode** (`IHP_RELATION_SUPPORT=0`):
```haskell
data Post' = Post {
    id :: (Id' "posts"), title :: Text, userId :: (Id' "users"),
    meta :: MetaBag
}
type Post = Post'
```

This trades `fetchRelated`/`Include` support for faster compilation, which is useful for projects with large schemas that prefer explicit queries.

## Test plan

- [x] All 359 existing tests pass unchanged
- [x] Two new tests verify simple mode output (no type params, no QB fields, simplified Record instance)
- [ ] Manual verification: build a project with `IHP_RELATION_SUPPORT=0` and confirm generated types are simplified


Guide/relationships.markdown
```diff
@@ -505,3 +505,77 @@ html PostsView { .. } = [hsx|
...
+
+IHP generates type-level machinery to support `fetchRelated` and `Include`: type parameters on record types, `QueryBuilder` fields for has-many relationships, and `Include` type family instances. For large schemas this adds significant compile time.
+
+If your project does not use `fetchRelated` or `Include` types, you can disable this machinery by setting the `IHP_RELATION_SUPPORT` environment variable to `0` before building:
+
+```bash
+export IHP_RELATION_SUPPORT=0
+```
+
+Add this to your `.envrc` or `.env` file to apply it persistently during development.
```
amitaibu, line 519: Could we make this automatic? That is, if `fetchRelated` or `Include` are used, then use a different compilation?

Or, have a breaking change, and remove `fetchRelated` and `Include` entirely in favor of the simpler yet more robust [multiple nested records](https://ihp.digitallyinduced.com/Guide/relationships.html#multiple-nested-records)?

mpscholten: I'd not like to make this a breaking change in IHP v1 as we have some projects making have use of this. But now when the AI is writing all the code anyways we don't need these helpers (and they are very costly for the compiler)

Happy to remove it in IHP v2.

Regarding automatic: I prefer to keep this explicit otherwise a developer could accidentally use a `fetchRelated` and then toggle this on again.

(The option moved from the env var to a nix flag btw https://github.com/digitallyinduced/ihp/commit/2813e3f021545faa86eb12f75e3ff651febbb0dd as otherwise it's hard to deploy)

PR #2256, opened 2026-01-30
AI assisted upgrade notes

amitaibu:
Adds an “AI Assisted Upgrade” section to UPGRADE.md, including the recommended
  update commands and a copy/paste prompt for using an AI agent to fix compile
  errors.


UPGRADE.md
```diff
@@ -31,6 +31,35 @@ The welcome page controller has been moved to its own package `ihp-welcome`. If
...
+```
+I ran devenv up 2>&1 | tee /tmp/devenv.log. Keep tailing /tmp/devenv.log and fix compile errors iteratively until the app builds.
+  Rules:
+
+  - Don’t touch anything under IHP/ (IHP core).
+  - Always tail -n 120 /tmp/devenv.log after each change.
+  - Fix only the app code (Web/, Application/, Config/, etc.).
+  - If a fix requires new deps/config, update project files accordingly (e.g., add p.ihp-mail in flake.nix when IHP.Mail imports fail) and tell me if I need to restart
+    devenv up.
+  - Keep going until the log shows no compile errors.
```
mpscholten, line 209: Maybe we can add another point for the AI to look up the actual git history in https://github.com/digitallyinduced/ihp to understand what has changed. This way it does less guessing and can get a better understanding of changes

amitaibu: Sure, I'll change that.

(Although, on a custom app, I've used this prompt, it fixed it well, without it)

I'm now also trying to fix my compilation errors in the test suite. However, I see codex can't run `runghc $(make print-ghc-extensions) -i. -ibuild -iConfig Test/Main.hs` itself.

@mpscholten were you able to make claude run the tests itself? how?

mpscholten: I often tell it load the test modules in ghci and run it. Because then it automatically uses the `.ghci` config sets all the required flags. It also likes to run `nix flakes check --impure` but that's very slow

amitaibu: Can you please share the command that is needed for that

mpscholten:
```
direnv exec . ghci -e ':l Test/Main.hs' -e 'main' 2>&1
```

(In my CLAUDE.md I have a line that tells it to use `direnv exec . $CMD` for all commands, such that it always loads the ihp dev env)

amitaibu: Cool. Is your CLAUDE.md different from the one https://github.com/digitallyinduced/ihp/blob/2813e3f021545faa86eb12f75e3ff651febbb0dd/CLAUDE.md ?

If so, maybe worth updating that one?

mpscholten: Hm yes I'll update it

Montmorency:
Claude can also curl localhost:8000 :)

amitaibu:
All my projects have `process.manager.implementation = "overmind";` So I realize that on an out-of-the-box project the `devenv up 2>&1 | tee /tmp/devenv.log` won't work

amitaibu:
So we could ask AI to look at the logs created by devenv

`tail -f .devenv/state/process-compose/process-compose.log`

amitaibu:
@copilot fix based on https://github.com/digitallyinduced/ihp/pull/2256#issuecomment-3831626755

PR #2257, opened 2026-01-30
Fix deployment guide to match ihp-boilerplate structure

### LINKED ISSUE #2037: Seeming discrepancies between flake.nix in guide and what is generated by default
There seem to be discrepancies between the `flake.nix` shown in the guide under [Configuring the Instance](https://ihp.digitallyinduced.com/Guide/deployment.html#configuring-the-instance) (under _Deploying with deploy-to-nixos_) and what is generated by default with `ihp-new`.

For example, the default `flake.nix` contains this in the `flake-parts.lib.mkFlake` block:

```
services.nginx = {
                            virtualHosts."CHANGE-ME.com" =  {
                                # Uncomment to have http auth with username `foo` and password `bar`.
                                # basicAuth = { foo = "bar"; };
                            };
                        };
```

but in the guide, you are instructed: "Make sure you put this into the flake-parts.lib.mkFlake block. The final flake.nix should look like this:" with a config that lacks the above nginx config. Maybe it was excluded for brevity but it is slightly confusing when first modifying `flake.nix` for NixOS deployment since the guide seems to imply that that whole chunk of `flake.nix` can be replaced with what is shown in the guide.

mpscholten:
## Summary
- Rewrites the "Configuring the Instance" section of the deployment guide to describe the modular `Config/nix/hosts/production/` structure that `ihp-new` actually generates, replacing the outdated inline `flake.nix` approach
- Updates SSH config, `deploy-to-nixos` command, and all related references to use `production` instead of `ihp-app`
- Fixes stale references to `flake.nix` in the RDS, CloudWatch, and troubleshooting sections to point to `Config/nix/hosts/production/configuration.nix`

## Test plan
- [ ] Read through the updated guide and verify it matches the current [ihp-boilerplate](https://github.com/digitallyinduced/ihp-boilerplate) file structure
- [ ] Verify `deploy-to-nixos production` matches the `nixosConfigurations."production"` key in the boilerplate's `flake.nix`
- [ ] Confirm the `hardware-configuration.nix` instructions (uncomment AWS AMI import) match the boilerplate file

Fixes #2037


Guide/deployment.markdown
```diff
@@ -190,133 +190,87 @@ After you've created the instance, configure your local SSH settings to point to
...
-+                            fixtures = ./Application/Fixtures.sql;
-+                            sessionSecret = "xxx";
-+                        };
-+
-+                        # Job workers are active by default. Disable them like this:
-+                        # systemd.services.worker.enable = pkgs.lib.mkForce false;
-+
-+                        # Add swap to avoid running out of memory during builds
-+                        # Useful if your server have less than 4GB memory
-+                        swapDevices = [ { device = "/swapfile"; size = 8192; } ];
```
amitaibu, line 296: Don't we need it anymore?

mpscholten: for small servers I think its still recommended

PR #2270, opened 2026-02-01
Makefile: Use $DATABASE_URL instead of hardcoded build/db socket path

mpscholten:
## Summary

- Updated `make db`, `make psql`, `make dumpdb`, and `make sql_dump` targets to use `$DATABASE_URL` instead of the hardcoded `-h $PWD/build/db` socket path
- Updated `make postgres` to inform users that postgres is now managed by devenv
- Removed `rm -rf build/db` from `make clean` since that directory no longer exists

Postgres is now managed by devenv, which exposes `$DATABASE_URL` via nix flakes. The old `build/db` socket path no longer exists, causing all database-related make targets to fail with "Connection refused" errors.

Fixes https://github.com/amitaibu/ihp-eheza/issues/71

## Test plan

- [ ] Run `make db` with devenv postgres running and verify schema + fixtures are loaded
- [ ] Run `make psql` and verify it connects to the devenv-managed database
- [ ] Run `make dumpdb` and verify it dumps fixtures correctly
- [ ] Run `make sql_dump` and verify output

### DIFF
--- a/ihp-ide/data/lib/IHP/Makefile.dist
+++ b/ihp-ide/data/lib/IHP/Makefile.dist
@@ -7,6 +7,9 @@
 MAKEFLAGS += --warn-undefined-variables
 MAKEFLAGS += --no-builtin-rules
 
+PSQL_ARGS = "$$DATABASE_URL"
+PG_DUMP_ARGS = "$$DATABASE_URL"
+
 GHC_EXTENSIONS=
 GHC_EXTENSIONS+= -XGHC2021
 GHC_EXTENSIONS+= -XOverloadedStrings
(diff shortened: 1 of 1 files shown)


amitaibu:
Thanks. I'm wondering if we can have some test that would confirm it's working well?

mpscholten:
investigating this

mpscholten:
Added regression tests in b8c427dc that verify:

- `Makefile.dist` has no references to the old `build/db` socket path
- `PSQL_ARGS` and `PG_DUMP_ARGS` are defined using `$DATABASE_URL`

All 363 IDE tests pass including the 4 new ones.

mpscholten:
the regression test is too complicated for now. we have to do that properly in a different pr

PR #2271, opened 2026-02-01
Fix Schema Designer column names invisible on dark background

mpscholten:
## Summary
- Bootstrap's `text-body` class on column name `<a>` tags overrides inherited white text color with near-black `#212529`, making column names invisible against the dark `#063642` background
- Bootstrap's `--bs-table-bg` CSS variable defaults to white, giving table rows a white background that breaks the dark theme

## Test plan
- [ ] Open the Schema Designer and select a table
- [ ] Verify column names are visible (white text on dark background)
- [ ] Verify table rows have a transparent/dark background instead of white
- [ ] Verify hover highlighting still works on column rows

### DIFF
--- a/ihp-ide/data/static/IDE/schema-designer.css
+++ b/ihp-ide/data/static/IDE/schema-designer.css
@@ -206,6 +206,11 @@
     background-color: #063642;
 }
 
+.column-selector .table {
+    --bs-table-bg: transparent;
+    --bs-table-hover-bg: hsla(192, 81%, 26%, 0.25);
+}
+
 .object {
     color: #E5EAEB;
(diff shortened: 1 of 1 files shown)


ihp-ide/data/lib/IHP/Makefile.dist
```diff
@@ -7,6 +7,9 @@
 MAKEFLAGS += --warn-undefined-variables
 MAKEFLAGS += --no-builtin-rules
 
+PSQL_ARGS = "$$DATABASE_URL"
```
amitaibu, line 10: Related to another PR

PR #2295, opened 2026-02-05
Fix AutoRefresh and DataSync breaking after `make db`

mpscholten:
## Summary

- Fixes AutoRefresh and DataSync stopping after `make db` recreates the database
- Both systems cached which tables had triggers installed, but after `make db` the triggers were gone while caches still thought they existed
- Adds `connectionEpoch` to PGListener that increments on reconnect
- Adds `runOncePerConnection` helper that auto-clears caches when epoch changes

## Test plan

- [ ] CI passes
- [ ] @amitaibu tests with https://github.com/amitaibu/ihp-sensors to verify the fix

Fixes https://github.com/amitaibu/ihp-sensors/pull/2#issuecomment-3853136095

### DIFF
--- a/Guide/database.markdown
+++ b/Guide/database.markdown
@@ -32,7 +32,9 @@ Database name: `app`.
 
 Once you have created your project, the first step is to define a database schema. The database schema is a SQL file with a lot of `CREATE TABLE ...` statements. You can find it at `Application/Schema.sql`.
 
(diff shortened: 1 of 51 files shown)


mpscholten:
@amitaibu Could you please test this fix with your ihp-sensors project after CI passes? This should resolve the AutoRefresh issue you reported where it stops working after `make db`.

Issue #1284, opened 2021-12-20
Add a quasi-quoter for type checked SQL queries


mpscholten:
For complex sql queries it would be very useful to be able to write code like this:

```sql
action MyAction = do
    result <- [query|select id, title from users|]
```

Specifically more complex queries that do aggregations.

The queries should be type checked against the `Schema.sql` at compile-time.

PR #2304, opened 2026-02-06
Typed sql

vcombey:
Limit: from what i understood, the strategy using prepare statement in sql is not enough to infer the nullability of LEFT JOIN / RIGHT JOIN. so any queries with those would put all fields in maybe..
We will see when I add tests.

### DIFF
--- a/ihp/default.nix
+++ b/ihp/default.nix
@@ -5,16 +5,18 @@
 , data-default, deepseq, directory, fast-logger, filepath, ghc-prim
 , hashable, haskell-src-exts, haskell-src-meta, hasql
 , hasql-dynamic-statements, hasql-implicits, hasql-mapping
-, hasql-pool, hasql-postgresql-types, hasql-transaction, hspec
-, http-client, http-client-tls, http-media, http-types, ihp-context
-, ihp-hsx, ihp-imagemagick, ihp-log, ihp-modal, ihp-pagehead
-, ihp-pglistener, inflections, interpolate, ip, lens, lib
-, mime-types, minio-hs, mono-traversable, mtl, neat-interpolation
-, network, network-uri, parser-combinators, postgresql-simple
-, postgresql-simple-postgresql-types, postgresql-types, process
-, pwstore-fast, random, random-strings, regex-tdfa, resource-pool
-, resourcet, safe-exceptions, scientific, slugger, split, stm
-, string-conversions, template-haskell, temporary, text, time
+, hasql-pool, hasql-postgresql-types, hasql-transaction, hspec, http-client
+, http-client-tls, http-media, http-types, ihp-context, ihp-hsx
+, ihp-imagemagick, ihp-log, ihp-modal, ihp-pagehead, ihp-pglistener
+, inflections, interpolate, lens
+, lib, mime-types, minio-hs, mono-traversable, mtl
+, neat-interpolation, network, network-uri, parser-combinators
+, postgresql-simple, postgresql-simple-postgresql-types, postgresql-types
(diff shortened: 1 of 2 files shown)


mpscholten:
Fixes https://github.com/digitallyinduced/ihp/issues/1284

vcombey:
should i rebase it to switch-to-hasql branch ? is the migration to hasql finished. I honestly doesn't understand anything about hasql. Postgres-simple was so much simpler. 
Are users suposed to migrate from postgres-simple to hasql ?
Is it to have better type check ? or for the performance ?
then could this typedsql quasiquoter become the user facing default for doing sql queries along with ihp "orm" ?
@mpscholten

mpscholten:
likely we'll not merge the switch-to-hasql branch branch. rebasing with master should already bring in all the hasql stuff. Current ihp master has a hybrid postgresql-simple and hasql way for now. Dropping pg simple would break too much, so we just use hasql wherever possible and stick to pg simple for the rest.

hasql is much faster (2 - 3x) because it uses prepared statements and the binary protocol of Postgres (pg simple is textual). I think it's in fact faster than the normal postgres c bindings.

its most complex but the majority of code will be written by ai agents in the future anyways, so we can accept this additional complexity.

the user facing parts of the ihp arm (createRecord, updateRecord, deleteRecord, query |> fetch) will not change, so for most users nothing changes except that it's 2x faster.

> then could this typedsql quasiquoter become the user facing default for doing sql queries along with ihp "orm" ?

yes 👍 I think together with the other crud functions. especially since it hides all the complexity of hasql. also llms are very good at raw sql, so this should be a good addition for the agentic future.

ihp-typed-sql/IHP/TypedSql/Bootstrap.hs
```diff
@@ -0,0 +1,183 @@
+module IHP.TypedSql.Bootstrap
```
mpscholten: Is it possible to move this to its own `ihp-typed-sql` package? If that works, we should likely to it to keep compile times of IHP core small

ihp-typed-sql/IHP/TypedSql/ParamHints.hs
```diff
@@ -0,0 +1,309 @@
...
+    { phIndex  :: !Int
+    , phTable  :: !Text
+    , phColumn :: !Text
+    , phArray  :: !Bool
+    }
+    deriving (Eq, Show)
+
+-- | Tokenize SQL just enough to locate placeholders and nearby identifiers.
+-- This feeds alias detection and parameter hint extraction in typedSql.
+tokenizeSql :: String -> [SqlToken]
```
mpscholten: Maybe we could make use of https://github.com/nikita-volkov/postgresql-syntax with is used by https://github.com/nikita-volkov/hasql-th likely it already contains the parsing logic we need

vcombey:
I have no understanding of what you are doing, but thanks for pushing this !

Maybe we can also track table automatically for autorefresh with the new macro. 
(before it needed user indication when pgquery was used)

Do you think left and right join nullability inferance is solvable with (/postgresql-syntax](https://github.com/nikita-volkov/postgresql-syntax) ? what's your take on that @mpscholten ?

mpscholten:
The changes were mostly moving the modules into their own package and then refactoring a bit to simplify the code.

> Maybe we can also track table automatically for autorefresh with the new macro.
(before it needed user indication when pgquery was used)

Good point, that should work actually

> Do you think left and right join nullability inferance is solvable

Not sure, I'm often using subqueries instead of joins :D Maybe claude can figure it out

PR #2309, opened 2026-02-06
Auto-reset hasql pool on stale prepared statements after make db

mpscholten:
## Summary
- After `make db`, the hasql connection pool retains connections with stale cached prepared statements from the old schema
- PostgreSQL returns error `0A000` ("cached plan must not change result type") on every query, requiring a dev server restart
- This adds auto-detection and recovery: on this specific error, the pool is reset via `HasqlPool.release` (which keeps the pool usable) and the query is retried once with a fresh connection

Fixes the issue reported in https://github.com/amitaibu/ihp-sensors/pull/2#issuecomment-3860911336

## Test plan
- @amitaibu Can you test this branch with your ihp-sensors project? After `make db`, requests should auto-recover instead of showing the "cached plan must not change result type" error.

### DIFF
--- a/ihp-datasync/IHP/DataSync/ChangeNotifications.hs
+++ b/ihp-datasync/IHP/DataSync/ChangeNotifications.hs
@@ -154,11 +154,11 @@ createNotificationFunction uuidFunction table = Snippet.sql [i|
 
 -- Statements
 
-retrieveChangesStatement :: Statement.Statement UUID ByteString
+retrieveChangesStatement :: Statement.Statement UUID Text
 retrieveChangesStatement = Statement.preparable
     "SELECT payload FROM large_pg_notifications WHERE id = $1 LIMIT 1"
     (Encoders.param (Encoders.nonNullable Encoders.uuid))
-    (Decoders.singleRow (Decoders.column (Decoders.nonNullable Decoders.bytea)))
+    (Decoders.singleRow (Decoders.column (Decoders.nonNullable Decoders.text)))
 
 -- Sessions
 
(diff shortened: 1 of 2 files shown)


amitaibu:
@mpscholten I've tried this branch but I'm getting a compile error on `ihp-sensors` repo

image

amitaibu:
@mpscholten app now compiles, thanks! However, `autoRefresh` still doesn't work after `make db`. I'm actually not able to see it working now also right after `devenv up`

amitaibu:
(It also seems that Jobs are not executed)

mpscholten:
@amitaibu I've added resilient error recovery so that Job workers, the job poller, and AutoRefresh survive database errors after `make db`. Could you test again with this latest commit?

Changes:
- **Job worker loop**: `fetchNextJob` is now wrapped in `tryAny` — on failure it logs the error, waits 1s, and retries instead of dying
- **Job poller**: The polling query is wrapped in `tryAny` — on failure it logs and continues (the existing jitter+delay provides backoff)
- **`runSessionHasql`**: Added the same `isCachedPlanError` retry+pool-reset as `sqlQueryHasql`/`sqlExecHasql` (this covers trigger creation in `watchForJob`)
- **AutoRefresh re-render loop**: Added an outer `SomeException` catch so DB errors don't silently kill the async thread

amitaibu:
PR doesn't compile

image

mpscholten:
Sorry, fixing it now

amitaibu:
After `make db`, I get

image


Also, @mpscholten, if you'd run my app (you have access to it), you will see that now the returned HTML from autoRefresh is taking over the entire page, instead of replacing the needed element. Maybe something with the `defaultLayout` ?

image

mpscholten:
I think the error on your screenshot is related to an older IHP version. I updated the branch in flake.nix and now it works fine. Can you try again (after pulling the ihp-sensors-test branch)

amitaibu:
Eralier I was using a local IHP. Anyway, switched to a remote, but now getting

image

amitaibu:
I've checked with https://github.com/digitallyinduced/ihp/pull/2309/commits/210f11a46f93af598cc45243207a25f61b297dea and still no luck - autoRefresh doesn't work after `make db`

I can check tomorrow the latest merge, although I don't think there were PRs dealing with that area.

mpscholten:
ok thanks 👍 so far the project runs well for me again. auto refresh works as well. but haven't tested the make db yet

amitaibu:
Yeah, it works only on the first go. After `make db` no autoRefresh

mpscholten:
@amitaibu works now after make db :)

PR #2314, opened 2026-02-06
Add integration test for IHP's public API and withIHPApp

### LINKED ISSUE #2183: Add test coverage for `withIHPApp`
> I think we should introduce a test that would use `withIHPApp`, so we have test coverage for it. 

 _Originally posted by @amitaibu in [#2173](https://github.com/digitallyinduced/ihp/issues/2173#issuecomment-3641940743)_

## Task

- Create a test that uses `withIHPApp` to verify that other projects running the command function correctly.

mpscholten:
## Summary

- Adds `integration-test/` — a minimal but representative IHP app (3 tables, CRUD controller, HSX views) that compiles and runs as part of `nix flake check`
- Tests `withIHPApp` from `ihp-hspec` with a real PostgreSQL database, verifying record CRUD and `callActionWithParams`
- Fixes operator precedence bug in schema compiler's `hasqlCreateManyBody` where `$` binding lower than `<>` caused `RETURNING` clause to be appended to a list instead of the SQL snippet

## What this catches

The existing `ihp-boilerplate` CI check has an empty schema and zero controllers/models/views. This integration test exercises:
- Schema compiler code generation (enums, FKs, nullable columns, TEXT[], INT, BOOLEAN, TIMESTAMP WITH TIME ZONE)
- Controllers (`query`, `fetch`, `newRecord`, `createRecord`, `updateRecord`, `deleteRecord`, `fetchRelated`, `fill`, `validateField`, `ifValid`, `redirectTo`, `render`)
- Views with HSX (`forEach`, `formFor`, `textField`, `textareaField`, `submitButton`, `Include "comments" Post`)
- `withIHPApp` test infrastructure and `callActionWithParams`

Closes #2183

## Test plan

- [x] `nix build .#checks.aarch64-darwin.integration-test` passes (4 examples, 0 failures)
- [ ] `nix flake check --impure` passes on CI

### DIFF
--- a/Guide/server-side-components.markdown
+++ b/Guide/server-side-components.markdown
@@ -125,13 +125,11 @@ instance View WelcomeView where
     html WelcomeView = [hsx|
         <h1>Counter</h1>
 
-        {counter}
+        {component @Counter}
     |]
-        where
-            counter = component @Counter
 ```
 
-If you wonder why we're using the `where` instead of writing [`{component @Counter}`](https://ihp.digitallyinduced.com/api-docs/IHP-ServerSideComponent-ViewFunctions.html#v:component): Currently the at-symbol `@` is not supported in HSX expressions.
+If this does not compile in your project, make sure `TypeApplications` is enabled for the module (e.g. add `{-# LANGUAGE TypeApplications #-}` at the top). In current IHP versions, HSX splices support type applications like `@Counter`.
 
 We also need to load the `ihp-ssc.js` from our `Layout.hs`. Open the `Web/View/Layout.hs` and add `<script src="/vendor/ihp-ssc.js"></script>` inside your `scripts` section:
 
--- a/ihp-hsx/Test/IHP/HSX/QQSpec.hs
+++ b/ihp-hsx/Test/IHP/HSX/QQSpec.hs
@@ -40,6 +40,9 @@ tests = do
             let project = Project { name = "Testproject" }
             [hsx|<h1>Project: {project.name}</h1>|] `shouldBeSameHtml` "<h1>Project: Testproject</h1>"
 
+        it "should support TypeApplications inside spliced expressions" do
+            [hsx|<h1>{show (id @Int 1)}</h1>|] `shouldBeSameHtml` "<h1>1</h1>"
+
         it "should support lambdas and pattern matching on constructors" do
             let placeData = PlaceId "Punches Cross"
             [hsx|<h1>{(\(PlaceId x) -> x)(placeData)}</h1>|] `shouldBeSameHtml` "<h1>Punches Cross</h1>"


amitaibu:
Great addition! Should we also include a `Job`? Also maybe ihp-mail as well? 

For email, we can use mailhog's API to verify. Here's a snippet from a project:

```haskell
it "should not send emails to comment author" $ withContext do
    -- Delete any previous email in Mailhog.
    Wreq.delete "http://0.0.0.0:8825/api/v1/messages"
    (user, lmName) <- createUserAndLmName

    -- Create a comment
    lmComment <-
        newRecord @LmComment
            |> set #lmNameId lmName.id
            |> set #body "This is a comment on this step"
            |> createRecord

    -- Assert email was sent, and caught by Mailhog.
    documentBody <- do
        -- Get emails.
        response <- Wreq.get "http://0.0.0.0:8825/api/v1/messages"
        pure (response ^. Wreq.responseBody)

    cs documentBody `shouldNotContain` "This is a comment on this step"
```

mpscholten:
Thanks @amitaibu! Added a Job test — `UpdatePostViewsJob` that increments a post's view count, exercising the `Job` typeclass, `perform`, and `callJob` test helper. Now at 5 tests (CRUD + controller action + job).

For ihp-mail with Mailhog — I think that's a great idea but would add meaningful complexity to the nix derivation (running a Mailhog service alongside PostgreSQL in the sandbox). Happy to add it as a follow-up PR. Would you prefer it in this PR or separate?

amitaibu:
We can have a separate one. This one is already covering a lot, so let's not block it.

mpscholten:
claude is on it already :D

integration-test/Test/IntegrationSpec.hs
```diff
@@ -102,11 +105,34 @@ tests = around (withIHPApp WebApplication testConfig) do
...
+                Nothing -> expectationFailure "No job found in queue"
+                Just lockedJob -> do
+                    lockedJob.status `shouldBe` JobStatusRunning
+
+                    -- Step 2: perform — execute the job logic
+                    let ?context = (?mocking).frameworkConfig
+                    perform lockedJob
+
+                    -- Step 3: jobDidSucceed — marks job as Succeeded in DB
+                    jobDidSucceed lockedJob
```
amitaibu, line 130: That's really nice!
I'm wondering if we could also have an integration testing that would test that a job is picked up just like it does when running `devenv up`? (doesn't need to be part of this PR)

mpscholten: Yes that is what I initially wanted but claude didn't found a good way yet 👍 let's improve this later on in another PR

PR #2321, opened 2026-02-06
docs(hsx): use {component @Counter} in SSC guide

mpscholten:
The SSC guide currently claims that HSX splices cannot use type applications (e.g. {component @Counter}).

This PR updates Guide/server-side-components.markdown to use {component @Counter} directly and adds a regression test to ihp-hsx to ensure TypeApplications inside HSX splices work (e.g. {show (id @Int 1)}).

If users still hit compile errors, they likely need to enable the TypeApplications extension in their module.

### DIFF
--- a/Guide/server-side-components.markdown
+++ b/Guide/server-side-components.markdown
@@ -125,13 +125,11 @@ instance View WelcomeView where
     html WelcomeView = [hsx|
         <h1>Counter</h1>
 
-        {counter}
+        {component @Counter}
     |]
-        where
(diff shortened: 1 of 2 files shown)


mpscholten:
CI is green (nix-flake-check). This is a small docs+test change to allow HSX splices to use TypeApplications (e.g. {component @Counter}).\n\nAuto-merge is enabled (squash) once the required approval is in. Could a maintainer please take a quick look/approve?

PR #2344, opened 2026-02-08
fix schema order parsing

vcombey:
## Title
Fix schema parser to accept PostgreSQL column constraint ordering (`NOT NULL DEFAULT ...`)

## Summary
The schema parser rejected valid PostgreSQL column definitions when constraints were written as `NOT NULL DEFAULT ...` (or other mixed orderings), because `parseColumn` expected a fixed order.

This PR makes column constraint parsing order-independent while preserving existing AST output semantics.

## What Changed
- Refactored column parsing to collect column attributes in any order, then fold into the `Column` record.
- Supported order-independent parsing for:
  - `DEFAULT`
  - `GENERATED ALWAYS AS ... [STORED]`
  - `PRIMARY KEY`
  - `NOT NULL`
  - `UNIQUE`
- Added duplicate-modifier validation to match PostgreSQL behavior more closely:
  - Rejects repeated `DEFAULT`, `GENERATED`, `PRIMARY KEY`, `NOT NULL`, `UNIQUE` on the same column.

## Regression Tests Added
- Parses `BOOLEAN NOT NULL DEFAULT false`.
- Parses mixed ordering like `UUID PRIMARY KEY DEFAULT uuid_generate_v4() NOT NULL`.
- Fails on duplicate `DEFAULT`.
- Fails on duplicate `NOT NULL`.

### DIFF
--- a/ihp-ide/Test/IDE/SchemaDesigner/ParserSpec.hs
+++ b/ihp-ide/Test/IDE/SchemaDesigner/ParserSpec.hs
@@ -537,14 +537,6 @@ tests = do
                     , primaryKeyConstraint = PrimaryKeyConstraint ["id"]
                     }
 
-        it "should fail to parse a column with multiple DEFAULT clauses" do
-            (evaluate (parseSql "CREATE TABLE tasks (is_completed BOOLEAN DEFAULT false DEFAULT true);")) `shouldThrow` anyException
-            pure ()
-
-        it "should fail to parse a column with multiple NOT NULL clauses" do
-            (evaluate (parseSql "CREATE TABLE tasks (is_completed BOOLEAN NOT NULL NOT NULL);")) `shouldThrow` anyException
-            pure ()
-
         it "should parse a CREATE TABLE statement with an array column" do
             parseSql "CREATE TABLE array_tests (\n    pay_by_quarter integer[]\n);\n" `shouldBe` StatementCreateTable (table "array_tests")
                     { columns = [ col "pay_by_quarter" (PArray PInt) ]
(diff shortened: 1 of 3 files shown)


mpscholten:
Thanks for the PR! The fix is correct — PostgreSQL does accept column constraints in any order, so the parser should too.

One suggestion: the implementation could be simplified by using a recursive accumulator instead of introducing the intermediate `ParsedColumnAttribute` data type + list comprehension extraction + duplicate counting. Something like:

```haskell
parseColumn :: Parser (Bool, Column)
parseColumn = do
    name <- identifier
    columnType <- sqlType
    space
    let col = Column { name, columnType, defaultValue = Nothing, notNull = False, isUnique = False, generator = Nothing }
    go col False
    where
        go col pk = choice
            [ try (lexeme "DEFAULT" >> expression) >>= \v -> go col { defaultValue = Just v } pk
            , try (lexeme "GENERATED" >> lexeme "ALWAYS" >> lexeme "AS" >> do
                generate <- expression
                stored <- isJust <$> optional (lexeme "STORED")
                go col { generator = Just ColumnGenerator { generate, stored } } pk)
            , try (lexeme "PRIMARY" >> lexeme "KEY") >> go col True
            , try (lexeme "NOT" >> lexeme "NULL") >> go col { notNull = True } pk
            , try (lexeme "UNIQUE") >> go col { isUnique = True } pk
            , pure (pk, col)
            ]
```

This eliminates the `ParsedColumnAttribute` sum type, the list comprehensions, the `length` counting for duplicate validation, and the custom `listToMaybe` redefinition.

The duplicate-modifier detection could be dropped — PostgreSQL itself rejects those, so the parser doesn't need to replicate that validation.

vcombey:
like that ?

mpscholten (approved):
Thanks looks good to me

PR #2345, opened 2026-02-08
Fix autoRefreshMeta: read AutoRefreshState from vault

mpscholten:
## Summary

- `autoRefreshMeta` used `fromFrozenContext` to read `AutoRefreshState`, but nothing ever inserted it into the controller context TMap — the state was wrapped in an `IORef` inside the request vault, invisible to views
- Remove the `IORef` wrapper from the vault key so `AutoRefreshState` is stored directly in the vault
- `autoRefreshMeta` now reads from `?context.request.vault` instead of `fromFrozenContext`
- Move the state update (`AutoRefreshEnabled`) before `freeze` so the frozen context captures it

## Test plan

- [ ] Verify `autoRefreshMeta` emits the `<meta property="ihp-auto-refresh-id">` tag when auto-refresh is active
- [ ] Verify auto-refresh WebSocket reconnection still works end-to-end
- [ ] Verify `nix flake check --impure` passes

### DIFF
--- a/Guide/auto-refresh.markdown
+++ b/Guide/auto-refresh.markdown
@@ -6,7 +6,7 @@
 
 ## Setup
 
-Auto Refresh is enabled by default in new IHP applications. If you've disabled it or are setting it up manually, you need these two components:
+Auto Refresh requires no global configuration. Just make sure these two components are in your layout:
 
 ### 1. Add Meta Tag to Layout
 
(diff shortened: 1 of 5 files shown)


amitaibu:
We can remove this line as well https://github.com/digitallyinduced/ihp/pull/2286/changes#diff-b4fd4b780794bbc3e25b384bc3158ce1ba0fe8dccf6cbee2ee082d6e3251d2e4R166

mpscholten:
Already done — commit 41aa5110 removed the `AutoRefreshState` hint from `knownTypes` in an earlier commit on this branch.

PR #2349, opened 2026-02-08
Add CI workflow to test devenv up end-to-end

### LINKED ISSUE #2332: Test a full app with `devenv up`
We already have integration testing added in https://github.com/digitallyinduced/ihp/pull/2314

What if we had a full directory called `ihp-app-test` and inside a GH action, we would:

1. symlink IHP under it (`/IHP`)
2. Then we would run the tests
3. And on top of the integration tests, we could examine `process-compose.log` to test stuff like, a Job worker triggered. Currently, checking that `devenv up` is working, needs to be done manually in other projects

amitaibu:
## Summary
- Adds a GitHub Actions workflow that verifies `devenv up` works end-to-end with IHP changes
- Clones `ihp-boilerplate`, patches `flake.nix` to point at the current IHP branch, runs `devenv up`, and smoke-tests that the server starts and responds on port 8000
- No files duplicated — uses the real boilerplate repo as the test app

Closes #2332

## Test plan
- [ ] Verify the new `ihp-app-test (devenv up)` workflow runs successfully in CI
- [ ] Check that `nix flake update` resolves the patched `ihp.url` correctly
- [ ] Confirm `devenv up` starts PostgreSQL + IHP dev server and port 8000 responds with 200

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,12 @@ jobs:
         os: [ubuntu-latest, macos-latest, ARM64]
     runs-on: ${{ matrix.os }}
     steps:
+    # On macOS case-insensitive filesystem, directory names from previous
+    # workflows can persist with wrong case (e.g. IHP/ instead of ihp/).
+    # Git won't rename them because core.ignorecase=true.
+    # Delete and let checkout recreate with correct case from the git index.
+    - run: rm -rf ihp
+      if: matrix.os == 'ARM64'
     - uses: actions/checkout@v6
     - uses: wimpysworld/nothing-but-nix@main
       with:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,5 +8,10 @@ jobs:
   nix-flake-check:
     runs-on: ARM64
     steps:
+    # On macOS case-insensitive filesystem, directory names from previous
+    # workflows can persist with wrong case (e.g. IHP/ instead of ihp/).
+    # Git won't rename them because core.ignorecase=true.
+    # Delete and let checkout recreate with correct case from the git index.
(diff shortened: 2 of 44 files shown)


amitaibu:
@mpscholten before I continue to work on this, is this a direction you'd like to take?

mpscholten:
I'd prefer this as a nix flake check, that makes it easy to run locally

mpscholten:
Heads up: this workflow's `path: IHP` checkout broke CI on master.

On the self-hosted macOS ARM64 runner (case-insensitive filesystem), checking out into `IHP/` renames the repo's `ihp/` directory to uppercase on disk. Git with `core.ignorecase=true` won't fix the case on subsequent checkouts, so the next `nix flake check` run copies the wrong-cased directory to nix's case-sensitive store and Cabal can't find any modules.

Fixed in 6bb309ff by adding `rm -rf ihp` before checkout in the CI workflows, but this workflow should use a non-colliding checkout path (e.g. `path: ihp-checkout` or `path: ihp-source`) to avoid the issue entirely.

mpscholten:
(prev comment by claude)

That's actually another good reason to go with nix flake check :D this build here did break master and it was very hard to find out why. Claude eventually found it, initially I thought claude was wrong, but appartnelty it was right

amitaibu:
Yikes! Didn't imagine a non-main branch breaking `master` :fearful:

amitaibu:
I'll close this, and maybe Claude can tackle https://github.com/digitallyinduced/ihp/issues/2332

I don't think my nix skills would be up for prompting AI in the correct direction without some effort :sweat_smile:

mpscholten:
just let claude handle, its it's our best nix programmer 😄

PR #2357, opened 2026-02-09
Use hasql for RLS queries instead of falling back to postgresql-simple

mpscholten:
## Summary

- When Row Level Security (RLS) is enabled, queries now go through hasql with prepared statement caching instead of falling back to postgresql-simple
- RLS is implemented by wrapping hasql queries in a transaction with `set_config()` to set the role and user id, matching the pattern already proven in DataSync
- Moves `setRLSConfigStatement` to `IHP.ModelSupport` as the canonical location; DataSync reuses it via import
- Changes `rlsUserId` type from `Action` to `Text` for simpler, hasql-compatible encoding
- Routes raw `sqlExec` through hasql when a pool is available

## Test plan

- [x] All 361 existing tests pass (including RLS-specific `withRLSParams` tests)
- [ ] Manual test with an IHP app that has RLS enabled to verify queries work correctly
- [ ] Verify DataSync RLS queries still work (imports changed)

### DIFF
--- a/ihp-ide/IHP/IDE/SchemaDesigner/Controller/Migrations.hs
+++ b/ihp-ide/IHP/IDE/SchemaDesigner/Controller/Migrations.hs
@@ -15,11 +15,12 @@ import qualified IHP.SchemaMigration as SchemaMigration
 import qualified IHP.IDE.CodeGen.MigrationGenerator as MigrationGenerator
 import IHP.IDE.CodeGen.Controller
 import IHP.IDE.ToolServer.Helper.Controller (openEditor, clearDatabaseNeedsMigration)
-import IHP.Log.Types
 import qualified Control.Exception.Safe as Exception
 import qualified System.Directory.OsPath as Directory
 import qualified Database.PostgreSQL.Simple as PG
 import System.OsPath (encodeUtf)
+import qualified Hasql.Connection as Connection
+import qualified Hasql.Connection.Settings as ConnectionSettings
 
 instance Controller MigrationsController where
     beforeAction = setLayout schemaDesignerLayout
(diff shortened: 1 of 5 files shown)


ihp-datasync/IHP/DataSync/Pool.hs
```diff
@@ -15,9 +15,7 @@ import qualified IHP.DataSync.Role as Role
 
 -- | Returns the hasql pool from the model context.
 requestHasqlPool :: (?modelContext :: ModelContext) => Hasql.Pool.Pool
-requestHasqlPool = case ?modelContext.hasqlPool of
-    Just pool -> pool
-    Nothing -> error "requestHasqlPool: No hasql pool available in ModelContext"
+requestHasqlPool = ?modelContext.hasqlPool
```
mpscholten: the name is misleading. likely we can just drop the function entirely

ihp-ide/IHP/IDE/Data/Controller.hs
```diff
@@ -189,7 +189,7 @@ instance Controller DataController where
             Nothing -> renderNotFound
 
 getPool :: (?modelContext :: ModelContext) => HasqlPool.Pool
-getPool = fromMaybe (error "No hasql pool available") ?modelContext.hasqlPool
+getPool = ?modelContext.hasqlPool
```
mpscholten: can be dropped

ihp/IHP/Hasql/Encoders.hs
```diff
@@ -162,6 +174,35 @@ instance DefaultParamEncoder (Maybe Polygon) where
                 Just pg -> pg
                 Nothing -> error "Polygon must have at least 3 points"
 
+-- | Encode 'PGInterval' as PostgreSQL interval (sent as text)
+-- PGInterval wraps the raw ByteString from postgresql-simple's interval representation
+instance DefaultParamEncoder PGInterval where
+    defaultParam = Encoders.nonNullable (contramap (\(PGInterval bs) -> Text.decodeUtf8 bs) Encoders.text)
```
mpscholten: this will cause an oid mismatch, same with the other ones below

PR #2374, opened 2026-02-10
Fix AutoRefresh crash: multi-statement SQL incompatible with hasql pipeline mode

mpscholten:
## Summary

- AutoRefresh's `notificationTrigger` sent a multi-statement SQL block (BEGIN + CREATE FUNCTION + DROP/CREATE TRIGGERs + COMMIT) as a single hasql statement, which crashes because hasql's extended query protocol only supports one statement per execution
- Split into individual statements using `notificationTriggerStatements` and execute via `runSessionHasql` with `HasqlSession.script`, matching the pattern from the Job/Queue.hs fix in #2373

## Test plan

- [ ] Verify AutoRefresh works in a project with tables that trigger `pg_notify` (e.g. any page using `autoRefresh`)
- [ ] Confirm no `StatementSessionError` / pipeline mode errors on first page load

### DIFF
--- a/ihp/IHP/AutoRefresh.hs
+++ b/ihp/IHP/AutoRefresh.hs
@@ -24,8 +24,7 @@ import IHP.WebSocket
 import IHP.Controller.Context
 import IHP.Controller.Response
 import qualified IHP.PGListener as PGListener
-import qualified Database.PostgreSQL.Simple.Types as PG
-import Data.String.Interpolate.IsString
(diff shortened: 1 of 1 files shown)


mpscholten:
@amitaibu I think this one should fix your issue now :D

amitaibu:
@mpscholten thanks, it's loading properly now. Still no autoRefrsh after `make db`, though.

PR #2384, opened 2026-02-11
Fix JSON content-type matching to handle charset parameter

amitaibu:
## Summary
- The `requestBodyMiddleware` did an exact match on `"application/json"` for the Content-Type header, which failed when the header included parameters like `application/json; charset=utf-8`
- Changed to use `BS.isPrefixOf "application/json"` so that any Content-Type starting with `application/json` is correctly parsed as JSON
- Added a test case verifying that `application/json; charset=utf-8` is parsed as `JSONBody`

## Test plan
- [ ] Existing tests continue to pass (exact `application/json` still works)
- [ ] New test verifies `application/json; charset=utf-8` is parsed as JSONBody
- [ ] Run `cabal test wai-request-params`


mpscholten (approved):
Thanks 👍

PR #2396, opened 2026-02-12
Improve requestBodyJSON error messages for different failure modes

amitaibu:
## Summary

- **Fix: Use `strictRequestBody` instead of `lazyRequestBody` for JSON body parsing** — `lazyRequestBody` uses `unsafeInterleaveIO`, deferring the actual body read until the lazy bytestring is forced. By the time it's evaluated in the controller, other middlewares may have already consumed the body stream, resulting in an empty body. `strictRequestBody` eagerly reads the entire body upfront.
- `requestBodyJSON` now distinguishes three failure modes instead of a generic "Expected JSON body" error:
  - **Invalid JSON** (with `application/json` content type): shows the raw request body for debugging
  - **Empty body** (with `application/json` content type): reports the body is empty
  - **Form content type**: hints at missing `Content-Type: application/json` header
- Adds tests for the new error messages in `ControllerSupportSpec`
- Adds middleware-level tests documenting behavior for invalid JSON, missing content type, and empty body scenarios

## Test plan

- [x] All 4 `ControllerSupportSpec` tests pass
- [x] All 14 `MiddlewareSpec` tests pass (3 new)
- [x] Full IHP test suite passes (427 examples, 0 failures)
- [x] Manually verified curl POST with JSON body now works correctly

### DIFF
--- a/ihp/IHP/ControllerSupport.hs
+++ b/ihp/IHP/ControllerSupport.hs
@@ -35,7 +35,7 @@ import Data.IORef (IORef, modifyIORef', readIORef)
 import Data.ByteString (ByteString)
 import qualified Data.ByteString.Lazy as LBS
 import Data.Maybe (fromMaybe)
-import Control.Exception.Safe (SomeException, fromException, try, catches, Handler(..), throwIO)
+import Control.Exception.Safe (SomeException, fromException, try, catch, throwIO)
 import Data.Typeable (Typeable)
 import IHP.HaskellSupport
 import Network.Wai
@@ -94,9 +94,9 @@ runAction controller = do
             (action controller)
             ErrorController.handleNoResponseReturned controller
 
-    let handleResponseException (ResponseException response) = ?respond response
-
-    doRunAction `catches` [ Handler handleResponseException, Handler (\exception -> ErrorController.displayException exception controller "")]
+    doRunAction `catch` \(exception :: SomeException) -> case fromException exception of
+        Just (ResponseException response) -> ?respond response
+        Nothing -> ErrorController.displayException exception controller ""
 
 {-# INLINE newContextForAction #-}
 newContextForAction
@@ -174,6 +174,7 @@ prepareRLSIfNeeded modelContext = do
     case rowLevelSecurityContext of
         Just context -> pure modelContext { rowLevelSecurity = Just context }
         Nothing -> pure modelContext
+{-# INLINE prepareRLSIfNeeded #-}
 
 rlsContextVaultKey :: Vault.Key (IORef (Maybe RowLevelSecurityContext))
 rlsContextVaultKey = unsafePerformIO Vault.newKey
--- a/ihp/IHP/RouterSupport.hs
+++ b/ihp/IHP/RouterSupport.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE AllowAmbiguousTypes, UndecidableInstances, LambdaCase #-}
+{-# LANGUAGE AllowAmbiguousTypes, UndecidableInstances, LambdaCase, ScopedTypeVariables #-}
 module IHP.RouterSupport (
 CanRoute (..)
 , HasPath (..)
@@ -28,6 +28,8 @@ CanRoute (..)
 , onlyAllowMethods
 , getMethod
 , routeParam
+, withImplicits
+, applyConstr
 ) where
 
 import Prelude hiding (take)
@@ -53,6 +55,8 @@ import Data.Data
 import qualified Control.Monad.State.Strict as State
 import qualified Data.Text as Text
 import Network.HTTP.Types.URI
+import Network.HTTP.Types.Status (status400)
+import Network.HTTP.Types.Header (hContentType)
 import qualified Data.List as List
 import Unsafe.Coerce
 import IHP.HaskellSupport hiding (get)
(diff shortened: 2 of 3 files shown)


ihp/Test/Test/ControllerSupportSpec.hs
```diff
@@ -0,0 +1,41 @@
...
+                let requestBody = FormBody { params = [], files = [] }
+                let request = Wai.defaultRequest { Wai.vault = Vault.insert requestBodyVaultKey requestBody Vault.empty }
+                let ?request = request
+                (IO.evaluate requestBodyJSON) `shouldThrow` errorCall "Expected JSON body, but the request has a form content type. Make sure to set 'Content-Type: application/json' in the request header."
+
+            it "should throw descriptive error for JSONBody with empty body" do
+                let requestBody = JSONBody { jsonPayload = Nothing, rawPayload = "" }
+                let request = Wai.defaultRequest { Wai.vault = Vault.insert requestBodyVaultKey requestBody Vault.empty }
+                let ?request = request
+                (IO.evaluate requestBodyJSON) `shouldThrow` errorCall "Expected JSON body, but could not decode the request body. The request body is empty. This usually means another WAI middleware consumed the body before it could be parsed."
```
amitaibu: I don't think this error message is correct.

wai-request-params/Wai/Request/Params/Middleware.hs
```diff
@@ -60,7 +60,7 @@ requestBodyMiddleware parseRequestBodyOptions app req respond = do
             let contentType = lookup hContentType (requestHeaders req)
             case contentType of
                 Just ct | isJsonContentType ct -> do
-                    rawPayload <- lazyRequestBody req
+                    rawPayload <- strictRequestBody req
```
amitaibu: This change allowed my previously failing curl to work

mpscholten:
@codex review

ihp/IHP/ControllerSupport.hs
```diff
@@ -274,7 +274,13 @@ requestBodyJSON :: (?request :: Request) => Aeson.Value
 requestBodyJSON =
     case ?request.parsedBody of
         JSONBody { jsonPayload = Just value } -> value
-        _ -> error "Expected JSON body"
+        JSONBody { jsonPayload = Nothing, rawPayload } ->
+            error ("Expected JSON body, but could not decode the request body"
```
mpscholten: instead of calling error we should render an error http response (with some status code for invalid request) and the body being the error message

wai-request-params/Wai/Request/Params/Middleware.hs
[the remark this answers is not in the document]
amitaibu: I guess this is an important part, but I couldn't figure out what consumed the body before it reached my controller.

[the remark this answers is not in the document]
mpscholten: lets keep the strictRequestBody here, it's only triggered for json requests anyways, so fine to me :)

ihp/IHP/ControllerSupport.hs
```diff
@@ -274,7 +274,13 @@ requestBodyJSON :: (?request :: Request) => Aeson.Value
...
+                <> (if LBS.null rawPayload
+                    then ". The request body is empty."
+                    else ". The raw request body was: " <> show rawPayload))
```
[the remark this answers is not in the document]
mpscholten: let's output the full response only when app is running in dev mode

amitaibu:
Updated `requestBodyJSON` to return HTTP 400 instead of crashing with `error` (500):

**Changes:**
- **Signature**: `Aeson.Value` → `IO Aeson.Value` (to support `respondAndExit` pattern)
- **Error responses**: All error cases now return HTTP 400 with `Content-Type: application/json` and a `{"error": "..."}` body
- **Raw payload truncation**: Invalid JSON payloads are truncated to 200 chars in error messages, and only shown in dev mode (`IHP_ENV != Production`)
- **Callers updated**: `ihp-datasync/IHP/DataSync/REST/Controller.hs` updated from `let` to monadic `<-` binding

**Implementation note:** Uses `throwIO (ResponseException ...)` directly rather than `renderJsonWithStatusCode` because `requestBodyJSON` returns `IO Aeson.Value` (not `IO ()`). Both use the same underlying `ResponseException` mechanism — `throwIO` just gives us the polymorphic return type needed here.

All tests pass (427 examples, 0 failures).

ihp/IHP/ControllerSupport.hs
```diff
@@ -270,11 +272,33 @@ getFiles =
...
-requestBodyJSON :: (?request :: Request) => Aeson.Value
+requestBodyJSON :: (?request :: Request) => IO Aeson.Value
 requestBodyJSON =
     case ?request.parsedBody of
-        JSONBody { jsonPayload = Just value } -> value
-        _ -> error "Expected JSON body"
+        JSONBody { jsonPayload = Just value } -> pure value
+        JSONBody { jsonPayload = Nothing, rawPayload } -> do
+            ihpEnv <- System.Environment.lookupEnv "IHP_ENV"
+            let isDev = ihpEnv /= Just "Production"
```
mpscholten: Should be something like:
```suggestion
            let isDev = ?request.frameworkConfig.environment == Development
```

ihp/IHP/ControllerSupport.hs
```diff
@@ -270,11 +272,33 @@ getFiles =
         FormBody { files } -> files
         _ -> []
 
-requestBodyJSON :: (?request :: Request) => Aeson.Value
+requestBodyJSON :: (?request :: Request) => IO Aeson.Value
```
amitaibu, line 273: @mpscholten add to Changelog  [Breaking Changes](https://github.com/digitallyinduced/ihp/blob/ac6f9660db73244d6449bbdc09147832edb880be/CHANGELOG.md#breaking-changes) that `requestBodyJSON` now returens IO?

mpscholten: 👍

mpscholten (approved):
Looks good to 👍 Can you tell claude to resolve merge conflicts? Then it's ready to go

PR #2401, opened 2026-02-12
Fix AutoRefresh losing query parameters during re-render

mpscholten:
## Summary
- When AutoRefresh re-renders a view after a database change, it uses the WebSocket request context which has an empty query string
- This caused `ParamNotFoundException` for actions that rely on query parameters (e.g. `?marketId=...&tradingAssetId=...`)
- Fix: copy `rawQueryString` and `queryString` from the original HTTP request into the re-render request, alongside the `vault` that was already being preserved

## Test plan
- [ ] Test with an action that uses query parameters and AutoRefresh enabled
- [ ] Verify that re-renders after database changes correctly preserve query parameters

### DIFF
--- a/ihp/Test/Test/AutoRefreshSpec.hs
+++ b/ihp/Test/Test/AutoRefreshSpec.hs
@@ -1,49 +1,146 @@
-module Test.AutoRefreshSpec where
+{-|
+Module: Test.AutoRefreshSpec
+Tests that AutoRefresh preserves query parameters when re-rendering
+with a bare WebSocket request (no query params).
+-}
+{-# LANGUAGE AllowAmbiguousTypes #-}
 
-import Prelude
-import Data.Text (Text)
-import Data.IORef
+module Test.AutoRefreshSpec where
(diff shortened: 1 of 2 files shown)


ihp/Test/Test/AutoRefreshSpec.hs
```diff
@@ -0,0 +1,51 @@
...
+
+                -- Set up controller context with the original request (mirrors autoRefresh setup)
+                controllerContext <- newControllerContext
+                let ?context = controllerContext
+                putContext originalRequest
+                frozenControllerContext <- freeze ?context
+
+                -- This mirrors the renderView closure from IHP.AutoRefresh
+                -- Previously it used: let waiRequest' = waiRequest { Wai.vault = Wai.vault originalRequest }
+                -- which lost the query string. Now it uses originalRequest directly.
```
amitaibu:
```suggestion
                -- Ensure we don't lose the query string. Use originalRequest directly.
```

PR #2407, opened 2026-02-14
Fix: getRequestBody returns empty for non-JSON requests

amitaibu:
## Summary

- `getRequestBody` returned empty bytes for non-JSON POST requests because the WAI body stream was already consumed by the form parser
- Added `rawPayload` field to `FormBody` (it already existed on `JSONBody`) so raw bytes are always preserved
- Middleware now reads raw bytes via `strictRequestBody` before form parsing, then replays them for the form parser via an IORef-backed body reader
- Simplified `getRequestBody` to `pure ?request.parsedBody.rawPayload` — works uniformly for both constructors

## Test plan

- [x] 2 new tests in `MiddlewareSpec.hs` confirm raw body is preserved for form-encoded and non-JSON POST requests
- [x] Tests fail without the fix (body returns `""`) and pass with it (TDD verified)
- [x] All 361 existing IHP tests pass
- [x] All 13 middleware tests pass

### DIFF
--- a/ihp/IHP/ControllerSupport.hs
+++ b/ihp/IHP/ControllerSupport.hs
@@ -215,7 +215,6 @@ jumpToAction theAction = do
     beforeAction @action
     action theAction
 
-{-# INLINE getRequestBody #-}
 getRequestBody :: (?request :: Request) => IO LBS.ByteString
 getRequestBody =
     pure ?request.parsedBody.rawPayload


wai-request-params/Test/Wai/Request/Params/MiddlewareSpec.hs
```diff
@@ -20,13 +20,24 @@ inspectBodyApp :: Application
...
             respond $ responseLBS status200 [] (cs $ "FormBody params=" <> show (length params) <> " files=" <> show (length files))
         Just (JSONBody jsonPayload _) ->
             respond $ responseLBS status200 [] (cs $ "JSONBody payload=" <> show jsonPayload)
         Nothing ->
             respond $ responseLBS status200 [] "no body in vault"
 
+-- | An app that reads the rawPayload from the middleware-parsed body.
+-- This simulates what getRequestBody does: it accesses parsedBody.rawPayload.
+-- Before the fix, FormBody had no rawPayload field, so getRequestBody fell
+-- back to lazyRequestBody which returned empty (stream already consumed).
```
amitaibu:
```suggestion
```

mpscholten (approved):
Thanks 👍 looks good

PR #2423, opened 2026-02-16
Hsx svg

Montmorency:
This PR adds a few missing svg attributes and tags as advised by claude and builds out svg suite. It also allows hsx to parse some additional characters that appear in htmx/hyperscript attributes like ':' and '_="disable #target on click"' etc.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -35,50 +35,8 @@
     },
     "cachix_2": {
       "inputs": {
-        "devenv": [
-          "ihp-boilerplate",
-          "ihp",
-          "devenv"
-        ],
+        "devenv": "devenv_3",
         "flake-compat": [
-          "ihp-boilerplate",
-          "ihp",
-          "devenv"
-        ],
-        "git-hooks": [
-          "ihp-boilerplate",
-          "ihp",
-          "devenv",
-          "git-hooks"
-        ],
-        "nixpkgs": [
-          "ihp-boilerplate",
-          "ihp",
-          "devenv",
-          "nixpkgs"
-        ]
-      },
-      "locked": {
-        "lastModified": 1748883665,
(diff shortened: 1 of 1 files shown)


ihp-hsx/parser/IHP/HSX/Parser.hs
```diff
@@ -276,10 +276,11 @@ hsxAttributeName = do
             "data-" `Text.isPrefixOf` name
             || "aria-" `Text.isPrefixOf` name
             || "hx-" `Text.isPrefixOf` name
+            || name == "_"
```
mpscholten, line 281: can you add comments here  why this is needed? otherwise we might accidentally remove it in the future

ihp-hsx/parser/IHP/HSX/Parser.hs
```diff
@@ -276,10 +276,11 @@ hsxAttributeName = do
...
             || name `Set.member` attributes
             || name `Set.member` ?settings.additionalAttributeNames
 
-        rawAttribute = takeWhile1P Nothing (\c -> Char.isAlphaNum c || c == '-' || c == '_')
+        rawAttribute = takeWhile1P Nothing (\c -> Char.isAlphaNum c || c == '-' || c == '_' || c == ':')
```
mpscholten, line 287: same here

ihp-hsx/HSX_SVG.md
```diff
@@ -0,0 +1,178 @@
+# SVG Support in HSX
```
mpscholten: Can this be removed?

Montmorency:
Added comments saying that the underscore is valid WHATG (and kind of nice) syntax for attribute that defines some scripting commands. removed excess doc file.

mpscholten:
Looks good, but there's a flake.lock file change. Can you remove that from the PR ?

Montmorency:
Reverted to master flake lock.

Issue #2417, opened 2026-02-16
Bounds on hasql >=1.8 && <1.10


Montmorency:
```
$ nix flake update
$ nix develop --impure
```
Giving a bounds check error from cabal

```
       > Using Parsec parser
       > Configuring hasql-interpolate-1.0.1.0...
       > Error: [Cabal-8010]
       > Encountered missing or private dependencies:
       >     hasql >=1.8 && <1.10
       > CallStack (from HasCallStack):
       >   dieWithException, called at libraries/Cabal/Cabal/src/Distribution/Simple/Configure.hs:1457:11 in Cabal-3.12.1.0-5e6a:Distribution.Simple.Configure
       >
       For full logs, run:
         nix log /nix/store/map2j7kk6ll1lr1wcl9c3c98ijy57vx2-hasql-interpolate-1.0.1.0.drv
error: Cannot build '/nix/store/p77l05vxmv78nsrn6g0r5z3hgvfi1jkw-ghc-9.10.3-with-packages.drv'.
       Reason: 1 dependency failed.
       Output paths:
         /nix/store/3dwdi3albqw57s4yds1r76m50fak51a2-ghc-9.10.3-with-packages
error: Build failed

Building off latest master rev=c9d3fff4
```
$ nix log 
```Configuring hasql-interpolate-1.0.1.0...
Error: [Cabal-8010]
Encountered missing or private dependencies:
    hasql >=1.8 && <1.10
CallStack (from HasCallStack):
  dieWithException, called at libraries/Cabal/Cabal/src/Distribution/Simple/Configure.hs:1457:11 in Cabal-3.12.1.0-5e6a:Distribution.Simple.Configure
```

mpscholten:
`hasql-interpolate` hasn't been updated to latest version yet. See https://github.com/awkward-squad/hasql-interpolate/pull/27 I'd go with [hasql-th](https://hackage.haskell.org/package/hasql-th) as that's by the same developer that also makes hasql

PR #2426, opened 2026-02-16
Fix hasql-interpolate Hackage cabal revision fetch failure

Montmorency:
## Summary

  - Fixes hash mismatch error when downstream apps run `nix develop` with `hasql-interpolate` as a dependency
  - Replaces `callCabal2nix` with `overrideCabal` to avoid IFD and stale Hackage cabal revision hashes
  - Related: #2419, #2417

  ## Problem

  PR #2419 added `hasql-interpolate` to the overlay using `callCabal2nix`, which uses **IFD (Import From Derivation)** — a Nix mechanism where a
  derivation is built *during evaluation* (not during the build phase) so its output can be imported as a Nix expression. In this case,
  `callCabal2nix` builds `cabal2nix` at eval time to convert the `.cabal` file into a Nix derivation.

  This approach still triggered a fetch of the Hackage cabal file revision (`hasql-interpolate-1.0.1.0-r3.cabal`), which fails with a hash mismatch
  when nixpkgs' `all-cabal-hashes` snapshot becomes stale relative to Hackage.

  ## Fix

  Switch to `overrideCabal super.hasql-interpolate` which:
  1. Reuses the existing nixpkgs derivation structure (no IFD needed)
  2. Replaces `src` with ChrisPenner's hasql-1.10-compatible fork
  3. Sets `revision = null` and `editedCabalFile = null` to prevent the Hackage cabal revision fetch entirely
  4. Adds `doJailbreak` to relax the `hasql <1.10` version bound

  ## Test plan

  - [ ] `nix develop --impure` succeeds in a downstream app with `hasql-interpolate` in `haskellPackages`
  - [ ] `import Hasql.Interpolate` works in application code

### DIFF
--- a/NixSupport/overlay.nix
+++ b/NixSupport/overlay.nix
@@ -119,10 +119,15 @@ let
             hasql-notifications = final.haskell.lib.dontCheck (self.callHackageDirect { pkg = "hasql-notifications"; ver = "0.2.5.0"; sha256 = "11jkrngiy175wc5hqx8pgagj4fdg42ry7afp4g4rr5hw8h43zg48"; } {});
             # hasql-interpolate: upstream 1.0.1.0 requires hasql <1.10; use fork with hasql 1.10 support
             # https://github.com/awkward-squad/hasql-interpolate/pull/27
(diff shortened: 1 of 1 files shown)


Montmorency:
Above created with Claude, tag missing because could not submit via cli with gh.

PR #2429, opened 2026-02-16
Fix tooltip positioning in IDE Data tab

### LINKED ISSUE #2408: Can't Add or Edit records on IDE's `Data`
https://github.com/user-attachments/assets/8eca2081-9695-4225-a4f8-41dce3d739de

mpscholten:
## Summary
- Fixes the mispositioned tooltip on the `+` button in the IDE's Data tab, where "Add TableName" was appearing at the top-right corner of the viewport instead of near the button
- Uses Popper.js `strategy: 'fixed'` so tooltips use `position: fixed` instead of `position: absolute`, avoiding incorrect offset calculations caused by the body's flex layout (`d-flex flex-row`)
- Cleans up orphaned tooltip DOM elements during Turbolinks/morphdom page transitions

## Test plan
- [ ] Open IDE Data tab, hover over the `+` button in the table header — tooltip should appear near the button, not at the top-right corner
- [ ] Navigate between Schema/Data tabs and verify tooltips still work correctly
- [ ] Verify Schema Designer tooltips (Add Column, Add Table) still position correctly

Fixes #2408

### DIFF
--- a/ihp-ide/data/static/IDE/ihp-schemadesigner.js
+++ b/ihp-ide/data/static/IDE/ihp-schemadesigner.js
@@ -187,10 +187,13 @@ function initCodeEditor() {
 }
 
 function initTooltip() {
+    // Remove any orphaned tooltip elements left over from Turbolinks/morphdom transitions
+    document.querySelectorAll('body > .tooltip').forEach(function (el) { el.remove(); });
+
     document.querySelectorAll('[data-bs-toggle="tooltip"]').forEach(function (el) {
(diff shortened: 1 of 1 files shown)


amitaibu:
How did Claude find the issue? Did it curl the HTML and see something was wrong there?

mpscholten:
I'm not sure, I didn't watch (I do have the chrome plugin enabled, so it could have used chrome to take a closer look)

PR #2431, opened 2026-02-16
Add FromField/ToField instances for postgresql-types types

Montmorency:
When IHP migrated to the postgresql-types package, hasql encoders were added but postgresql-simple FromField/ToField orphan instances were not. This caused compile errors like "No instance for 'FromField Point'" for any table using Point, Polygon, Inet, Interval, or Tsvector columns.

Uses IsScalar from postgresql-types-algebra to provide a generic bridge via textualDecoder/textualEncoder, keeping each instance a one-liner.

Useful with apps requiring postgis Points.

### DIFF
--- a/ihp/IHP/PGSimpleCompat.hs
+++ b/ihp/IHP/PGSimpleCompat.hs
@@ -13,7 +13,7 @@ import Data.String (IsString(..))
 import Data.String.Conversions (cs)
 import Data.Data (Typeable)
 import GHC.TypeLits (KnownSymbol, symbolVal)
-import Database.PostgreSQL.Simple.FromField (FromField(..), ResultError(..), returnError)
+import Database.PostgreSQL.Simple.FromField (FromField(..))
 import Database.PostgreSQL.Simple.ToField (ToField(..), Action(..))
 import qualified Database.PostgreSQL.Simple.FromRow as PGFR
 import qualified Database.PostgreSQL.Simple.Types as PG


mpscholten:
@Montmorency `postgresql-types` also has a package for `postgresql-simple` bindings. E.g. here https://github.com/nikita-volkov/postgresql-simple-postgresql-types/blob/master/src/library/Database/PostgreSQL/Simple/PostgresqlTypes.hs it seems that it defines all these instances for us. Maybe we should just use that module directly?

NixSupport/overlay.nix
```diff
@@ -131,6 +131,12 @@ let
 
             # postgresql-types for proper binary encoders of Point, Polygon, Inet, Interval
             ptr-poker = self.callHackageDirect { pkg = "ptr-poker"; ver = "0.1.3"; sha256 = "0jl9df0kzsq5gd6fhfqc8my4wy7agg5q5jw4q92h4b7rkdf3hix7"; } {};
+            # postgresql-simple-postgresql-types: bridge providing FromField/ToField instances
+            # for all postgresql-types types (Point, Polygon, Inet, Interval, etc.) in postgresql-simple
+            postgresql-simple-postgresql-types = final.haskell.lib.dontCheck (final.haskell.lib.doJailbreak (self.callCabal2nix "postgresql-simple-postgresql-types" (builtins.fetchTarball {
+                url = "https://github.com/nikita-volkov/postgresql-simple-postgresql-types/archive/1aebd39.tar.gz";
+                sha256 = "0000000000000000000000000000000000000000000000000000";
```
mpscholten: needs a real sha hash

Issue #1162, opened 2021-10-21
Allow multiple Paginations per page


amitaibu:
Currently the Pagination module assumes there's only a single pager on the page.

This assumption is in the query string - https://github.com/digitallyinduced/ihp/blob/6dd1b6e2f3263ae7531831c3a649e7229df6abfc/IHP/Pagination/ControllerFunctions.hs#L91-L92

And in the HTML (`id="maxItems...`)- https://github.com/digitallyinduced/ihp/blob/6dd1b6e2f3263ae7531831c3a649e7229df6abfc/IHP/Pagination/ViewFunctions.hs#L50

Maybe Pagination should get a `Text` prefix that would prefix to the query strings

PR #2444, opened 2026-02-17
Remove postgresql-simple dependency from Pagination module

mpscholten:
## Summary
- Rewrites `paginatedSqlQuery`/`paginatedSqlQueryWithOptions` to use the hasql Snippet API directly instead of routing through `sqlQuery`/`sqlQueryScalar` with postgresql-simple types (`Query`, `Only`, `(:.)`)
- Changes the SQL parameter type from `Query` to `Text` and removes the `ToSnippetParams (parameters :. Only Int :. Only Int)` constraint
- Removes the `Database.PostgreSQL.Simple` import from `IHP.Pagination.ControllerFunctions` entirely

## Test plan
- [x] Verified `ihp` package compiles successfully via `nix flake check --impure`
- [ ] Test `paginatedSqlQuery` in an IHP app with raw SQL pagination queries

### DIFF
--- a/ihp/Test/Test/Pagination/ControllerFunctionsSpec.hs
+++ b/ihp/Test/Test/Pagination/ControllerFunctionsSpec.hs
@@ -26,10 +26,10 @@ tests = do
                 let ?context = contextWithParams []
                 let ?request = ?context.request
 
-                (results :: [PG.Only Int], pagination) <-
+                (results :: [PG.Only Int32], pagination) <-
                     paginatedSqlQueryWithOptions
                         defaultPaginationOptions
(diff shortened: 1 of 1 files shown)


amitaibu:
If we're already touching pagination, I think currently we allow only a single pagination per page. Maybe we can extend ot have multiple.

mpscholten:
> If we're already touching pagination, I think currently we allow only a single pagination per page. Maybe we can extend ot have multiple.

I only want to make it use hasql because it prevents one app from upgrading. For a feature change we should do that in a different pr

amitaibu:
Sure. I found this original issue https://github.com/digitallyinduced/ihp/issues/1162

PR #2446, opened 2026-02-17
Allow multiple paginations per page (#1162)

amitaibu:
## Summary
- Adds `paramSuffix` field to `Pagination` and `PaginationView` types, auto-assigned via a `PaginationCounter` in the ControllerContext
- First `paginate` call per request uses no suffix (backward compatible), subsequent calls get `_2`, `_3`, etc.
- Query params (`page`, `maxItems`, `filter`) and HTML element IDs (`maxItemsSelect`) are suffixed accordingly
- Adds `renderFilterFor` for multi-pagination filter support while keeping `renderFilter` backward compatible

## Test plan
- [ ] Verify existing pagination behavior is unchanged (first pagination uses unsuffixed params)
- [ ] Test two paginated lists on the same page use independent query params (`page` vs `page_2`, `maxItems` vs `maxItems_2`)
- [ ] Verify `maxItemsSelect` HTML IDs are unique per pagination
- [ ] Run `echo -e ':l ihp/Test/Test/Main.hs\nmain' | ghci` (blocked by pre-existing PGSimpleCompat issue on base branch)

### DIFF
--- a/ihp/Test/Test/Pagination/ControllerFunctionsSpec.hs
+++ b/ihp/Test/Test/Pagination/ControllerFunctionsSpec.hs
@@ -16,6 +16,7 @@ import qualified Network.Wai as Wai
 import qualified Database.PostgreSQL.Simple.Types as PG
 import System.Environment (lookupEnv)
 import qualified Control.Exception as Exception
+import Data.IORef (newIORef)
 
 tests :: Spec
 tests = do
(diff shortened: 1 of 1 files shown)


amitaibu:
I'll wait for https://github.com/digitallyinduced/ihp/pull/2444, and then continue with this

PR #2455, opened 2026-02-19
Show friendly error when migrations page can't connect to DB

mpscholten:
## Summary
- Previously, if the database was unreachable when opening the migrations page, the user saw an unhelpful `ExitFailure 1` exception with a raw HasCallStack backtrace
- Now the error is caught and displayed as a flash message ("Could not connect to the database: ..."), and the page still renders with all migrations shown as pending

## Test plan
- [ ] Stop the database and open the migrations page — verify a flash error message is shown instead of a crash
- [ ] Start the database and open the migrations page — verify it works as before

### DIFF
--- a/ihp-ide/IHP/IDE/SchemaDesigner/Controller/Migrations.hs
+++ b/ihp-ide/IHP/IDE/SchemaDesigner/Controller/Migrations.hs
@@ -26,7 +26,13 @@ instance Controller MigrationsController where
 
     action MigrationsAction = do
         migrations <- findRecentMigrations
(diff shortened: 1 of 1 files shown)


mpscholten:
Re: the Codex review about `setErrorMessage` only showing on the next request —

This is actually fine in IHP. `render` internally calls `consumeFlashMessagesMiddleware` (see `ihp/IHP/Controller/Render.hs:126`), which reads flash messages from the session during the same request. Both `setErrorMessage` and the middleware operate on the same session IORef via the vault, so the write is visible to the read within the same request. The error message will display immediately.

PR #2458, opened 2026-02-19
Recreate job notification triggers after PGListener reconnects

mpscholten:
## Summary

- Adds a reconnection callback mechanism to `PGListener` that fires after connection recovery, giving callers the live `Hasql.Connection` to run SQL directly
- Job queue registers a callback to recreate its `pg_notify` triggers on reconnect, fixing ~40s delays after `make db` drops the database
- Adds debug logging to the `pg_notify` subscription callback for better observability

## Context

After `make db` drops and recreates the database, the notification triggers that fire `pg_notify` are destroyed. PGListener reconnects and re-LISTENs correctly, but without triggers `pg_notify` never fires. Jobs are only picked up by the 60-second poller (average ~30-40s delay).

Fixes: https://github.com/amitaibu/ihp-sensors/issues/18

## Test plan

- [x] Type-checks: `ihp-pglistener/IHP/PGListener.hs` and `ihp/IHP/Job/Queue.hs`
- [x] All 361 IDE tests pass
- [ ] Manual: run `make db` while job worker is running, verify jobs are picked up immediately via `pg_notify` (not delayed by poller)

### DIFF
--- a/ihp/IHP/Job/Queue.hs
+++ b/ihp/IHP/Job/Queue.hs
@@ -118,8 +118,10 @@ watchForJob pool pgListener tableName pollInterval onNewJob = do
 
         -- Recreate notification triggers when PGListener reconnects (e.g. after `make db` drops the database)
         PGListener.onReconnect (\connection -> do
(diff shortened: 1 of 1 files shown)


amitaibu:
I'll try this PR on my repo

PR #2465, opened 2026-03-01
Make autoRefresh gracefully degrade without PGListener

amitaibu:
## Summary

- When PGListener is not in the request vault (e.g. in hspec tests using `callActionWithParams`), `autoRefresh` now runs the action directly instead of crashing with `error "PGListener not found in request vault"`
- The `autoRefreshMeta` view helper already returns `mempty` when no `AutoRefreshEnabled` is in the vault, so views render correctly in degraded mode
- Adds a test verifying the graceful degradation path

## Test plan

- [ ] Verify the new "graceful degradation without PGListener" test passes
- [ ] Verify existing AutoRefresh tests still pass (renderView query param preservation)
- [ ] Verify `autoRefresh do` controllers work normally in production (with PGListener present)

### DIFF
--- a/ihp-hspec/IHP/Hspec.hs
+++ b/ihp-hspec/IHP/Hspec.hs
@@ -24,6 +24,7 @@ import IHP.Log.Types
 
 import qualified System.Process as Process
 import IHP.Test.Mocking (MockContext(..), runTestMiddlewares)
+import qualified IHP.PGListener as PGListener
 
 withConnection :: ByteString -> (Hasql.Connection -> IO a) -> IO a
 withConnection databaseUrl action = do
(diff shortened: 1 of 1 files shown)


mpscholten:
I wonder whether we should just enable PGListener in tests as well. E.g. what if you have end to end tests using a headless chromium

amitaibu:
Good point! I think both changes are valuable:

1. **Enable PGListener in the test harness** when a DB connection is available — this way e2e tests with headless Chromium get real autoRefresh with WebSocket support.
2. **Keep the graceful degradation** as a safety net — `PGListener.withPGListener` needs a working PostgreSQL connection for LISTEN/NOTIFY, so for unit tests without a real DB (e.g. `mockContextNoDatabase`), PGListener genuinely won't be available.

I'll update the PR to also wire PGListener into the test mocking infrastructure (`withMockContext`, `callActionWithParams`, `initTestApplication`).

PR #2466, opened 2026-03-02
Fix contentDispositionAttachmentAndFileName to use RFC 5987 encoding

andreytroeglazov:
- Replaces the raw filename="..." format with filename*=UTF-8''<percent-encoded> (RFC 5987) in contentDispositionAttachmentAndFileName, preventing SignatureDoesNotMatch errors from S3 when filenames contain spaces, special characters, or non-ASCII bytes.
- Adds hspec tests covering plain ASCII, spaces, special characters, and Unicode filenames.


mpscholten (approved):
Thanks 👍 great!

PR #2491, opened 2026-03-10
Support datetime-local input format in ParamReader and InputValue

amitaibu:
## Summary
- Add `%Y-%m-%dT%H:%M` (datetime-local) and `%Y-%m-%dT%H:%M:%S` (seconds without Z) formats to `ParamReader UTCTime` and `ParamReader LocalTime`, so form submissions from `dateTimeField` are parsed correctly
- Add `%Y-%m-%dT%H:%M:%S%Q` format to support fractional seconds without Z (e.g. `2020-11-08T12:03:35.123`), as browsers can submit these when `step` is set to sub-second precision
- Introduce `DateTimeValue` typeclass used by `dateTimeField` to format values in `datetime-local` compatible format (`YYYY-MM-DDTHH:MM`), without changing `InputValue UTCTime` (which retains full ISO 8601 precision for other uses)
- Deduplicate date parsing logic into shared `readDateTimeParameter` function
- Add tests for all new formats (both `readParameter` and `readParameterJSON`)

## Problem
HTML `<input type="datetime-local">` submits values like `2020-11-08T12:03` (no seconds, no timezone suffix). The existing `ParamReader UTCTime` only accepted `2020-11-08T12:03:35Z` or `2020-11-08`, causing `dateTimeField` values to fail on form submission with "has to be a valid date and time".

Similarly, `dateTimeField` needed to format values in `datetime-local` compatible format for the browser to display them correctly. This is handled by the new `DateTimeValue` typeclass, keeping `InputValue UTCTime` unchanged for other contexts.

## Test plan
- [x] All existing wai-request-params tests pass (110 examples, 0 failures)
- [x] New tests for `datetime-local` format, seconds-without-Z, and fractional-seconds-without-Z parsing added for both `UTCTime` and `LocalTime`
- [x] Verified in a real IHP app that `dateTimeField` now displays defaults and round-trips correctly

### DIFF
--- a/wai-request-params/Test/Wai/Request/ParamsSpec.hs
+++ b/wai-request-params/Test/Wai/Request/ParamsSpec.hs
@@ -280,6 +280,9 @@ spec = do
                 it "should accept datetime-local format with seconds (no Z)" $ do
                     (tshow (readParameter @UTCTime "2020-11-08T12:03:35")) `shouldBe` ("Right 2020-11-08 12:03:35 UTC")
 
+                it "should accept fractional seconds without Z" $ do
+                    (tshow (readParameter @UTCTime "2020-11-08T12:03:35.123")) `shouldBe` ("Right 2020-11-08 12:03:35.123 UTC")
+
                 it "should accept dates" $ do
                     (tshow (readParameter @UTCTime "2020-11-08")) `shouldBe` ("Right 2020-11-08 00:00:00 UTC")
 
@@ -295,6 +298,9 @@ spec = do
                 it "should accept JSON datetime-local strings with seconds (no Z)" $ do
                     (tshow (readParameterJSON @UTCTime (json "\"2020-11-08T12:03:35\""))) `shouldBe` ("Right 2020-11-08 12:03:35 UTC")
 
+                it "should accept fractional seconds without Z via JSON" $ do
+                    (tshow (readParameterJSON @UTCTime (json "\"2020-11-08T12:03:35.123\""))) `shouldBe` ("Right 2020-11-08 12:03:35.123 UTC")
+
             describe "LocalTime" $ do
                 it "should accept timestamps" $ do
                     (tshow (readParameter @LocalTime "2020-11-08T12:03:35Z")) `shouldBe` ("Right 2020-11-08 12:03:35")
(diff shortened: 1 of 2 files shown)


amitaibu:
| Before | Before validtion error | After |
| - | - | - |
| image | image | image |

amitaibu:
@mpscholten I think the review was for the older version? As we don't have changes on `ModelSupportSpec`

amitaibu:
Back to draft, I want to ensure it's working on my app, after the changes

amitaibu:
Seems to be working :smile:

mpscholten (commented):
Thanks for the update, the `InputValue` revert looks good!

One suggestion for the `dateTimeField` implementation: instead of the `Text.take 16` string truncation, we could introduce a small `DateTimeValue` typeclass:

```haskell
class DateTimeValue a where
    dateTimeValue :: a -> Text
    default dateTimeValue :: FormatTime a => a -> Text
    dateTimeValue time = cs (formatTime defaultTimeLocale "%Y-%m-%dT%H:%M" time)

instance DateTimeValue UTCTime
instance DateTimeValue LocalTime

instance DateTimeValue a => DateTimeValue (Maybe a) where
    dateTimeValue (Just time) = dateTimeValue time
    dateTimeValue Nothing = ""
```

Then `dateTimeField` becomes:

```haskell
dateTimeField :: forall fieldName model value.
    ( ?formContext :: FormContext model
    , HasField fieldName model value
    , HasField "meta" model MetaBag
    , KnownSymbol fieldName
    , InputValue value
    , DateTimeValue value
    , KnownSymbol (GetModelName model)
    ) => Proxy fieldName -> FormField
dateTimeField alpha =
    (textField alpha)
        { fieldType = DateTimeInput
        , fieldValue = dateTimeValue ((getField @fieldName model) :: value)
        }
```

This has two benefits:

1. **Type safety**: Right now `dateTimeField #isEnabled` compiles fine on a `Bool` field — it just renders nonsense in a datetime-local input. With the `DateTimeValue value` constraint, that becomes a compile error. Only `UTCTime`, `LocalTime`, and their `Maybe` variants are accepted.

2. **No string manipulation**: Uses `formatTime` directly instead of serializing to ISO 8601 and then truncating the string. The formatting is explicit rather than relying on the output of `inputValue` being at least 16 chars with a `T` in the right place.

Also worth considering: `<input type="datetime-local">` can also submit with seconds when `step` is set (e.g. `step="1"` produces `2020-11-08T12:03:35` without `Z`). Might be worth adding `%Y-%m-%dT%H:%M:%S` (no `Z`) to the `ParamReader` as well to cover that case.

PR #2501, opened 2026-03-11
Log IHP source info at startup

amitaibu:
## Summary
- Adds `IHP.IDE.SourceInfo` module that detects whether IHP runs from the nix store or a local checkout
- Logs the source path, git branch, and commit at startup via `Log.info` (always visible, not gated behind `DEBUG=1`)
- Example output: `IHP Source: Local checkout (/home/user/IHP) on branch feature-x at commit abc1234`

## Test plan
- [x] Unit tests for `formatSourceInfo` covering all variants (NixStore, LocalCheckout with/without git info)
- [ ] Manual: run `devenv up` with a local IHP checkout and verify the log line appears


amitaibu:
image

PR #2505, opened 2026-03-12
Add PostgreSQL table inheritance (INHERITS) support

mpscholten:
## Summary

- Adds `inherits :: !(Maybe Text)` field to the `CreateTable` AST type
- Parses the `INHERITS (parent_table)` clause in `CREATE TABLE` statements
- Compiles the `INHERITS` clause back to SQL
- Adds `allColumnsIncludingInherited` helper that recursively resolves inherited columns from parent tables
- Updates all code generation functions (data types, FromRow, Create, Update, hasql statements, etc.) to include inherited columns
- Updates `findForeignKeyConstraint` to also check parent table FK constraints for inherited columns

Supersedes #1998 which targeted the old file structure before the monorepo restructuring.

## Test plan

- [x] Parser test: parse `CREATE TABLE child (...) INHERITS (parent);`
- [x] Compiler test: round-trip INHERITS clause
- [x] SchemaCompiler tests: verify child table code includes inherited columns in FromRow, Table, and Record instances
- [x] All 383 existing + new tests pass

### DIFF
--- a/ihp-schema-compiler/IHP/SchemaCompiler.hs
+++ b/ihp-schema-compiler/IHP/SchemaCompiler.hs
@@ -601,15 +601,13 @@ isRefCol :: (?schema :: Schema) => CreateTable -> Column -> Bool
 isRefCol table column = isJust (findForeignKeyConstraint table column)
 
 -- | Returns the foreign key constraint bound on the given column.
--- For inherited columns, also checks the parent table's constraints.
+-- For inherited columns, recursively checks ancestor table constraints.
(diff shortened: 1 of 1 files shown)


ihp-datasync-typescript/Test/Spec.hs
```diff
@@ -397,7 +397,7 @@ tests = do
                             ]
                         , primaryKeyConstraint = PrimaryKeyConstraint ["id"]
                         , constraints = []
-                        , unlogged = False
+                        , unlogged = False, inherits = Nothing
```
amitaibu: Here and below

```suggestion
                        , unlogged = False
                        , inherits = Nothing
```

PR #2509, opened 2026-03-12
Add app selector to Generate Controller preview

### LINKED ISSUE #1479: Select app other than Web on "generate controller" from schema
If I right-click a table in schema in the admin page and select "generate controller", it'd be nice if I could switch from e.g. the Web to the Admin app there, instead of having to cancel and then go to codegen to do the same thing :)

mpscholten:
## Summary

- When right-clicking a table in the schema designer and selecting "Generate Controller", the user can now choose which application (Web, Admin, etc.) to generate the controller for
- Previously the preview page hardcoded "Web" with no way to change it (#1479)
- The app selector dropdown appears in the generator-options area alongside the pagination checkbox, and auto-submits to regenerate the preview when changed

## Test plan

- [ ] In a multi-app IHP project, right-click a table in schema designer → "Generate Controller"
- [ ] Verify the application dropdown appears in the preview page
- [ ] Change the selected app and confirm the preview regenerates for that app
- [ ] Single-app projects should not show the dropdown (hidden input fallback)
- [ ] Existing "Generate Controller" flow from the code generators page still works

Fixes #1479

### DIFF
--- a/ihp-ide/IHP/IDE/CodeGen/View/NewController.hs
+++ b/ihp-ide/IHP/IDE/CodeGen/View/NewController.hs
@@ -54,7 +54,7 @@ instance View NewControllerView where
                 <div class="generator-options">
                     <form method="POST" action={NewControllerAction}>
                         <input type="hidden" name="name" value={controllerName}/>
-                        <input type="hidden" name="applicationName" value={applicationName}/>
+                        {renderApplicationNameSelector}
                         <input
                             type="checkbox"
                             name="pagination"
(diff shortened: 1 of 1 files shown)


amitaibu:
Out of curiosity, why do you prefer to separate the Admin and Web -- instead of having different User roles?

mpscholten:
Useful when e.g. the admin backend has a different layout. E.g. I have an app where Web is mostly a react frontend and Admin is a sever rendered "normal" app

PR #2523, opened 2026-03-13
Enable -fspec-eval and -fspec-eval-dict-fun across all packages

mpscholten:
## Summary
- Add `-fspec-eval` and `-fspec-eval-dict-fun` GHC flags to all 26 IHP library cabal files
- Add flags to production build options in `NixSupport/default.nix` (prodGhcOptions, models package, app-lib package) and `Makefile.dist`
- Regenerated nix files from cabal via `./update-nix-from-cabal.sh`

These flags improve runtime performance by specializing functions on evaluated arguments and dictionary function arguments.

## Test plan
- [ ] Verify IHP compiles with the new flags
- [ ] Run test suite
- [ ] Benchmark runtime performance improvement

### DIFF
--- a/NixSupport/default.nix
+++ b/NixSupport/default.nix
@@ -109,7 +109,7 @@ CABAL_EOF
         DuplicateRecordFields
         OverloadedLabels
         OverloadedRecordDot
-    ghc-options: -Wno-unused-imports -Wno-dodgy-imports -Wno-unused-matches -fspec-eval -fspec-eval-dict-fun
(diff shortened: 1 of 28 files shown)


mpscholten:
Closing: -fspec-eval and -fspec-eval-dictfun are already enabled by default at all optimization levels in GHC 9.10 ([DynFlags.hs#L1305](https://gitlab.haskell.org/ghc/ghc/-/blob/ghc-9.10/compiler/GHC/Driver/DynFlags.hs#L1305)). Adding them explicitly is a no-op. The 8.2% latency improvement in the benchmark was noise (core size was identical).

PR #2557, opened 2026-03-20
Fix duplicate HasField "id" when table has composite PK and id column

### LINKED ISSUE #989: Multiple primary keys creates unhelpful generated code errors
IHP's UI does little to stop you from creating a table with two primary keys.

Using the UI I was able to make [this schema](https://gist.github.com/DrewFenwick/093a8e2cd7d69c45f2155f5ba49ef3e6),
which caused the following errors to show up in HLS

```hs

mpscholten:
## Summary

- Fixes #989: When a table has a composite primary key AND a non-PK column named `id`, the SchemaCompiler generated a duplicate `HasField "id"` instance causing a compile error
- `needsHasFieldId` now checks whether the table already has a column named `id` and skips generation in that case (GHC already generates `HasField "id"` from the record field)
- Added 3 test cases covering the conflict scenario, single non-id PK with id column, and preserved existing behavior for composite PKs without an id column

## Test plan

- [x] All 392 existing tests pass
- [ ] CI passes (`nix flake check --impure`)

### DIFF
--- a/ihp-ide/Test/SchemaCompilerSpec.hs
+++ b/ihp-ide/Test/SchemaCompilerSpec.hs
@@ -770,6 +770,53 @@ tests = do
                         {-# INLINE filterWhereId #-}
                     |]
 
+        describe "needsHasFieldId" do
+            let
+                isNamedTable :: Text -> Statement -> Bool
+                isNamedTable targetName (StatementCreateTable CreateTable { name }) = name == targetName
+                isNamedTable _ _ = False
+            it "should not generate HasField id for composite PK table with an id column" do
+                let statements = parseSqlStatements [trimming|
+                    CREATE TABLE ideas_votes (
+                        id INT NOT NULL,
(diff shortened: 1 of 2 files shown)


mpscholten:
Re: the codex-connector observation about `deleteRecord`/`modelFormAction` not working for tables with a non-PK `id` column:

This is a valid observation but not a regression — before this fix, these tables didn't compile at all due to the duplicate `HasField "id"` instance. After this fix, they compile successfully and the `id` column works as a normal field.

For the single non-id PK case (e.g. `code TEXT PRIMARY KEY` with a separate `id` column): the generated `HasField "id"` returning `Id'` was never correct for these tables since `id` isn't the primary key — it was just generated and immediately conflicted with GHC's record-field instance.

Fully supporting non-PK `id` columns with all framework helpers would be a separate enhancement, orthogonal to this bugfix.

PR #2562, opened 2026-03-20
Add OpenAPI support for AutoRoute JSON views

vcombey:
I needed json routes for a cli using my ihp app. I wanted to be fully generated and typed from the definition.
Was more easy that what I thought to add support for it in ihp, with a typedsql method and associated type in Views.

What do you think @mpscholten ?

## Summary
- add OpenAPI 3 generation for pure `AutoRoute` JSON views via `documentRoute`
- derive response schemas from typed `View` JSON payloads and fail fast when documented JSON output drifts from the implementation
- add optional Swagger UI routes backed by the generated OpenAPI document
- sync the Nix derivation with the new `openapi3` dependency so app `devenv up` builds the forked `ihp` package

## Details
- add `JsonResponse` / `jsonTyped` support on `View`
- add `IHP.OpenApiSupport` generation for documented `AutoRoute` controllers only
- emit `components.schemas` for referenced response/parameter schemas
- omit `customRoutes` / `customPathTo` endpoints from the generated spec to avoid drift
- add runtime validation for documented JSON views
- add `swaggerUi` / `swaggerUiWithOptions` helpers to serve Swagger UI and the generated spec from the mounted front controller
- document the new routing / JSON view workflow

### DIFF
--- a/Guide/json-api.markdown
+++ b/Guide/json-api.markdown
@@ -412,7 +412,9 @@ instance View ShowView where
     |]
 
 instance JsonView ShowView where
-    json ShowView { post } = toJSON post
+    type JsonResponse ShowView = Post
+
+    jsonTyped ShowView { post } = post
 ```
 
 When a browser requests the page (sending `Accept: text/html`), it gets the HTML response. When an API client requests it with `Accept: application/json`, it gets JSON. You can test this:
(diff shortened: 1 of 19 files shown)


mpscholten:
👍 like the general idea of this, but don't want to merge larger changes for the next days so we can stabilize for ihp v1.5 release

PR #2570, opened 2026-03-22
Fix Flatpickr timezone shift on datetime-local fields

### LINKED ISSUE #2565: Broken Flatpickr integration
This is the summary of the situation.

Old IHP behavior:
  - not really correct for native datetime-local
  - because it rendered zoned/ISO-style timestamps for UTCTime
  - but that matched IHP’s Flatpickr config using dateFormat: 'Z'
  - so UTC round-tripping worked

New IHP behavior:
  - correct for native datetime-local
  - because it renders YYYY-MM-DDTHH:MM with no timezone
  - but that no longer matches Flatpickr dateFormat: 'Z'
  - so the UTC/local distinction is lost on edit, and round-tripping breaks

So the regression is really: IHP fixed dateTimeField toward HTML standard semantics, but did not update the Flatpickr integration and
overall API semantics to match.

What happens now is:
Flatpickr gets a UTC timestamp from the database as a plain value (without timezone information) and displays it as local time. When saving, it still converts from local time back to UTC. As a result, if a user opens a record for editing, makes no changes, and saves it, the stored time is shifted by the user’s timezone offset.

mpscholten:
## Summary

- Fixes the Flatpickr datetime shift bug where UTC values were shifted by the user's timezone offset on every save
- Keeps `dateFormat: 'Z'` so Flatpickr displays times in the user's local timezone and submits in UTC
- Appends `Z` to the input value in JS before Flatpickr init, bridging the gap between the HTML `datetime-local` spec (no Z) and Flatpickr's UTC mode

**Root cause:** PR #2491 changed `dateTimeField` to render `UTCTime` as `YYYY-MM-DDTHH:MM` (no `Z` suffix) to match the HTML5 `datetime-local` spec, but the Flatpickr config still used `dateFormat: 'Z'` which expects a Z-suffixed value to know it's UTC. Without the Z, Flatpickr interpreted the UTC wall-clock time as local time.

**Why not just change `dateFormat`?** Removing `dateFormat: 'Z'` would fix the round-trip bug, but Flatpickr would then display raw UTC times instead of the user's local timezone. The JS-side Z append preserves both: spec-compliant HTML values for the no-JS fallback, and correct UTC→local display in Flatpickr.

Fixes #2565

## Test plan

- [ ] Open a record with a `UTCTime` datetime field, make no changes, save — verify the time is unchanged
- [ ] Verify Flatpickr displays times in the user's local timezone
- [ ] Verify users who set `data-date-format` explicitly are unaffected (the override check is preserved)
- [ ] Disable JS and verify the native `datetime-local` input still shows the correct value

### DIFF
--- a/ihp/data/static/helpers.js
+++ b/ihp/data/static/helpers.js
@@ -398,10 +398,14 @@ function initDatePicker() {
     });
 
     document.querySelectorAll("input[type='datetime-local']").forEach(el => {
+        // Append Z so Flatpickr knows the server value is UTC
+        if (el.value && !el.value.endsWith('Z')) {
+            el.value = el.value + 'Z';
+        }
         flatpickr(el, {
             ...(el.dataset.enableTime ? {} : { enableTime: true }),
             ...(el.dataset.time_24hr ? {} : { time_24hr: true }),
-            ...(el.dataset.dateFormat ? {} : { dateFormat: 'Y-m-d\\TH:i' }),
+            ...(el.dataset.dateFormat ? {} : { dateFormat: 'Z' }),
             ...(el.dataset.altFormat ? {} : { altFormat: 'd.m.y, H:i' }),
             ...(el.dataset.altInput ? {} : { altInput: true }),
         });


avitkauskas:
Will this make Flatpickr work with UTC time zone: it will get the UTC wall-time without the time zone and will save the value that user sets as UTC time zone. So the functionality that Flatpicker uses user time zone by default will be lost? So, if you want time zones handled as before, you'll need to set the 'Z' option for Flatpicker yourself via addtional params - and this will work as before, also being not datetime-local standard compatible (as it was). Correct?

mpscholten:
@avitkauskas Good catch — you're right that the previous approach would have lost local timezone display.

I've updated the PR to keep `dateFormat: 'Z'` (so Flatpickr still displays in local time and submits in UTC) and instead append `Z` to the input value in JS before Flatpickr initializes. This bridges the gap between the HTML `datetime-local` spec (which requires no Z suffix for the native input to work) and Flatpickr's UTC mode (which needs the Z to know the value is UTC).

So the behavior is now:
- **Server renders** `value="2026-03-22T14:30"` (no Z, spec-compliant for native fallback)
- **JS appends Z** before Flatpickr init → Flatpickr sees `2026-03-22T14:30Z`
- **Flatpickr displays** in the user's local timezone (as before)
- **On save**, Flatpickr submits UTC → correct round-trip, no shift

avitkauskas:
This did not solve the problem. I think #2491 has to be reviewed to check again what problem it was solving and then see if the same problem can be solved without breaking the Flatpickr functionality of displaying in local time and saving in UTC. I don't even see how the datetime-local is used directly with dateTimeField - in my understanding dateTimeField always uses Flatpickr, no?

Now it happens that if time in db is 14:00+02:00, Flatpickr field value somehow is set to 10:00Z (not 12:00Z as it should be) and then Flatpickr displays 12:00 local time.

avitkauskas:
Well, I hope I understood that you can use just datetime-local native field if you remove Flatpickr entirely from your project, yes? Maybe this is the way if it let's you achieve the same: display local times and save UTC?

mpscholten:
tried another fix 👍

avitkauskas:
I'll test it when it is merged. But I have one more related question. What do you usually do when you want just to display the datetime for the users in UI in their local time? In my project I have the following JS for that:

```js
function formatLocalISO(date) {
    const pad = (n) => String(n).padStart(2, '0');

    const year = date.getFullYear();
    const month = pad(date.getMonth() + 1);
    const day = pad(date.getDate());

    const hours = pad(date.getHours());
    const minutes = pad(date.getMinutes());
    const seconds = pad(date.getSeconds());

    return `${year}-${month}-${day} ${hours}:${minutes}:${seconds}`;
}

function localizeTimes(root = document) {
    root.querySelectorAll(".local-time").forEach(el => {
        if (el.dataset.localized) return;

        let timeStr = el.dataset.time;
        timeStr = timeStr.replace(" UTC", "Z").replace(" ", "T");

        const date = new Date(timeStr);
        if (isNaN(date)) return; // safety guard

        el.textContent = formatLocalISO(date);
        el.dataset.localized = "true";
    });
}
```

Then I have a helper:

```haskell
renderTime :: UTCTime -> Html
renderTime time =
    [hsx|
        <span class="local-time" data-time={tshow time}></span>
    |]
```

If we make a local time default in input fields via the usage of the datetime-local, then should the display of all the times in IHP be also in local time by default and provided by the framework itself? Could all these things be solved together?

Additional aspect is the formatting of the local time. One option is to leave that to the user locale. But sometimes you just want to enforce the ISO standard formatting for all the users. It would be good that IHP allow this choice for developer.

And what about date fields? Flatpickr will still be used? Maybe we should refuse Flatpickr at all?

PR #2619, opened 2026-04-07
Add isEmpty helper for Html/MarkupM

amitaibu:
## Summary
- Adds `isEmpty :: MarkupM a -> Bool` to `IHP.HSX.Markup` — renders the builder and checks if the result is zero-length
- Adds `IsEmpty (MarkupM a)` instance in `IHP.HaskellSupport` so IHP app code can use the existing `IsEmpty` class with `Html` values
- Adds 6 test cases in a new `MarkupSpec` module

Closes #2618

## Test plan
- [x] `MarkupSpec` tests pass (mempty, empty string, non-empty, raw bytestring, concatenation cases)
- [ ] `nix flake check --impure` passes


mpscholten (approved):
Thanks 👍

PR #2621, opened 2026-04-07
Disallow SELECT * in typedSql by default

mpscholten:
## Summary

- `SELECT *` and `SELECT table.*` in `[typedSql| ... |]` now produce a compile-time error by default, because `*` is expanded at compile time and can cause runtime decoder failures if the production schema differs
- The error message suggests the exact column names to use instead
- A new `[typedSqlStar| ... |]` quasiquoter is provided as an explicit opt-in for users who understand the risk
- Added 8 new tests (6 unit tests for `detectStarSelects`, 2 compile-fail tests)

## Test plan

- [x] All 90 existing tests pass (0 failures)
- [x] New `detectStarSelects` unit tests verify detection of `SELECT *`, `SELECT table.*`, `SELECT alias.*`
- [x] Correctly ignores `COUNT(*)`, explicit column lists, and composite expansion `(expr).*`
- [x] Compile-fail tests verify that `[typedSql| SELECT * ... |]` and `[typedSql| SELECT table.* ... |]` produce errors
- [x] Existing `table.*` tests updated to use `typedSqlStar` and still pass

### DIFF
--- a/Guide/typed-sql.markdown
+++ b/Guide/typed-sql.markdown
@@ -60,12 +60,28 @@ forEach rows \(itemId, name, views) -> do
 
 Primary key columns are automatically typed as `Id' "table_name"` rather than raw `UUID`.
 
-## Selecting All Columns (`table.*`)
+## Selecting All Columns
 
-Use `table.*` to select all columns from a table, which returns the model type directly:
+### Why `SELECT *` is disallowed by default
+
+`SELECT *` and `SELECT table.*` are not allowed in `typedSql` by default. At compile time, `*` is expanded to whatever columns exist in the development database and a decoder is built for those exact columns. If the production database has a different schema (e.g., a migration added or removed a column), the query will return different columns than the decoder expects, causing a runtime error.
+
+Instead, list columns explicitly:
 
 ```haskell
 items <- sqlQueryTyped [typedSql|
+    SELECT id, name, views FROM items ORDER BY name
+|]
+```
+
+The compile error message will suggest the exact column names to use.
+
+### Opting in with `typedSqlStar`
+
+If you understand the risk and want to use `table.*` anyway (e.g., during rapid prototyping), use the `typedSqlStar` quasiquoter:
+
+```haskell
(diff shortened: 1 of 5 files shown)


vcombey:
I had this problem, that is anoying.
But is there not better mitigations ?
1. It seems typedSql SELECT my_table.* base the order from the schema order, which is often wrong. As schema migration always happens the new column at the end, even if it is declared at the middle in the schema.sql.
2. What about INSERT ... RETURNING *, seems we have the same issue.

So isnt the correct fix to make sure that Schema.sql column order doesnt drift from the runtime column order ?

mpscholten:
> So isnt the correct fix to make sure that Schema.sql column order doesnt drift from the runtime column order ?

It cannot really be prevented when you have a dev db, a staging db and prod db. Eventually some migrations will run in a different order on production than on development.

An alternative solution could be to replace all `*` with an explicit column list based on the Schema.sql with which the app is compiled.

 So when I write `SELECT * from table` it will rewrite to `SELECT col1, col2 FROM table`.

vcombey:
> It cannot really be prevented when you have a dev db, a staging db and prod db. Eventually some migrations will run in a different order on production than on development.

ihp could refuse to generate migrations if column is not added at the end in the Schema.sql ?
and refuse to run a migration if any open migration revision is lower than the max already-applied revision ?
Seems fragile and cumbersome anyway.

vcombey:
there is also the same issue with:

```
_ <- sqlExecTyped [typedSql|
    INSERT INTO typed_sql_order_tests
    VALUES (${valuesId}, ${("values-first" :: Text)}, ${("values-second" :: Text)})
|]

```

vcombey:
that should be, to be safe:

```
_ <- sqlExecTyped [typedSql|
    INSERT INTO typed_sql_order_tests (id, first_name, second_name)
    VALUES (${valuesId}, ${("values-first" :: Text)}, ${("values-second" :: Text)})
|]
```

PR #2622, opened 2026-04-08
Force Set-Cookie flush for deferred clientsession store

Montmorency:
Ran into this issue which was a bit tricky to debug setting up an OAUTH callback loop for authenticating again a client user identity pool.

wai-session-clientsession-deferred only emits Set-Cookie when the session vault is touched during a request. Actions that just render a view without any IO never mark the session dirty, which breaks Azure SSO and similar OAuth flows: the user lands on a no-IO view, no Set-Cookie is sent, and the next navigation bounces back through reauthentication.

Add Session.forceSessionCookie which marks the session dirty via a namespaced __ihp_flush sentinel key, and call it once from runAction so every controller response carries the cookie regardless of what the action does.

### DIFF
--- a/ihp/IHP/Controller/Session.hs
+++ b/ihp/IHP/Controller/Session.hs
@@ -26,6 +26,7 @@ module IHP.Controller.Session
   , getSessionAndClear
   , sessionVaultKey
   , lookupSessionVault
+  , forceSessionCookie
   ) where
 
 import Prelude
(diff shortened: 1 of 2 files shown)


mpscholten:
I can't imagine a case where the missing Set-Cookie header really breaks an oauth flow?

Montmorency:
So the only work around I've found so far is to run IO in certain actions where the views were rendering pure html with no db round trip. 

There was only one view in application where the OAUTH loop was triggered unnecessarily. The action was triggered and then the OAUTH loop was retriggered, however the user didn't need to reauthenticate because the session cookie was already set so they get redirected to landing page. 

Adding db IO to the action stopped the extra redirect. Could that have forced something?

Montmorency:
Found better diagnosis with claude:

  IHP only freezes the controller context inside renderHtml. Code in the action body —
  beforeAction = ensureAzureUser, WelcomeAction, AzureLoginAction — always sees the unfrozen
  flavor (the IORef-backed ControllerContext {} arm), and our function returned Nothing for it.
  So ensureAzureUser thought no one was logged in and bounced every authenticated user back
  through /AzureLogin. The session cookie was perfectly fine; we just never read the value out of
   the unfrozen context.

  Why it looked IO-correlated and we suspected wai-defer: actions that did a fetch before
  respondHtml happened to work, while pure-render actions looped. We pattern-matched on "no DB IO
   → broken", which pointed at deferred-cookie writes. In reality the working actions just didn't
   call currentAzureUserOrNothing from the action body — they only consulted it inside view
  rendering, where renderHtml had already frozen the context, so the FrozenControllerContext arm
  fired and the lookup succeeded. Pure coincidence with the IO presence.

mpscholten:
I'd recommend using the request vault instead of the controller context, it's a bit simpler because it doesnt have the frozen/unfrozen state

Montmorency:
Will close as this was issue at app level implementation.

PR #2626, opened 2026-04-10
Drop lens + wreq from ihp core library

mpscholten:
## Summary

Replace the two trivial call sites in `ihp` that used `wreq` (and `lens`, which was only imported to destructure `wreq` responses) with direct `http-client` calls, and drop both packages from `ihp`'s library deps.

- `IHP/Server.hs` `heartbeatCheck` — localhost `/_healthz` probe for `warp-systemd`
- `IHP/FileStorage/ControllerFunctions.hs` `storeFileFromUrl`

Both rewritten on top of `http-client` / `http-client-tls`, which are already listed in `ihp.cabal`. Both set `checkResponse = HTTP.throwErrorStatusCodes` on the request so non-2xx responses still raise an exception, matching the original `Wreq.get` semantics — a 500 from `/_healthz` fails the warp-systemd heartbeat, a 404 from the upstream URL doesn't silently become the stored "file". (Both fixes landed in `8de78917` after Codex P1 review comments.)

## Closure impact (measured)

Walked `propagatedBuildInputs` against IHP's pinned nixpkgs (`nixos-25.11`). **8 transitive Haskell packages** leave the production `ihp` closure, totalling about **81 MB** of self-size:

| Package | Self-size |
|---|---|
| lens | 46.0 MB |
| haskell-lexer | 9.9 MB |
| crypto-pubkey-types | 7.2 MB |
| wreq | 6.3 MB |
| reflection | 4.0 MB |
| authenticate-oauth | 2.8 MB |
| lens-aeson | 2.1 MB |
| RSA | 1.5 MB |
| parallel | 1.1 MB |
| **Total** | **~81 MB** |

Things that do NOT leave the closure despite being pulled in by `lens`/`wreq`: `profunctors`, `kan-extensions`, `adjunctions`, `distributive`, `bifunctors`, `comonad`, `semigroupoids`, etc. — they're pulled in by `conduit-extra`, `classy-prelude`, `postgresql-simple`, `aeson`, and the other kept deps.

## What's NOT in this PR

The original plan also tried to:
1. Drop `basic-prelude`, `haskell-src-exts`, `haskell-src-meta` as "dead deps". `basic-prelude` turned out to actually be used — `IHP/Prelude.hs:59` imports `CorePrelude`, which is a module exported by the `basic-prelude` package (name mismatch I missed on first grep). ghci found it because `devenv-module.nix` puts `basic-prelude` in the dev-shell package db, but the cabal build correctly rejected the drop. For `haskell-src-exts` + `haskell-src-meta`, no importers exist, but the closure walk showed they remain pulled in transitively via `interpolate → haskell-src-meta → haskell-src-exts`, so dropping them would be hygiene-only with zero closure impact. All three restored in `56f89ae9`.
2. Move `IHP.Test.Mocking` from `ihp` to `ihp-hspec` so `hspec` could drop from `ihp`'s library deps (~38 MB additional closure shrink). Creates a cyclic derivation dep at nix eval time: `ihp.testHaskellDepends → ihp-hspec → ihp` (cabal2nix merges library + test components into one nix derivation). Reverted in `b271d646`. The right way to land this follow-up is to move the 6 tests that use `IHP.Test.Mocking` (`MockingSpec`, `AutoRefreshSpec`, `RouterSupportSpec`, `Controller.NotFoundSpec`, `Controller.AccessDeniedSpec`, `ViewSupportSpec`) into a new test-suite inside `ihp-hspec`.

## API impact

None. The source-level rewrites are transparent to users.

## Test plan

- [x] `echo ':l ihp/IHP/Server.hs' | ghci` — clean
- [x] `echo ':l ihp/IHP/FileStorage/ControllerFunctions.hs' | ghci` — clean
- [x] `nix build .#default -L` — exit 0 locally
- [ ] `nix flake check --impure` — CI
- [ ] Manual: hit `/_healthz` on a running IHP app returning 500, confirm warp-systemd heartbeat fails (via `throwErrorStatusCodes`)
- [ ] Manual: call `storeFileFromUrl` against a URL returning 404, confirm it raises instead of silently storing the error page


unhammer:
… one day hopefully wreq will switch to microlens (or even nothing)

PR #2630, opened 2026-04-12
Wrap job execution in a database transaction for automatic crash recovery

mpscholten:
## Summary

- Wraps the entire job lifecycle (fetch → perform → result) in a PostgreSQL transaction on a dedicated raw connection
- If the worker process crashes, the connection drops, PostgreSQL auto-rolls back, and the job immediately reverts to `job_status_not_started` — no more waiting for stale recovery timers
- Replaces the old pool-based `fetchNextJob`/`jobDidSucceed`/`jobDidFail` with `withNextJob`/`jobDidSucceedConn`/`jobDidFailConn` that operate within the open transaction

## How it works

```
1. Open dedicated connection (not from app pool)
2. BEGIN
3. UPDATE status='running' ... FOR UPDATE SKIP LOCKED  (atomic fetch+lock)
4. perform job  (uses app's ModelContext pool — separate connections)
5. UPDATE status='succeeded/failed' + COMMIT
6. Release connection

On crash: connection drops → auto-ROLLBACK → job reverts to not_started
```

## Test plan

- [x] All 465 existing tests pass (0 failures)
- [x] Type-checks cleanly via ghci
- [ ] Integration test updated to use `withNextJob` pattern
- [ ] Manual crash test: enqueue job, kill -9 worker, verify job immediately available

### DIFF
--- a/ihp/IHP/Job/Queue.hs
+++ b/ihp/IHP/Job/Queue.hs
@@ -20,7 +20,7 @@ module IHP.Job.Queue
 ) where
 
 import IHP.Job.Queue.Pool (runPool)
-import IHP.Job.Queue.Fetch (withNextJob, pendingJobConditionSQL)
+import IHP.Job.Queue.Fetch (withNextJob, pendingJobConditionSQL, advisoryLockKey)
 import IHP.Job.Queue.Watch
     ( watchForJob
     , watchForJobWithPollerTriggerRepair
--- a/ihp/IHP/Job/Queue.hs
+++ b/ihp/IHP/Job/Queue.hs
@@ -20,7 +20,7 @@ module IHP.Job.Queue
 ) where
 
 import IHP.Job.Queue.Pool (runPool)
-import IHP.Job.Queue.Fetch (withNextJob, pendingJobConditionSQL, advisoryLockKey)
+import IHP.Job.Queue.Fetch (withNextJob, pendingJobConditionSQL)
 import IHP.Job.Queue.Watch
     ( watchForJob
     , watchForJobWithPollerTriggerRepair
--- a/ihp/IHP/Job/Queue.hs
+++ b/ihp/IHP/Job/Queue.hs
@@ -1,6 +1,6 @@
 module IHP.Job.Queue
 ( runPool
-, fetchNextJob
+, withNextJob
 , pendingJobConditionSQL
 , watchForJob
 , watchForJobWithPollerTriggerRepair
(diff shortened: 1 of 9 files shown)


mpscholten:
Both review comments addressed in 9586ad1d:

**P1 (row lock during perform):** The fetch UPDATE is now committed immediately, releasing the row lock. `perform` can freely update the job row without deadlocking.

**P2 (running status not visible):** Since the fetch is committed right away, `status='running'` is visible to `recoverStaleJobs` and monitoring queries.

**Crash recovery:** A session-level `pg_advisory_lock` on the dedicated connection replaces the transaction-held row lock as the crash detector. When the worker process dies, the connection drops and the advisory lock is released instantly. `recoverStaleJobs` (which still runs periodically) handles resetting stale `running` jobs.

mpscholten:
## Closing — surveyed how other Postgres queues handle this

After the review feedback exposed real concerns with holding the transaction across `perform`, I surveyed 10 production Postgres job queues to see what the field has converged on:

**Rust:** pgmq, Apalis, sqlxmq, Underway · **Go:** River · **Elixir:** Oban · **Node:** Graphile Worker, pg-boss · **Ruby:** Que · **Python:** Procrastinate

### The 2025 consensus (9 of 10)

```sql
WITH locked AS (
  SELECT id FROM jobs
  WHERE state='available' AND scheduled_at <= now()
  ORDER BY priority, scheduled_at, id
  LIMIT N FOR UPDATE SKIP LOCKED
)
UPDATE jobs SET state='running', attempted_at=now()
FROM locked WHERE jobs.id = locked.id RETURNING *;
-- COMMIT, execute job OUTSIDE any transaction, then short tx for result
```

**No library except Que holds a transaction or row lock during job execution.** Recovery is timeout-based (River, Oban, Graphile, pg-boss, pgmq, sqlxmq) or heartbeat-based (Underway, Procrastinate, Apalis).

### The one outlier: Que

Que uses session-level `pg_advisory_lock` (not transactions). When the worker connection dies, Postgres releases the lock automatically. A sweeper checks `pg_stat_activity` for dead worker PIDs. **Trade-off:** requires a dedicated connection per worker (can't pool). Nobody else copies this.

### Where IHP sits

The IHP code **before this PR** is already the textbook consensus pattern:
- `FOR UPDATE SKIP LOCKED` ✓
- Commit immediately, execute outside transaction ✓
- Timeout-based `recoverStaleJobs` ✓
- Pool connections ✓
- LISTEN/NOTIFY wake-ups ✓

**This PR's design (transaction held during execution) matches no surveyed library.** The reviewer's two concerns were exactly the reasons nobody else does this. The "5 minute stale recovery delay" we were trying to eliminate is the trade-off every other library accepts for not holding transactions.

### Closing

Closing this PR. If instant crash recovery genuinely matters in the future, the **Que pattern (session advisory lock + `pg_stat_activity` sweeper)** is the only proven design — but that's a different PR (dedicated connections per worker, real architectural change).

The genuinely useful improvements from the original Postgres article — **partial index on pending jobs, completed-job cleanup, autovacuum tuning** — are orthogonal to the locking model and would be small focused PRs.

PR #2638, opened 2026-04-14
Re-add ConvertibleStrings Text/String Html instances

amitaibu:
## Summary

- When HSX moved from Blaze to the direct `ByteString Builder` backend (e5c6fbd1), `Html` became a type alias for `IHP.HSX.Markup.Markup` instead of `Text.Blaze.Html5.Html`. The old Blaze `ConvertibleStrings` instances in `ihp-hsx/blaze/IHP/HSX/ConvertibleStrings.hs` still exist, but they target the now-unused Blaze `Html5.Html` type, so `cs (someText :: Text) :: Html` no longer resolves.
- Re-adds `instance ConvertibleStrings Text Markup` and `instance ConvertibleStrings String Markup` on the direct backend, mirroring Blaze's semantics (HTML-escape then output).
- Adds tests covering both happy-path conversion and HTML escaping of special characters (`<`, `>`, `&`).

## Note on `blaze/IHP/HSX/ConvertibleStrings.hs`

That file still has instances, but they target `Text.Blaze.Html5.Html` / `AttributeValue` — the old Blaze types, which are no longer what a user's `Html` resolves to. It's still transitively imported via `IHP.HSX.ToHtml` (which also uses the old `Html5.Html`) and a handful of `()` imports in `ihp/IHP/View/Form/*.hs` and `ihp/IHP/View/TimeAgo.hs`, but none of those files reference `Text.Blaze` directly anymore. So it's effectively stale — but removing it is a bigger refactor than this fix, and is left out of scope.

## Test plan

- [x] `ihp-hsx` test suite passes (116 examples, 0 failures) including the 4 new `ConvertibleStrings` cases
- [x] `IHP.HSX.Markup` type-checks standalone via `ghci`

### DIFF
--- a/ihp-hsx/direct/IHP/HSX/Markup.hs
+++ b/ihp-hsx/direct/IHP/HSX/Markup.hs
@@ -293,8 +293,6 @@ stringValue = escapeHtml . Text.pack
 {-# INLINE stringValue #-}
 
 -- | Convert 'Text' into HTML-escaped 'Markup' via @cs@.
--- Mirrors Blaze's @instance ConvertibleStrings Text Html@ that was lost
--- when the HSX backend moved from Blaze to the direct ByteString builder.
 instance ConvertibleStrings Text Markup where
     {-# INLINE convertString #-}
     convertString = escapeHtml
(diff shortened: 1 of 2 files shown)


ihp-hsx/direct/IHP/HSX/Markup.hs
```diff
@@ -291,3 +291,15 @@ preEscapedTextValue = Markup . TE.encodeUtf8Builder
 stringValue :: String -> Markup
 stringValue = escapeHtml . Text.pack
 {-# INLINE stringValue #-}
+
+-- | Convert 'Text' into HTML-escaped 'Markup' via @cs@.
+-- Mirrors Blaze's @instance ConvertibleStrings Text Html@ that was lost
+-- when the HSX backend moved from Blaze to the direct ByteString builder.
```
amitaibu:
```suggestion
```

mpscholten (approved):
👍

PR #2639, opened 2026-04-14
Fix withUser/authMiddleware: preserve mock session across sessionMiddleware in tests

amitaibu:
## Summary

- `withUser` + `callAction` against a protected action was returning `302` redirects to the login page instead of running the authenticated action.
- Root cause is a timing mismatch introduced by the move of authentication from `initAuthentication` (controller-context initializer) to the WAI-based `authMiddleware` in `Config`:
  1. `withUser` inserts a mock `Network.Wai.Session.Maybe.Session` at `sessionVaultKey` in the request vault.
  2. `callActionWithParams` runs the request through the full middleware stack from `IHP.Server.initMiddlewareStack`.
  3. That stack includes `sessionMiddleware = withSession store "SESSION" sessionCookie sessionVaultKey` (from `wai-session-maybe`), which *unconditionally* overwrites the vault key with a fresh empty session built from the request cookie — wiping out the mock session.
  4. `authMiddleware @User` then runs, reads the (now empty) session, writes `Nothing` into `currentUserVaultKey`, and the controller's `ensureIsUser` redirects to login.
- Under the old `initAuthentication` model this worked because auth was resolved *after* the middleware stack inside `runActionWithNewContext`, where `callActionWithParams`' inner `controllerApp` still re-inserted the mock session. That path runs too late once auth has moved into the stack.

## Fix

Add an `afterSessionMiddleware :: Middleware` slot to `initMiddlewareStack` that is spliced into the stack immediately after `sessionMiddleware` and before `authMw`:

```haskell
. sessionMiddleware
. afterSessionMiddleware  -- new slot
. approotMiddleware
...
. authMw
```

- Production callers (`IHP.Server.run`, `runTestMiddlewares`, `initTestApplication`, existing test helpers in `AutoRefreshSpec`/`ControllerSupportSpec`) pass `id` — no behavior change.
- `IHP.Test.Mocking.callActionWithParams` captures `mockSession` from `?request` and passes a `preserveMockSession` middleware that re-inserts it into the vault. Now the mock session is present when `authMw` reads it.
- The stale inner re-insertion that used to live in `controllerApp` is removed — it ran after `authMw` and never served its intended purpose after the auth migration.

## Test plan

Regression test added to `ihp/Test/Test/MockingSpec.hs` under the `withUser + AuthMiddleware` describe:

- Configures a custom `AuthMiddleware` via `authMiddlewareWith` that reads `lookupSessionVault` and parses the cereal-encoded UUID written by `withUser`, storing the result at a private vault key.
- A `RequireUserAction` controller reads that vault key and renders the user id (or `"anonymous"`).
- `withUser testUser $ callAction RequireUserAction` → asserts `status200` and that the body is the user's UUID. Without the fix, the mock session is wiped by `sessionMiddleware` and the body is `"anonymous"`.
- A second test asserts that calling `RequireUserAction` *without* `withUser` returns `"anonymous"`, confirming the auth middleware is actually gating.

Full IHP test suite: `487 examples, 0 failures, 23 pending` (pending are PostgreSQL-dependent integration tests with no DB available).

- [x] Run `ihp/Test/Test/Main.hs` via ghci
- [x] New regression tests pass (`propagates the mock session from withUser to authMiddleware`, `returns anonymous when withUser is not used`)
- [x] No regressions in existing tests (`IHP.Test.Mocking`, `Auth Middleware`, `IHP.ControllerSupport`, `IHP.AutoRefresh`)

## Notes / non-goals

- Does **not** revert the auth-middleware migration or touch `initAuthentication` — the deprecation stands.
- Does **not** patch `wai-session-maybe`.
- `sessionVaultKey` and its mock-session type are reused unchanged.

### DIFF
--- a/ihp/IHP/Server.hs
+++ b/ihp/IHP/Server.hs
@@ -63,7 +63,7 @@ run configBuilder = do
                 PGListener.withPGListener frameworkConfig.databaseUrl frameworkConfig.logger \pgListener -> do
                     let ?modelContext = modelContext
 
-                    middleware <- initMiddlewareStack frameworkConfig modelContext (Just pgListener) id
+                    middleware <- initMiddlewareStack frameworkConfig modelContext (Just pgListener)
                     staticApp <- initStaticApp frameworkConfig
                     let requestLoggerMiddleware = frameworkConfig.requestLoggerMiddleware
 
@@ -143,18 +143,12 @@ initCorsMiddleware FrameworkConfig { corsResourcePolicy } = case corsResourcePol
         Just corsResourcePolicy -> Cors.cors (const (Just corsResourcePolicy))
         Nothing -> id
 
--- | Initialize the complete middleware stack.
+-- | Initialize the complete middleware stack
 --
--- Pass 'Nothing' for 'PGListener' in tests, 'Just' in production.
---
--- The @afterSessionMiddleware@ argument is spliced into the stack
--- immediately after 'sessionMiddleware' and before 'authMw'. In production
--- this is 'id'. 'IHP.Test.Mocking.callActionWithParams' uses it to
--- re-inject the mock session that 'withUser' places in the request vault,
--- which would otherwise be wiped out by 'sessionMiddleware' (from
--- wai-session-maybe) before 'authMw' can read it.
-initMiddlewareStack :: FrameworkConfig -> ModelContext -> Maybe PGListener.PGListener -> Middleware -> IO Middleware
-initMiddlewareStack frameworkConfig modelContext maybePgListener afterSessionMiddleware = do
+-- Pass Nothing for PGListener in tests
+-- Pass Just pgListener in production for full functionality
+initMiddlewareStack :: FrameworkConfig -> ModelContext -> Maybe PGListener.PGListener -> IO Middleware
+initMiddlewareStack frameworkConfig modelContext maybePgListener = do
     sessionMiddleware <- initSessionMiddleware frameworkConfig
     approotMiddleware <- Approot.envFallback
     assetPathMiddleware <- AssetPath.assetPathFromEnvMiddleware "IHP_ASSET_VERSION" "IHP_ASSET_BASEURL"
@@ -174,7 +168,6 @@ initMiddlewareStack frameworkConfig modelContext maybePgListener afterSessionMid
         . corsMiddleware
         . methodOverridePost
         . sessionMiddleware
-        . afterSessionMiddleware
         . approotMiddleware
         . viewLayoutMiddleware
         . responseHeadersMiddleware
--- a/ihp/IHP/Server.hs
+++ b/ihp/IHP/Server.hs
@@ -63,7 +63,7 @@ run configBuilder = do
                 PGListener.withPGListener frameworkConfig.databaseUrl frameworkConfig.logger \pgListener -> do
                     let ?modelContext = modelContext
 
-                    middleware <- initMiddlewareStack frameworkConfig modelContext (Just pgListener)
+                    middleware <- initMiddlewareStack frameworkConfig modelContext (Just pgListener) id
                     staticApp <- initStaticApp frameworkConfig
                     let requestLoggerMiddleware = frameworkConfig.requestLoggerMiddleware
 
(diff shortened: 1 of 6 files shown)


mpscholten:
Thanks for the thorough write-up on the root cause. I verified the diagnosis: `wai-session-maybe`'s `withSession` does unconditionally `Vault.insert` the session at `vkey` (see the installed source of `Network.Wai.Session.Maybe.withSession`), so yes, it really does wipe whatever `withUser` seeded.

One red herring to rule out up front: the encoding difference between the new `login` path (raw 36-byte UUID ASCII, `ihp/IHP/LoginSupport/Helper/Controller.hs:125`) and `withUser`'s `Serialize.encode user.id` (8-byte length prefix + 36-byte ASCII = 44 bytes total, via the `Serialize (Id' table)` instance at `ihp/IHP/Controller/Session.hs:152-155`) is already handled by `parseSessionUUID`'s legacy branch (`ihp/IHP/LoginSupport/Middleware.hs:82-90`). So the clobbering is the real issue, not the encoding.

## Suggestion: a fix that doesn't change `initMiddlewareStack`

The cleanest approach I can see avoids threading a `Middleware` param through the stack and instead leans on pure middleware composition: have `withUser` build a middleware and let the test harness apply it **innermost**, after the whole stack (including `authMiddleware`) has run. Last write wins, by construction.

```haskell
-- ihp/IHP/Test/Mocking.hs

{-# NOINLINE mockOverrideVaultKey #-}
mockOverrideVaultKey :: Vault.Key Wai.Middleware
mockOverrideVaultKey = unsafePerformIO Vault.newKey

withUser user callback =
    let ?request = ?request { Wai.vault = newVault } in callback
  where
    userMw app req respond =
        let req' = req
                { Wai.vault
                    = Vault.insert currentUserVaultKey   (Just user)
                    . Vault.insert currentUserIdVaultKey (Just (unpackId user.id))
                    $ Wai.vault req
                }
        in app req' respond

    existingMw = fromMaybe id (Vault.lookup mockOverrideVaultKey (Wai.vault ?request))
    newVault   = Vault.insert mockOverrideVaultKey (existingMw . userMw) (Wai.vault ?request)
```

And in `callActionWithParams`:

```haskell
let overrideMw = fromMaybe id (Vault.lookup mockOverrideVaultKey (Wai.vault ?request))
middlewareStack <- initMiddlewareStack frameworkConfig modelContext pgListener  -- signature unchanged
_ <- middlewareStack (overrideMw controllerApp) baseRequest captureRespond
```

Trace:

1. Stack runs as in production. `sessionMiddleware` clobbers — doesn't matter.
2. `authMiddleware @User` reads the empty session, writes `Nothing` to `currentUserVaultKey`.
3. Stack hands off to `overrideMw controllerApp`.
4. `overrideMw` (= `userMw`) inserts the mock user into `currentUserVaultKey`.
5. Controller sees the mock user.

### Why I think this is cleaner than the current PR

- **Zero production surface.** `IHP.Server` stays untouched. `initMiddlewareStack` keeps its signature — no breaking change for any downstream caller.
- **No invariants pushed into auth middleware.** No \"don't overwrite if set\" guards, no backup vault key read by `authMiddleware`. Auth is unchanged.
- **Session encoding sidestepped entirely** for the test path — no cereal, no mock `Session` tuple.
- **Composes naturally.** `withUser`, `withAdmin`, future `withHeaders`, etc. all stack into `mockOverrideVaultKey` via plain function composition, and the harness applies the whole chain once.
- **Smaller diff.** Roughly `Mocking.hs` only. Revert the `afterSessionMiddleware` parameter, the `preserveMockSession` closure, and the five call-site updates (`Server.run`, `runTestMiddlewares`, `initTestApplication`, `AutoRefreshSpec.callActionWithQueryParams`, `ControllerSupportSpec.withMiddleware`).

The whole `MockingSpec` regression test you added should pass unchanged — it's a good test and worth keeping.

## Follow-up cleanup (separate PR)

Worth tracking independently: remove the `Serialize (Id' table)` session encoding path (`ihp/IHP/Controller/Session.hs:152-155`) and the 44-byte legacy branch in `parseSessionUUID`. The in-source `TODO: Remove old format support after 2026-05-01` is already there.

amitaibu:
@mpscholten ready for review

mpscholten:
The rework reads cleanly — `initMiddlewareStack`'s signature is back to unchanged, the session/cereal machinery is gone from `withUser`, and `callActionWithParams` is now just `middlewareStack (overrideMiddleware controllerApp) baseRequest captureRespond`. Exactly the shape the earlier comment sketched.

A few small things worth a look:

**1. `withUser`'s type got narrowed to `user ~ CurrentUserRecord`.** Previously the helper was polymorphic; now it only works for the app's configured `CurrentUserRecord`. Breaking change for anyone using it with another user type (e.g. seeding an `Admin`). It's the right narrowing for the new implementation (`currentUserVaultKey` is typed `Key (Maybe CurrentUserRecord)`), but we should be explicit about the semantic shift.

**2. No `withAdmin` companion.** With `withUser` locked to `CurrentUserRecord`, apps that use `adminAuthMiddleware @Admin` can't seed an admin in tests. Same mechanism works — just write `currentAdminVaultKey` / `currentAdminIdVaultKey` via the same `mockOverrideVaultKey`:

```haskell
withAdmin :: forall admin result.
    ( ?request :: Request
    , ?respond :: Respond
    , admin ~ CurrentAdminRecord
    , HasField "id" admin (Id admin)
    , PrimaryKey (GetTableName admin) ~ UUID
    ) => admin -> ((?request :: Request, ?respond :: Respond) => IO result) -> IO result
```

Can be a follow-up, but if we ship `withUser` with this constraint without `withAdmin`, admin tests silently have no migration path.

**3. Style nit:** `HasField "id" user (Id' (GetTableName user))` is equivalent to `HasField "id" user (Id user)` since `Id` is the alias — the shorter form matches how `login` spells it (`ihp/IHP/LoginSupport/Helper/Controller.hs:122`).

**Heads-up on #2640:** I merged the `parseSessionUUID` / `Serialize.encode`→`UUID.toASCIIBytes` cleanup separately. It touches `withUser`, so when this lands there'll be a trivial conflict in `IHP.Test.Mocking` — your version wins (the session path is gone entirely in your rework), just re-pin that hunk to the new `withUser` body.

mpscholten:
Ah sorry ignore the prev comment :D it was done by claude, to me (the human) it looks fine 👍

PR #2642, opened 2026-04-15
Document MemoryHigh/MemoryMax for app and worker services

AronNovak:
## Summary

Add a "Memory Limits and Restart Hardening" subsection to the `systemd` Integration section of the deployment guide.

`services.app` and `services.worker` already set `Restart = "always"` and a watchdog, but no cgroup memory bounds. Without `MemoryMax`, a leaking unit can consume all host RAM before the kernel OOM killer intervenes, and the kill may land on unrelated processes (notably a colocated PostgreSQL under `appWithPostgres`) rather than the offending unit.

The new subsection shows how to set `MemoryHigh` / `MemoryMax` / `RestartSec` on both services so an OOM stays inside a single unit's cgroup and gets cleanly restarted by the existing `Restart = "always"` policy. Worker tuning is called out explicitly since long-running job handlers are the most common source of leaks in a long-lived Haskell process.

## Test plan

- [ ] Docs-only change — renders as expected in the Guide

### DIFF
--- a/Guide/deployment.markdown
+++ b/Guide/deployment.markdown
@@ -1613,3 +1613,31 @@ systemctl status migrate
 journalctl -u app -n 50 --no-pager
 ```
 
+### Memory Limits and Restart Hardening
+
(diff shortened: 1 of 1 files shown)


mpscholten (approved):
Great thanks 👍

PR #2645, opened 2026-04-16
Document system.autoUpgrade allowReboot for kernel patching

AronNovak:
## Summary

Adds an `Automatic Upgrades and Kernel Reboots` subsection under `Configuring the Instance` in the deployment guide. Without `allowReboot`, `system.autoUpgrade` installs new kernels into `/boot` but the host keeps running the old one until rebooted manually, so kernel CVEs stay unpatched on long-lived hosts.

## Test plan

- [ ] Docs-only change — renders as expected in the Guide


mpscholten (approved):
Thanks 👍

PR #2651, opened 2026-04-23
Deprecate sqlQuery/sqlExec, add unsafe* variants

mpscholten:
## Summary

AI agents and humans frequently reach for the string-based `sqlQuery` / `sqlExec` family instead of the type-safe `[typedSql| ... |]` quasi quoter. Because the string-based APIs only decode results at runtime, mistakes surface as production crashes rather than compile errors.

This PR keeps the old functions working but nudges callers toward the typed API:

- **Adds `unsafe*` variants** (`unsafeSqlQuery`, `unsafeSqlQuerySingleRow`, `unsafeSqlExec`, `unsafeSqlExecDiscardResult`, `unsafeSqlQueryScalar`, `unsafeSqlQueryScalarOrNothing`) in `IHP.ModelSupport`. These carry the real implementation.
- **Deprecates the old names** — they become thin aliases with `{-# DEPRECATED #-}` pragmas that point users at `[typedSql| ... |]` first (via `sqlQueryTyped`/`sqlExecTyped`) and `unsafeSql*` as an explicit escape hatch.
- **Repoints internal IHP callers** (just the TypedSql test fixtures that need raw DDL) at the `unsafe*` names so the library itself compiles without warnings.
- **Rewrites the "Raw SQL Queries" section of the database guide** to lead with `typedSql` and describe `unsafeSqlQuery` as the escape hatch for cases `typedSql` can't express (dynamic identifiers, DDL without a dev DB, etc.).

## Why

`typedSql` connects to the dev database at compile time and infers parameter and result types from the SQL itself, so schema drift and column mistakes become compile errors. Most callers of `sqlQuery`/`sqlExec` are writing queries that `typedSql` could handle — they just reach for the older API out of habit. A deprecation warning at every call site is the cheapest way to surface the better alternative without breaking anyone.

`-Werror=deprecations` is not enabled in IHP's cabal files, so existing user code continues to build — the warning is visible but non-fatal.

## Example of the warning

```
/path/to/MyController.hs:42:11: warning: [GHC-68441] [-Wdeprecations]
    In the use of ‘sqlQuery’ (imported from IHP.ModelSupport):
    Deprecated: "Use the typed quasi quoter '[typedSql| ... |]' with 'sqlQueryTyped' (from IHP.TypedSql) for compile-time type checking. If you really need untyped raw SQL (e.g. dynamic table names), use 'unsafeSqlQuery' instead."
```

## Test plan

- [x] `ghci` loads `ihp/IHP/ModelSupport.hs` cleanly
- [x] `ghci` loads `ihp-typed-sql/Test/Test/TypedSqlSpec.hs` cleanly (fixtures updated to `unsafeSqlExecDiscardResult`)
- [x] `ghci` loads `ihp-ide/Test/Main.hs` (208 modules) without any deprecation warnings from IHP's own sources
- [x] `ghci` loads `ihp-datasync/IHP/DataSync/RowLevelSecurity.hs` (94 modules) cleanly
- [x] Test compile of an external module confirms `sqlQuery`/`sqlQueryScalar`/`sqlExec` emit `-Wdeprecations` warnings with the migration hint, while `unsafeSqlQuery` is silent
- [ ] CI passes on GitHub Actions

### DIFF
--- a/Guide/database.markdown
+++ b/Guide/database.markdown
@@ -305,80 +305,126 @@ do
 
 The IHP query builder is designed to be able to easily express many basic sql queries. When your application is growing you will typically hit a point where a complex SQL query cannot be easily expressed with the IHP query builder. In that case it's recommended to use handwritten SQL to access your data.
 
-> **Tip:** For compile-time type-checked SQL queries, see the [Typed SQL Guide](typed-sql.html). Typed SQL automatically infers Haskell types from your SQL at compile time, eliminating the need for manual `FromRow` instances.
+> **Recommended:** Use the `[typedSql| ... |]` quasi quoter from [`IHP.TypedSql`](typed-sql.html). It connects to your development database at compile time and infers Haskell types for parameters and results — so mistakes surface as compile errors instead of runtime crashes. `sqlQuery`, `sqlQuerySingleRow`, `sqlQueryScalar`, `sqlQueryScalarOrNothing`, `sqlExec`, and `sqlExecDiscardResult` are **deprecated**; use `typedSql` with `sqlQueryTyped` / `sqlExecTyped`, or fall back to the `unsafeSql*` variants when a truly raw query is required.
 
-Use the function [`sqlQuery`](https://ihp.digitallyinduced.com/api-docs/IHP-ModelSupport.html#v:sqlQuery) to run a raw SQL query:
+### Typed queries (preferred)
+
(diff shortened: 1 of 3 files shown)


amitaibu:
Maybe rename from `untyped` to `unsafe` ?

mpscholten:
👍 yeah after thinking about it more I like `unsafe` more (initially it was unsafe, then went with untyped)

PR #2654, opened 2026-04-24
Fix broken link in Guide

joaompfe:


joaompfe:
But then the original URL is again referenced in https://ihp.digitallyinduced.com/Guide/hsx.html on the first sentence of the Introduction...

mpscholten:
thx, I guess it might be better to set up a redirect on the server for this 👍

mpscholten:
Update from my end: I went down the server-side-redirect path mentioned above and ended up reverting it — turns out the `/api-docs` tree itself is fully deployed and serves fine, only `IHP-ViewPrelude.html` (the module the dead link pointed at) is genuinely missing now that `hsx` has moved out. The blanket redirect I tried just shadowed working URLs.

So per-link Guide fixes like this one are the right approach. Merging — thanks for the contribution! 🙏

I'll follow up with a sweep for any other stale `/api-docs/IHP-ViewPrelude.html` references in the Guide.

PR #2656, opened 2026-04-25
docs: add optional AutoRoute → routes DSL migration to UPGRADE.md

mpscholten:
## Summary

Follow-on to #2652 (explicit-routes DSL). Adds an **optional** migration section to `UPGRADE.md` walking through how to convert a controller from `instance AutoRoute Foo` to the new `[routes|…|]` block.

`AutoRoute` is fully supported and existing apps need no changes — this section is a recipe for projects that want the migration, not a forced upgrade step.

## What's covered

- **Why migrate** — RESTful URLs, compile-time validation, 10–50× faster dispatch on apps with many controllers
- **The recipe** — replace `instance AutoRoute Foo` with a `[routes|…|]` block whose URL shapes are byte-identical to AutoRoute's defaults (so deep links keep working)
- **`?fieldName` semantics** — required / optional / list inferred from the record field's Haskell type
- **Coverage check** — every record field must be path-bound or query-bound; the splice errors at compile time with the exact missing fields
- **Field renaming** via `{ field = #name }`
- **HTTP methods** — `GET|POST` and `ANY`; how to translate existing `allowedMethodsForAction` overrides
- **RESTful path captures** — `{name}` syntax with a note that changing path shapes breaks deep links
- **Mixed mode** — one controller on `AutoRoute`, another on the DSL, in the same `FrontController`
- **Whole-app blocks** — the lowercase-header `webRoutes` binding form
- **When *not* to migrate** — custom `autoRoute` for non-UUID id types, exotic `allowedMethodsForAction` logic, programmatic route generation
- **Codegen** — `ihp-new` and `new-controller` default to the DSL; existing projects keep `AutoRoute` until manually converted

Cross-references [Guide/routing.markdown#explicit-routes-dsl](Guide/routing.markdown#explicit-routes-dsl) for the full syntax reference.

## Test plan

- [ ] Render the section locally — table renders, code blocks fenced correctly, anchor link to Guide works
- [ ] Verify the recipe by following it on a test controller in a sandbox app

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -176,6 +176,211 @@ forEach rows \row -> putStrLn (show row.id <> ": " <> row.title)
 
 The following types have also been removed: `HasQueryBuilder`, `JoinQueryBuilderWrapper`, `NoJoinQueryBuilderWrapper`, `LabeledQueryBuilderWrapper`, `LabeledData`, `NoJoins`. If your code references these types, replace with `QueryBuilder table` directly. For `LabeledData`, use a tuple with `typedSql` instead (see `labelResults` example above).
 
+## Optional: Migrate from `AutoRoute` to the explicit-routes DSL
+
+1.6.0 ships a new `[routes|…|]` quasi-quoter for declaring routes explicitly. `instance AutoRoute` is **fully supported** and existing apps need no changes — this is a recommendation, not a breaking change. New projects scaffolded with `ihp-new` and new controllers scaffolded with `new-controller` use the DSL by default; you can convert existing controllers at your own pace, one at a time, or leave them on `AutoRoute` forever.
+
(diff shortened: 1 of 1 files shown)


amitaibu:
@mpscholten I like `instance AutoRoute PostsController` for how concise it is, but in the AI world, it can write the `[routes||]` for me.

Maybe we should keep only the new DSL, for the sake of a smaller code base. I think v16 will have other breaking changes anyway?

mpscholten:
agree, AutoRpute was an api for humans 😄 doesn't make sense anymore when agents do all the work.

agree on removal. not decided on whether we should already remove it in v1.6 or rather keep it around until v1.7. Alternatively i was also thinking about possible starting a v2 branch soon and then remove it only in v2

PR #2665, opened 2026-04-26
Add GADT typed actions with OpenAPI routes

vcombey:
## Summary
- add GADT typed action request/response specs with JSON/form body handling
- derive OpenAPI route docs from the route DSL and typed action metadata
- add routable Swagger UI/OpenAPI JSON controller actions configurable through [routes|...|]


Guide/form.markdown
```diff
@@ -79,6 +79,139 @@ Rendering [`{textField #title}`](https://ihp.digitallyinduced.com/api-docs/IHP-V
...
+
+First define an input record that is also used by the action body:
+
+```haskell
+data PostInput = PostInput
+    { title :: Text
+    , body :: Text
+    , published :: Bool
+    }
+    deriving (Generic)
```
mpscholten, line 119: Can we avoid Generic?

Generics are very slow on large records and break IHPs live reloading (because eventually it needs 10s for a reload)

With coding agents at least the boilerplate is quickly done :D

Guide/form.markdown
```diff
@@ -79,6 +79,139 @@ Rendering [`{textField #title}`](https://ihp.digitallyinduced.com/api-docs/IHP-V
 <input ... value="Hello World" />
 ```
 
+## Forms for Typed Actions
```
mpscholten, line 82: Can you add a motivation why as a user from `formFor` I would want to use `formForAction`?

Guide/form.markdown
vcombey: hum, didn't know about that.
it breaks live reloading when editing the record you mean ? 
Because I have 100 generic deriving types in my Application/Types and never saw a latency.
handwriten boilerplate can diverge though, could it be a macro that generates only the instance we need without the rest of generic ? 
But for a json api it will also need FromJSON instance and ToSchema for the openapi..

Guide/form.markdown
vcombey: I added more details, it is not entirerly clear for me what drift it prevent in details compared to formfor on a model.
But it adds the possibility to do typed form without a model.

Guide/form.markdown
mpscholten: It only happens when a data structure has many fields or many constructors. See https://gitlab.haskell.org/ghc/ghc/-/issues/5642


I guess your 100 generic types are mostly small data structures (few fields, single constructor). For these cases it works. But in large IHP apps sometimes data structures have 30 fields or more (e.g. a users table might have 30 columns)

For the aeson JSON instances, can't we use template haskell?

vcombey: should work for aeson yes,

What about ToSchema and form url encoding parsing ?
should we add some deriving TH, like https://github.com/ilyakooo0/deriving-openapi3/tree/master ?

vcombey: image
I benchmarked it with a record with 100 field.

PR #2671, opened 2026-04-27
Make formFor work with the lowercase routes DSL

mpscholten:
## Summary

When an app's routes are migrated from AutoRoute to the lowercase routes DSL (e.g. `/users/{userId}/edit`), every `formFor record` call has to be rewritten as `formFor' record (pathTo …)` to get a correct form `action` URL — the default `ModelFormAction` instance synthesizes AutoRoute-style URLs (`/CreateUser`, `/UpdateUser?userId=…`) by string-mangling the current request path, which produces nonsense like `/users/123/UpdateUser?userId=123` on lowercase DSL routes.

This PR makes `formFor` "just work" after a routes-DSL migration:

- **Default instance becomes overridable.** [ihp/IHP/View/Form/FormFor.hs](https://github.com/digitallyinduced/ihp/blob/claude/naughty-sinoussi-f06561/ihp/IHP/View/Form/FormFor.hs) — adds `{-# OVERLAPPABLE #-}` to the existing `ModelFormAction` instance and updates the doc comment. Body is unchanged; AutoRoute apps see no behavioural difference.
- **`ihpEmit` emits per-model overrides.** [ihp/IHP/Router/IHP.hs](https://github.com/digitallyinduced/ihp/blob/claude/naughty-sinoussi-f06561/ihp/IHP/Router/IHP.hs) — for each controller whose actions match the standard scaffold shape (nullary `Create<X>Action`, `Update<X>Action { <x>Id :: Id <X> }`, and `<X>` resolves via `lookupTypeName`), the splice now emits:
  ```haskell
  instance {-# OVERLAPPING #-} ModelFormAction X where
      modelFormAction record
        | isNew record = pathTo Create<X>Action
        | otherwise    = pathTo (Update<X>Action (record.id))
  ```
  Emission is opportunistic — irregular shapes (extra params on the action constructors, non-conventional names) silently fall through to the OVERLAPPABLE default or to a user-defined override.
- **Documentation.** [Guide/form.markdown](https://github.com/digitallyinduced/ihp/blob/claude/naughty-sinoussi-f06561/Guide/form.markdown) — explains the auto-emission convention and shows the manual `{-# OVERLAPPING #-}` override for irregular cases.

## Test plan

- [x] New spec `Test.View.FormForDSLSpec` — defines a `DslProject` model + a `[routes|…|]`-routed controller and asserts forms render `action="/dsl-projects"` for new records and `action="/dsl-projects/<uuid>"` for existing records.
- [x] Existing `Test.View.FormSpec` AutoRoute test (`/CreateProject`) still passes — confirms backward compatibility.
- [x] Full `ihp` test suite: 572 examples, 0 failures.
- [ ] Watch CI.

### DIFF
--- a/Guide/form.markdown
+++ b/Guide/form.markdown
@@ -756,7 +756,21 @@ renderForm post = formFor post [hsx|
 |]
 ```
 
-To override the auto-generated `action` attribute use the [`formFor'`](https://ihp.digitallyinduced.com/api-docs/IHP-View-Form.html#v:formFor-39-) function:
+`formFor` works out of the box for both routing styles:
+
+- **AutoRoute**: the form action defaults to `/CreatePost` or `/UpdatePost?postId=…`.
+- **Routes DSL** (`[routes|…|]`): when both submit actions are declared in the routes block and follow the standard scaffold shape — a nullary `Create<X>Action` and an `Update<X>Action { <x>Id :: Id <X> }` — the splice automatically emits a per-model `ModelFormAction` instance that resolves the form action via `pathTo`. New-record forms submit to the `Create<X>Action` route, and existing-record forms submit to the `Update<X>Action` route.
+
+If your action constructors don't fit that shape (e.g. `CreatePostAction { teamId :: Id Team }` taking extra parameters), provide a manual `ModelFormAction` instance once per model. Put the manual instance before the `[routes|…|]` splice that declares the controller, so the splice can detect it and avoid generating another instance:
+
+```haskell
+instance {-# OVERLAPPING #-} ModelFormAction Post where
+    modelFormAction post
+      | isNew post = pathTo (CreatePostAction { teamId = currentTeamId })
+      | otherwise  = pathTo (UpdatePostAction { postId = post.id })
+```
+
+To override the auto-generated `action` attribute at a single call site, use the [`formFor'`](https://ihp.digitallyinduced.com/api-docs/IHP-View-Form.html#v:formFor-39-) function:
 
 ```haskell
 renderForm :: Post -> Html
--- a/ihp/IHP/Router/IHP.hs
+++ b/ihp/IHP/Router/IHP.hs
@@ -53,6 +53,9 @@ import qualified Language.Haskell.TH.Quote as TH
 
 import IHP.Router.Capture (UrlCapture (..))
 import qualified IHP.ModelSupport as ModelSupport
+import IHP.ModelSupport (isNew)
+import IHP.Router.UrlGenerator (pathTo)
+import IHP.View.Form.FormFor (ModelFormAction (..))
 import IHP.Router.DSL.TH
     ( ParsedBlock (..)
     , HeaderForm (..)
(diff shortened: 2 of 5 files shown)


Guide/form.markdown
```diff
@@ -756,7 +756,21 @@ renderForm post = formFor post [hsx|
 |]
 ```
 
-To override the auto-generated `action` attribute use the [`formFor'`](https://ihp.digitallyinduced.com/api-docs/IHP-View-Form.html#v:formFor-39-) function:
+`formFor` works out of the box for both routing styles:
+
+- **AutoRoute**: the form action defaults to `/CreatePost` or `/UpdatePost?postId=…`.
+- **Routes DSL** (`[routes|…|]`): when a controller's actions follow the standard scaffold shape — a nullary `Create<X>Action` and an `Update<X>Action { <x>Id :: Id <X> }` — the splice automatically emits a per-model `ModelFormAction` instance that resolves the form action via `pathTo`. Forms point at the lowercase DSL paths (`/posts`, `/posts/{postId}`) without any code change in the view.
```
amitaibu: > Forms point at the lowercase DSL paths (`/posts`, `/posts/{postId}`) without any code change in the view.

That line isn't clear, as until now we're talking about Create and Update. Does it refer only to listings?

amitaibu: Sorry, I read it wrong, one talks about the Action and the other about the path.
Still, I think the doc is a bit confusing 😄

ihp/IHP/Router/IHP.hs
```diff
@@ -261,3 +269,88 @@ emitNamedBinding bindingTxt ctrls wsBindings = do
...
+                (TH.AppTypeE (TH.VarE 'GHC.Records.getField) (TH.LitT (TH.StrTyLit "id")))
+                (TH.VarE recordVar)
+        body = TH.CondE
+            (TH.AppE (TH.VarE isNewFn) (TH.VarE recordVar))
+            (TH.AppE (TH.VarE pathToFn) (TH.ConE createCon))
+            (TH.AppE (TH.VarE pathToFn)
+                (TH.AppE (TH.ConE updateCon) getIdExpr))
+        method = TH.FunD modelFormActionFn
+            [TH.Clause [TH.VarP recordVar] (TH.NormalB body) []]
+    in TH.InstanceD (Just TH.Overlapping) []
```
mpscholten, line 407: **[P2] Avoid duplicate instances for existing manual overrides**

This emits a concrete `ModelFormAction X` instance for every standard-looking controller. Apps that already worked around DSL form actions by adding the documented manual `instance ModelFormAction Post` will now have two identical instance heads after upgrading. `OVERLAPPING` only helps choose between overlapping instance heads; it does not allow two `ModelFormAction Post` instances to coexist, so those apps will fail to compile. This needs an upgrade path, an opt-out, or some other way to avoid generating the instance when users already provide one.

mpscholten (commented):
Adding one more compatibility issue; the other two findings from my local pass are already present as existing inline Codex comments on this PR.

PR #2680, opened 2026-05-07
Fix initAuthentication to read raw UUID ASCII bytes from session

Montmorency:
Auth Login Broke silently on a legacy app I had which used initAuthentication. Claude diagnosed this:

The login function was changed to write raw UUID ASCII bytes via sessionInsert, but initAuthentication still used getSession which goes through Serialize.decode expecting a length-prefixed wire format. This caused a silent format mismatch where sessions written by login could never be read back by initAuthentication.

Use the same lookupSessionVault + parseSessionUUID pattern used by userIdMiddlewareFor to read raw ASCII bytes directly.

### DIFF
--- a/ihp/IHP/LoginSupport/Middleware.hs
+++ b/ihp/IHP/LoginSupport/Middleware.hs
@@ -221,6 +221,13 @@ initAuthentication :: forall user normalizedModel.
         , KnownSymbol (GetModelName user)
     ) => IO ()
 initAuthentication = do
-    user <- getSession @(Id user) (sessionKey @user)
-            >>= fetchOneOrNothing
+    userId <- case lookupSessionVault ?request of
+        Just (lookupFn, _) -> do
+            rawValue <- lookupFn (sessionKey @user)
+            pure $ case rawValue of
+                Nothing -> Nothing
+                Just "" -> Nothing
+                Just bs -> Id <$> parseSessionUUID bs
+        Nothing -> pure Nothing
+    user <- fetchOneOrNothing userId
     putContext user
--- a/ihp/IHP/LoginSupport/Middleware.hs
+++ b/ihp/IHP/LoginSupport/Middleware.hs
@@ -221,13 +221,13 @@ initAuthentication :: forall user normalizedModel.
         , KnownSymbol (GetModelName user)
     ) => IO ()
 initAuthentication = do
-    userId <- case lookupSessionVault ?request of
+    maybeUuid <- case lookupSessionVault ?request of
         Just (lookupFn, _) -> do
             rawValue <- lookupFn (sessionKey @user)
             pure $ case rawValue of
                 Nothing -> Nothing
                 Just "" -> Nothing
-                Just bs -> Id <$> parseSessionUUID bs
+                Just bs -> parseSessionUUID bs
         Nothing -> pure Nothing
-    user <- fetchOneOrNothing userId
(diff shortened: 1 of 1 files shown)


mpscholten (changes requested):
Thanks for catching this — the diagnosis in the description is exactly right. But the new body doesn't type‑check. I reproduced it against the real IHP module set (GHC 9.10.3) and got two ambiguous‑type errors:

```
error: [GHC-05617]
    • Could not deduce ‘PrimaryKey table0 ~ UUID’
      The type variable ‘table0’ is ambiguous
    • In the first argument of ‘(<$>)’, namely ‘Id’
      In the expression: Id <$> parseSessionUUID bs

error: [GHC-27958]
    • Could not deduce ‘GetTableName (GetModelByTableName table0) ~ table0’
        arising from a use of ‘fetchOneOrNothing’
      The type variable ‘table0’ is ambiguous
    • Relevant bindings include
        userId :: Maybe (Id' table0)
```

### Why

`Id` is the constructor of `newtype Id' table = Id (PrimaryKey table)` and `PrimaryKey` is **not declared injective** (see `ihp/IHP/ModelSupport/Types.hs`). So `Id <$> parseSessionUUID bs :: Maybe (Id' t)` for some unknown `t`, and the only constraint mentioning `UUID` is `PrimaryKey (GetTableName normalizedModel) ~ UUID` — non‑injectivity prevents GHC from running that backwards to conclude `t ~ GetTableName normalizedModel`.

The original `getSession @(Id user)` dodged this by pinning the type via `TypeApplications`. The new code has no such anchor — `userIdMiddlewareFor` doesn't hit this bug because it returns `Maybe UUID` (no `Id` wrap), and `fetchUserMiddlewareFor` doesn't hit it because `Vault.insert userKey user` pins `user`'s type, which back‑propagates. `putContext :: forall v. Typeable v => v -> IO ()` is too permissive to constrain anything here. (`AllowAmbiguousTypes` only relaxes signature‑level checks, not use‑site inference.)

### Suggested fix

Mirror `userIdMiddlewareFor`'s shape — keep the helper returning `Maybe UUID`, wrap with `Id` at the fetch site:

```haskell
initAuthentication = do
    maybeUuid <- case lookupSessionVault ?request of
        Just (lookupFn, _) -> do
            rawValue <- lookupFn (sessionKey @user)
            pure $ case rawValue of
                Nothing  -> Nothing
                Just ""  -> Nothing
                Just bs  -> parseSessionUUID bs
        Nothing -> pure Nothing
    user <- fetchOneOrNothing (fmap Id maybeUuid :: Maybe (Id user))
    putContext user
```

A smaller diff that also compiles is to annotate the wrapped value in place: `Just bs -> (Id <$> parseSessionUUID bs) :: Maybe (Id user)`. Either works; I'd lean toward the version above because the two readers (`userIdMiddlewareFor` and `initAuthentication`) end up visibly the same shape.

Both options were verified to compile cleanly (`echo ':l ihp/IHP/LoginSupport/Middleware.hs' | direnv exec . ghci`, 69 modules loaded OK).

### Regression test

`ihp/Test/Test/LoginSupport/AuthVaultSpec.hs` already has `mockSessionMiddleware` and tests for `userIdMiddleware` + `parseSessionUUID`. Adding a spec that writes raw UUID ASCII via the session insert path and asserts `initAuthentication` populates `currentUserOrNothing` would catch both the original bug *and* the compile failure here, and pin the contract for the legacy path going forward. Worth including in this PR.

Montmorency:
Added type annotation and ran ghci compile test via claude.

PR #2681, opened 2026-05-07
Parse VARIADIC function arguments in pg_dump indexes

vcombey:
## Summary
- add a `VariadicExpression` node to the postgres parser AST
- parse and compile `VARIADIC <expression>` function arguments
- teach schema diff/reference walkers to recurse through variadic arguments
- add parser/compiler/migration regressions for pg_dump expression indexes using `jsonb_extract_path_text(..., VARIADIC ARRAY[...])`

## Root Cause
pg_dump emits JSONB path calls as `jsonb_extract_path_text(input, VARIADIC ARRAY[...])`. The parser did not understand `VARIADIC` inside function arguments, so `callExpr` backtracked and the following `(` was reported as unexpected while parsing CREATE INDEX expressions.

## Validation
- `ghci ihp-postgres-parser/Test/Postgres/ParserSpec.hs`: 34 examples, 0 failures
- `ghci ihp-postgres-parser/Test/Postgres/CompilerSpec.hs`: 24 examples, 0 failures
- `ghci ihp-ide/Test/Main.hs`: 403 examples, 0 failures

### DIFF
--- a/ihp-ide/IHP/IDE/CodeGen/MigrationGenerator.hs
+++ b/ihp-ide/IHP/IDE/CodeGen/MigrationGenerator.hs
@@ -529,6 +529,7 @@ normalizeExpression (DotExpression a b) = DotExpression (normalizeExpression a) 
 normalizeExpression (ExistsExpression a) = ExistsExpression (normalizeExpression a)
 normalizeExpression (InArrayExpression exprs) = InArrayExpression (map normalizeExpression exprs)
 normalizeExpression (ArrayLiteralExpression exprs) = ArrayLiteralExpression (map normalizeExpression exprs)
+normalizeExpression (VariadicExpression expr) = VariadicExpression (normalizeExpression expr)
 
 -- | Replaces @table.field@ with just @field@
 --
@@ -566,6 +567,7 @@ unqualifyExpression scope expression = doUnqualify expression
         doUnqualify (ExistsExpression a) = ExistsExpression (doUnqualify a)
         doUnqualify (InArrayExpression exprs) = InArrayExpression (map doUnqualify exprs)
         doUnqualify (ArrayLiteralExpression exprs) = ArrayLiteralExpression (map doUnqualify exprs)
+        doUnqualify (VariadicExpression expr) = VariadicExpression (doUnqualify expr)
         doUnqualify (DotExpression (VarExpression scope') b) | scope == scope' = VarExpression b
         doUnqualify (DotExpression a b) = DotExpression (doUnqualify a) b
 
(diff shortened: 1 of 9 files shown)


mpscholten:
## Review

Nice, surgical change — adds `VariadicExpression` to the AST so pg_dump's `jsonb_extract_path_text(input, VARIADIC ARRAY[...])` round-trips. I checked every exhaustive `Expression` walker (`compileExpression`, `compileExpressionWithOptionalParenthese`, `normalizeExpression`, `doUnqualify`, `resolveAlias`, `isRef`, `expressionReferencesColumn`) and they're all updated.

### Strengths

- Keyword boundary handled correctly: `notFollowedBy (satisfy \c -> isAlphaNum c || c == '_')` prevents `VARIADIC` from matching as a prefix of a longer identifier.
- The explicit `compileExpressionWithOptionalParenthese` clause is necessary — without it the default fallback would emit `f(input, (VARIADIC ARRAY[...]))`, which is invalid PostgreSQL.
- Tests cover the worst-case shape (variadic call nested inside `=` inside `AND` inside a WHERE).
- Drive-by trailing-newline fix on `SchemaOperations.hs`.

### Suggestions

1. **Add a comment about `expression` vs `term` in `variadicExpr`.** The body uses `VariadicExpression <$> expression`, which is over-permissive in theory — `VARIADIC arr = b` would parse as `VariadicExpression (EqExpression arr b)` if it appeared at the top level. It's safe in practice because PG's grammar restricts `VARIADIC` to function-argument position (bounded by `,` or `)`), and `expression` is actually the *right* choice because it correctly absorbs postfix typecasts like `VARIADIC ARRAY['a']::text[]` into the variadic operand. A one-line comment locking in that reasoning would help future maintainers.

2. **Add a regression test for one of the migration walkers.** The PR teaches `normalizeExpression`, `doUnqualify`, and `resolveAlias` to recurse through `VariadicExpression`, but no test exercises those paths directly. If a future refactor drops one of those clauses, only the index-diff path would catch it. A targeted `normalizeExpression` round-trip on a `CREATE INDEX … VARIADIC ARRAY[…]` would be enough.

3. *(Nice-to-have)* Mirror the parser test in `ihp-ide/Test/IDE/SchemaDesigner/CompilerSpec.hs` for symmetry with how `ArrayLiteralExpression` is tested at both layers. Lower priority since the postgres-parser compiler test already round-trips it.

### Risks

Adding a new `Expression` constructor is a breaking change for any downstream tooling that pattern-matches exhaustively on it. Worth calling out in release notes.

### Nit

The Validation section in the PR body has literal `\n` instead of line breaks — looks like double-escaping. Worth re-pasting so reviewers can copy-run the commands.

LGTM with the two small asks above.

PR #2682, opened 2026-05-07
Parse pgvector column and index syntax

vcombey:
## Summary
- parse custom postgres type modifiers such as `VECTOR(1536)` and compile them back unchanged
- add parser/compiler support for additional index methods, including `HNSW` and `IVFFLAT`
- preserve index column operator classes such as `vector_cosine_ops`
- keep schema diff/index normalization preserving operator classes

## Root Cause
pg_dump and pgvector schemas can emit SQL like `ALTER TABLE ... ADD COLUMN embedding VECTOR(1536)` and `CREATE INDEX ... USING hnsw (embedding vector_cosine_ops)`. The parser previously stopped at the custom type modifier `(1536)`, only accepted btree/gin/gist index methods, and had no AST slot for index operator classes.

## Validation
- `ghci ihp-postgres-parser/Test/Postgres/ParserSpec.hs`: 38 examples, 0 failures
- `ghci ihp-postgres-parser/Test/Postgres/CompilerSpec.hs`: 28 examples, 0 failures
- `ghci ihp-ide/Test/Main.hs`: 408 examples, 0 failures

### DIFF
--- a/ihp-ide/IHP/IDE/CodeGen/MigrationGenerator.hs
+++ b/ihp-ide/IHP/IDE/CodeGen/MigrationGenerator.hs
@@ -738,9 +738,10 @@ normalizeIndexType (Just Btree) = Nothing
 normalizeIndexType indexType = indexType
 
 normalizeIndexColumn :: IndexColumn -> IndexColumn
-normalizeIndexColumn IndexColumn { column, columnOrder } =
+normalizeIndexColumn IndexColumn { column, columnOperatorClass, columnOrder } =
     IndexColumn
         { column = normalizeExpression column
+        , columnOperatorClass
         , columnOrder = normalizeIndexColumnOrder columnOrder
         }
 
--- a/ihp-ide/IHP/IDE/SchemaDesigner/SchemaOperations.hs
+++ b/ihp-ide/IHP/IDE/SchemaDesigner/SchemaOperations.hs
@@ -230,7 +230,7 @@ newColumnIndex tableName columnName =
     { indexName = tableName <> "_" <> columnName <> "_index"
     , unique = False
     , tableName
-    , columns = [IndexColumn { column = VarExpression columnName, columnOrder = [] }]
+    , columns = [indexCol (VarExpression columnName)]
     , whereClause = Nothing
     , indexType = Nothing
     , nullsDistinct = True
@@ -257,7 +257,7 @@ addForeignKeyConstraint :: Text -> Text -> Text -> Text -> OnDelete -> [Statemen
 addForeignKeyConstraint tableName columnName constraintName referenceTable onDelete list = list <> [AddConstraint { tableName = tableName, constraint = ForeignKeyConstraint { name = Just constraintName, columnName = columnName, referenceTable = referenceTable, referenceColumn = "id", onDelete = (Just onDelete) }, deferrable = Nothing, deferrableType = Nothing }]
 
 addTableIndex :: Text -> Bool -> Text -> [Text] -> [Statement] -> [Statement]
-addTableIndex indexName unique tableName columnNames list = list <> [CreateIndex { indexName, unique, tableName, columns = columnNames |> map (\columnName -> IndexColumn { column = VarExpression columnName, columnOrder = [] }), whereClause = Nothing, indexType = Nothing, nullsDistinct = True }]
+addTableIndex indexName unique tableName columnNames list = list <> [CreateIndex { indexName, unique, tableName, columns = columnNames |> map (indexCol . VarExpression), whereClause = Nothing, indexType = Nothing, nullsDistinct = True }]
 
 -- | An enum is added after all existing enum statements, but right before @CREATE TABLE@ statements
 addEnum :: Text -> Schema -> Schema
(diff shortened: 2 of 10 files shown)


mpscholten:
## Code Review

### Overview

This PR teaches IHP's PostgreSQL parser/compiler to round-trip pgvector schemas:

1. **Type modifiers on custom types** — `VECTOR(1536)` is parsed into `PCustomType "VECTOR(1536)"` (modifier embedded in the name) and re-emitted unchanged.
2. **Six new index methods** — `hash`, `spgist`, `brin`, `hnsw`, `ivfflat` join `btree`/`gin`/`gist`.
3. **Per-column operator classes** — `IndexColumn` gains a `columnOperatorClass :: Maybe Text` field so `(embedding vector_cosine_ops)` survives parse → diff → compile.
4. Migration normalization (`normalizeIndexColumn`) and IDE constructors (`newColumnIndex`, `addTableIndex`) are updated for the new field.

### Strengths

- **Right design choice for scope** — embedding the modifier inside `PCustomType` (instead of introducing `PVector Int`) keeps the change small and unblocks pgvector users without committing IHP to a typed-vector model.
- **Operator classes propagate through `normalizeIndexColumn`**. Without that update, the migration diff would lose the operator class on every reload. Easy to forget — good catch.
- **Refactor to `indexCol`** in `SchemaOperations.hs` replaces two open-coded `IndexColumn { … columnOrder = [] }` literals with the existing helper, so the new field doesn't have to be repeated at each call site. Nice cleanup.
- **`compileIndexColumn` simplification** — collapsing the two equations into one `unwords` makes the new third token (operator class) compose cleanly.
- **Tests** cover both directions (parse + compile) for `VECTOR(1536)` and HNSW with `vector_cosine_ops`, plus an IVFFLAT-only parser test.

### Issues / Suggestions

#### 1. Custom-type modifier swallows `)` greedily — risk for multi-arg modifiers

In `customType`:

```haskell
typeModifier <- optional do
    char '('
    space
    value <- Text.strip <$> takeWhile1P (Just "Custom type modifier") (/= ')')
    char ')'
```

- `optional do …` is **not wrapped in `try`**, so once `char '('` succeeds, any later failure inside the block is fatal — it won't backtrack. For pgvector that's fine (modifiers are simple ints), but safer would be: `optional . try $ do …`.
- Nested parens inside the modifier (e.g. hypothetical `MYTYPE(numeric(10,2))`) are not supported because of `(/= ')')`. Worth a comment noting the limitation.
- Minor: the trailing `space` inside the `optional do` block is redundant — the next combinator already consumes leading whitespace via `lexeme`.

#### 2. New index types `Hash`, `Spgist`, `Brin` have no test

`ParserSpec.hs` and `CompilerSpec.hs` only add tests for HNSW / IVFFLAT. The other three constructors are added in the data type and parsed/compiled, but never exercised. A handful of parameterized tests covering each method (especially `spgist` ↔ `gist` — see next point) would be cheap insurance.

#### 3. `gist` / `spgist` ordering in `parseIndexType`

```haskell
[ ("btree", Btree)
, ("hash", Hash)
, ("gist", Gist)
, ("spgist", Spgist)
…
```

`symbol'` is a plain case-insensitive match without a word boundary. Input `spgist` starts with `s`, so `gist` won't accidentally match it — but the ordering still feels fragile. Consider adding a `notFollowedBy (alphaNumChar <|> char '_')` after the keyword, or at minimum reorder so the longest prefix (`spgist` before `gist`) comes first as a defensive habit. A test that parses `USING spgist` would make this an enforced invariant.

#### 4. `parseIndexColumnOperatorClass` keyword blacklist is incomplete

```haskell
parseIndexColumnOperatorClass = try do
    operatorClass <- qualifiedIdentifier
    when (Text.toUpper operatorClass `elem` ["ASC", "DESC", "NULLS"]) do
        fail "Expected index operator class"
    pure operatorClass
```

The blacklist guards against `ASC`/`DESC`/`NULLS`. Postgres also allows `<expr> COLLATE "C" <opclass>` in an index column. The parser doesn't support `COLLATE` today, but if it ever does, `COLLATE` would need to join the list. A short comment noting this would help future maintainers.

#### 5. `normalizeIndexType` only collapses `Btree`

```haskell
normalizeIndexType (Just Btree) = Nothing
normalizeIndexType indexType = indexType
```

This is correct for the new constructors (they're all distinct from "no index type specified"), but a round-trip test in `MigrationGeneratorSpec` (parse → normalize → diff against itself → expect empty diff) for an HNSW index with operator class would lock the no-spurious-migration behavior in.

#### 6. Compile/parse case asymmetry

The compiler emits `VECTOR(1536)` and `USING HNSW` (uppercase). The parser is case-insensitive. So `compile(parse("vector(1536)")) == "vector(1536)"` whereas `compile(parse("VECTOR(1536)")) == "VECTOR(1536)"` — case is preserved verbatim. Fine for pg_dump output (which is uppercase), but worth being aware of if mixed-case sources show up.

### Performance & Security

- **Performance**: parser additions are O(1) per index/column; no concerns.
- **Security**: no new external input handling — strictly schema-text round-tripping.

### Verdict

The change is well-scoped, follows existing parser/compiler patterns, and correctly threads the new field through normalization. Before merging I'd suggest:

1. Add tests for `hash`, `spgist`, `brin` (plus an `spgist` test specifically, which guards the keyword-prefix concern).
2. Add a migration round-trip test asserting an HNSW index with operator class produces no diff against itself.
3. Wrap the type-modifier `optional do` in `try` (or document why it's safe without).
4. Optional: a one-line comment in `parseIndexColumnOperatorClass` listing the keywords excluded so future extensions remember to grow the list.

Once those land it's ready to ship.

PR #2684, opened 2026-05-08
Parse pg_dump CREATE FUNCTION SET TO options

vcombey:
## What changed

- Accept `SET <name> TO <value>` inside `CREATE FUNCTION` options in `ihp-postgres-parser`.
- Preserve schema-qualified function names such as `private.sync_access` when parsing `CREATE FUNCTION`.
- Add a parser spec covering pg_dump-style `CREATE FUNCTION ... SET search_path TO ...` output.

## Why

PostgreSQL pg_dump can emit function options with `TO` instead of `=`, and can qualify functions with a non-public schema. The parser previously failed before IHP could load this schema output.

## Validation

- `direnv exec . cabal test --project-dir=ihp-postgres-parser spec` passed: 72 examples, 0 failures.

### DIFF
--- a/ihp-postgres-parser/IHP/Postgres/Parser.hs
+++ b/ihp-postgres-parser/IHP/Postgres/Parser.hs
@@ -698,7 +698,7 @@ createFunction = do
     lexeme "CREATE"
     orReplace <- isJust <$> optional (lexeme "OR" >> lexeme "REPLACE")
     lexeme "FUNCTION"
-    functionName <- qualifiedIdentifier
+    functionName <- functionIdentifier
     functionArguments <- between (char '(') (char ')') (functionArgument `sepBy` (char ',' >> space))
     space
     lexeme "RETURNS"
@@ -752,7 +752,7 @@ parseFunctionSetting :: Parser FunctionOption
 parseFunctionSetting = do
     symbol' "SET"
     settingName <- qualifiedIdentifier
-    symbol "="
+    symbol "=" <|> symbol' "TO"
     settingValue <- Text.strip . cs <$> someTill anySingle (lookAhead functionOptionBoundary)
     space
     pure (FunctionSettingOption FunctionSetting { settingName, settingValue })
@@ -982,6 +982,10 @@ qualifiedIdentifier = do
         lexeme "public"
         char '.'
     identifier
+
+functionIdentifier = do
+    schemaOrName <- identifier
+    maybe schemaOrName (\name -> schemaOrName <> "." <> name) <$> optional (char '.' >> identifier)
 
 addColumn tableName = do
     lexeme "COLUMN"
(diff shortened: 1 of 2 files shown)


mpscholten:
## Review

The bug is real and the fix shape is right — pg_dump emits both `TO`-style settings and non-`public` schema qualifiers, and the previous parser silently dropped the schema prefix or failed at the `.`. A few things worth addressing before merge:

### High — `dropFunction` asymmetry

`dropFunction` at `ihp-postgres-parser/IHP/Postgres/Parser.hs:1035` still uses `qualifiedIdentifier`, which only strips `public.` and otherwise stops at the dot. After this PR:

- `CREATE FUNCTION private.sync_access(...)` parses → `functionName = "private.sync_access"` ✓
- `DROP FUNCTION private.sync_access;` **fails to parse** — `qualifiedIdentifier` consumes only `private` and then chokes on the `.`.

`MigrationGenerator.hs:178` builds `DropFunction { functionName }` directly from a `CreateFunction`'s name, so the generated SQL is fine for Postgres, but if IHP ever re-parses such a `DROP` from disk (user-edited Schema.sql or a migration file) it breaks. Recommend using `functionIdentifier` in `dropFunction` too.

### Medium — Subtle behavior change for `public.` prefix

`qualifiedIdentifier` normalizes `public.foo` → `foo`. `functionIdentifier` preserves it: `public.foo` → `public.foo`. This makes function names behave differently from every other identifier in the parser (tables, indexes, types, policies, etc.).

- `Compiler.hs:46` emits `functionName` directly (not via `compileIdentifier`), so `public.foo` round-trips OK.
- But `SchemaOperations.hs:518` compares function names via `==`. Schema files that mix qualified and unqualified references to the same function (now possible if the user dumps with pg_dump and edits manually) won't match.

Consider stripping `public.` in `functionIdentifier` (matching `qualifiedIdentifier`) and only preserving non-`public` schemas. That preserves backward compatibility with pre-PR test expectations.

### Low — Missing type signature

`functionIdentifier` at `Parser.hs:986` has no signature. Adjacent `identifier :: Parser Text` does. Adding `functionIdentifier :: Parser Text` is a one-line readability improvement.

### Low — Test coverage gaps

The single new test bundles both fixes together. Consider:

- A test for `SET search_path TO ...` **without** a schema-qualified name, to isolate the `TO` change.
- A test for `CREATE FUNCTION private.x()` **with** `=`-style settings, to isolate the schema-prefix change.
- A round-trip test (parse → compile → parse) covering `private.sync_access`. This would have caught the `dropFunction` asymmetry above.

mpscholten:
Thanks @vcombey! Merged via #2757, which carries your original commit (authorship preserved) plus the review feedback applied on top:

- `functionIdentifier` now normalizes the default `public.` schema away (`public.foo` → `foo`) so function names stay comparable regardless of an explicit prefix, while still preserving non-`public` schemas like `private.sync_access`.
- `DROP FUNCTION` uses the same schema-aware parser, fixing the CREATE/DROP asymmetry.
- Added a type signature + Haddock to `functionIdentifier`.
- Added tests isolating the `SET ... TO ...` and schema-prefix changes plus `parse → compile → parse` round-trips.

Closing in favor of #2757.

PR #2692, opened 2026-05-21
[codex] Enable incomplete pattern errors for apps

mpscholten:
## Summary
- promote incomplete pattern warnings to compile errors in app-facing GHCi defaults
- update the generated Makefile and NixSupport Cabal configuration to use `-Werror=incomplete-patterns`
- document the behavior change and the `-Wwarn=incomplete-patterns` escape hatch for exploratory development

## Why
IHP framework packages already use `-Werror=incomplete-patterns` after #2260, but app-facing defaults still only warned. With agent-assisted coding, missing pattern matches should fail fast instead of being easy-to-miss warnings.

Follow-up to #2047 and #1681.

## Validation
- `git diff --check`
- `printf ':set -Werror=incomplete-patterns\n:set -Wwarn=incomplete-patterns\n:quit\n' | ghci -v0`
- verified a tiny incomplete-pattern module fails with `-Werror=incomplete-patterns`
- verified the same module compiles with a warning when `-Wwarn=incomplete-patterns` is added after the error flag

`nix develop --command ghci -v0` was not usable in this worktree: devenv failed before starting GHC because it could not determine the current directory. `direnv exec .` was also blocked because this worktree has not been allowed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Breaking Changes
 
+- IHP apps now treat incomplete pattern matches as compile errors in the default GHCi, generated Makefile, and generated Cabal configuration. This makes local development and agent-assisted coding fail fast when a pattern match misses a constructor.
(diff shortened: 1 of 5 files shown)


amitaibu:
Thanks!

Issue #1757, opened 2023-07-18
Use Enum types as route params


amitaibu:
We need to use `Text` and decode it in the action. We want to be able to define:

```
data PostController
    = PostController { someEnumType :: !SomeEnumType}
```

mpscholten:
I don't have any ideas how this can be implemented right now, but I assume there's a way to get this working

amitaibu:
Maybe enumParamReader can help us here?

https://github.com/digitallyinduced/ihp/blob/a95e837e0aca4b80579a7dfeff84800ea0dc9123/IHP/Controller/Param.hs#L556-L566

mpscholten:
Documented the current explicit-routes approach in https://github.com/digitallyinduced/ihp/pull/2693.

With the recent routing changes, enum route parameters are handled through `UrlCapture`; the docs now show how generated SQL enums can reuse `textToEnum...` with direct UTF-8 decoding and `inputValue` for rendering.

PR #2693, opened 2026-05-21
[codex] Document enum URL captures

mpscholten:
## Summary

Adds a routing guide note for using generated SQL enum types as explicit route captures and query parameters via `UrlCapture`.

The example reuses the generated `textToEnum...` parser and `inputValue` renderer, and shows direct UTF-8 decoding with `Data.Text.Encoding.decodeUtf8'` instead of a polymorphic string conversion helper.

Refs #1757.

## Validation

- `git diff --check`

### DIFF
--- a/Guide/routing.markdown
+++ b/Guide/routing.markdown
@@ -167,6 +167,20 @@ instance UrlCapture MarketStatus where
 
 The instance can live in `Web/Routes.hs`, in `Web/Types.hs` next to the data declaration, or in any module reachable from the splice site.
 
+For enum types generated from `Application/Schema.sql`, reuse the generated `textToEnum...` parser and `inputValue` renderer. Add `import qualified Data.Text.Encoding` in the module that defines the instance:
+
+```haskell
+instance UrlCapture Color where
+    parseCapture bytes =
+        case Data.Text.Encoding.decodeUtf8' bytes of
+            Right text -> textToEnumColor text
+            Left _ -> Nothing
+
+    renderCapture = inputValue
+```
+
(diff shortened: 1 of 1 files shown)


Guide/routing.markdown
```diff
@@ -167,6 +167,20 @@ instance UrlCapture MarketStatus where
...
+instance UrlCapture Color where
+    parseCapture bytes =
+        case Data.Text.Encoding.decodeUtf8' bytes of
+            Right text -> textToEnumColor text
+            Left _ -> Nothing
+
+    renderCapture = inputValue
+```
+
+Use a direct UTF-8 conversion for `parseCapture`, as shown above. This keeps invalid UTF-8 as a route miss and avoids relying on a polymorphic string conversion helper.
```
amitaibu, line 182: @mpscholten this line is not clear, and a bit complex. Should we improve or just remove it, as it's an edge case?

Guide/routing.markdown
```diff
@@ -167,6 +167,20 @@ instance UrlCapture MarketStatus where
 
 The instance can live in `Web/Routes.hs`, in `Web/Types.hs` next to the data declaration, or in any module reachable from the splice site.
 
+For enum types generated from `Application/Schema.sql`, reuse the generated `textToEnum...` parser and `inputValue` renderer. Add `import qualified Data.Text.Encoding` in the module that defines the instance:
+
+```haskell
+instance UrlCapture Color where
```
amitaibu, line 173: Maybe we can replace the `MarketStatus` with a more generic `Color` ?

PR #2695, opened 2026-05-21
[codex] Clarify enum UrlCapture docs

mpscholten:
## Summary

- make the custom `UrlCapture` example use a generic `Color` enum
- simplify the generated enum example to use `textToEnumColor . cs` with `inputValue`
- remove the extra UTF-8-specific explanation now that the example is simpler

### DIFF
--- a/Guide/routing.markdown
+++ b/Guide/routing.markdown
@@ -139,47 +139,35 @@ The DSL ships with `UrlCapture` instances for the common scalar types: `Text`, `
 For your own types — typically SQL enums — declare a `UrlCapture` instance alongside the type:
 
 ```haskell
-data MarketStatus
-    = MarketStatusDraft
-    | MarketStatusOpen
-    | MarketStatusClosed
-    | MarketStatusResolved
-    | MarketStatusRefunded
+data Color
+    = Red
+    | Green
+    | Blue
     deriving (Eq, Show)
 
-instance UrlCapture MarketStatus where
+instance UrlCapture Color where
     parseCapture = \case
-        "draft"    -> Just MarketStatusDraft
-        "open"     -> Just MarketStatusOpen
-        "closed"   -> Just MarketStatusClosed
-        "resolved" -> Just MarketStatusResolved
-        "refunded" -> Just MarketStatusRefunded
-        _          -> Nothing
+        "red"   -> Just Red
+        "green" -> Just Green
+        "blue"  -> Just Blue
+        _       -> Nothing
     renderCapture = \case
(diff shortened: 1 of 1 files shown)


amitaibu (approved):
Thanks 🙏

Guide/routing.markdown
```diff
@@ -139,48 +139,36 @@ The DSL ships with `UrlCapture` instances for the common scalar types: `Text`, `
...
 
 ```haskell
-data MarketStatus
-    = MarketStatusDraft
-    | MarketStatusOpen
-    | MarketStatusClosed
-    | MarketStatusResolved
-    | MarketStatusRefunded
+data Color
+    = Red
```
amitaibu: Nitpick: Should we prefix `ColorRed` | `ColorGreen` ?

avitkauskas:
Why `parseCapture bytes = textToEnumColor (cs bytes)` is better than `parseCapture = textToEnumColor . cs`?

Also, I noticed that if you provide some wrong param like `?color=yellow` you get a runtime exception "An exception was raised while running the action", unless you use explicit `parseCapture` with `_  -> Nothing`. Is that ok?

mpscholten:
explicit arguments are easier to read for people new to haskell compared to point free style 👍 

Will look into the exception

mpscholten:
https://github.com/digitallyinduced/ihp/pull/2696/changes adds a test for an invalid enum value 👍 I guess the error might be caused by the old routing

avitkauskas:
> explicit arguments are easier to read for people new to haskell compared to point free 

Even I (total Haskell newbie) understand the point free style :) It’s nice to teach newcomers the good Haskell style :)

unhammer:
Point-free is quite contentious. Although a lot of Haskell code is written that way, there are also quite a lot of people who don't like it (thus the alternative moniker "pointless"). Some people like it and that's fine, but a lot of newcomers are *very* put off by it.

Also, although it's more compact, I often find it harder to *modify* point-free code than explicit arguments. A lot of times in the past I have started with some elegant-looking point-free gem, and then noticed I need to change things up a bit, maybe add an argument somewhere, and suddenly we're looking at `&&&` and `***`  and `bimap` and `uncurry` for things that in regular style are completely trivial, so in the end I just go back to the regular style after having wasted much time on pointless brain-teasers. Good code is not only readable at this point in time, but easily changeable.

Also, now there's a [debugger](https://well-typed.github.io/haskell-debugger/), explicit args also become more useful (though I haven't tried the debugger with IHP).

PR #2701, opened 2026-05-24
Guide: document the self-hosted Attic binary cache

mpscholten:
Documents the new self-hosted Attic cache (`https://cache.digitallyinduced.com/public`) in the Binary Cache section of the package-management guide: what it is, that it's public + covers x86_64-linux/aarch64-darwin, and how to opt an existing project in.

New projects get it automatically via digitallyinduced/ihp-boilerplate#65. The cache is populated by CI via digitallyinduced/ihp#2700.


mpscholten:
Closing — the cache is configured automatically via the ihp-boilerplate flake (digitallyinduced/ihp-boilerplate#65), so the user-facing Guide doesn't need to call out the cache implementation.

PR #2720, opened 2026-06-13
Reduce generated code: collapse per-field SetField and FieldBit instances

mpscholten:
## Problem

The SchemaCompiler emits one tiny instance **per field per table** for `SetField`, `UpdateField`, and `FieldBit`. In real apps these three families dominate the per-table `Generated` modules — for our app (~90 tables) that's the bulk of ~24k lines, slow for GHC to typecheck and codegen.

## Change

`SetField` and `FieldBit` are now provided generically instead of per field, at **zero runtime cost** (the code still folds/inlines to exactly what the per-field instances produced):

- **`SetField`** — a single `{-# OVERLAPPABLE #-}` instance in `IHP.Record` delegates `setField = updateField @field`. The compiler now emits only the per-field `UpdateField` instances (the worker, including polymorphic belongs-to variants).
- **`FieldBit`** — changed from a class to a **constraint synonym** `type FieldBit name model = KnownNat (2 ^ FieldIndex name (ModelFields model))` plus a standalone `fieldBit` function. The compiler emits one `type instance ModelFields (Model') = '[...]` per table instead of N per-field `FieldBit` instances.

## Notes on the approach

A global instance on the fundep class `SetField` destabilizes GHC inference for code polymorphic over the model. So:

- Framework helpers that carried abstract `SetField` constraints (`copyRecord`, `fill`, file-upload / validation helpers, …) now use `UpdateField` constraints and call `updateField` directly.
- The bespoke hand-written `SetField` instances for non-model types (`SQLQuery`, `FormContext`, `MetaBag`, `Options`, `DataSyncController`, `ComponentInstance`) are marked `{-# OVERLAPPING #-}` so they keep winning over the generic instance.
- `FieldBit` avoids this entirely by being a constraint synonym rather than a class.

## Testing

- ✅ ihp-ide: **426/426** (incl. `SchemaCompilerSpec` golden output asserting the new `ModelFields` lines)
- ✅ ihp core: **567/567** (exercise `set`/`fill`/`copyRecord`/`didChange`/`fieldBit`)
- ✅ All preludes, `ihp-datasync`, `ihp-ssc`, `ihp-schema-compiler` compile


mpscholten:
Closing this — it breaks real-world apps.

Building belege.ai against this branch fails to compile (`belege-ai-models`/`Generated.Types` is fine, but app code isn't):

```
Application/Helper/Mollie.hs: Could not deduce 'HasField "isValid" MollieConnection' Bool'
  arising from a use of 'set'
  ... conn |> set #isValid True |> updateRecord
```

Root cause: the global `OVERLAPPABLE SetField field model value` instance (constraint `HasField field model value`, class fundep `field model -> value`) breaks type inference of the ubiquitous `record |> set #field value |> updateRecord` pattern on relation-support models that carry type parameters, e.g.:

```haskell
data MollieConnection' userId mollieInvoices invoiceAgentSessions = MollieConnection { …, isValid :: Bool, … }
type MollieConnection = MollieConnection' (Id "users") (QueryBuilder "mollie_invoices") (QueryBuilder "invoice_agent_sessions")
```

GHC can't pin `model` to the fully-applied type and fails to discharge `HasField`. The per-field concrete `SetField` instances (baseline) compile it fine.

CI and the IHP test suite passed because every test model is monomorphic (no type parameters); real app models always have them, which is what triggers the break.

The failure is purely the global `SetField` instance — the `FieldBit` → constraint-synonym reduction is unaffected. If we revisit this, options are: keep only the `FieldBit` reduction, or emit per-field one-line `setField = updateField @field` delegators (no global instance, no inference break, but no instance-count reduction).

PR #2727, opened 2026-06-19
Fix: remove historical nix build options in docs

avitkauskas:
This just removes the historical options in docs
```
nix build .#optimized-docker-image --option sandbox false --extra-experimental-features nix-command --extra-experimental-features flakes
```
and similar commands, as discussed on Slack.


mpscholten (approved):
Thanks 👍

PR #2737, opened 2026-06-23
Add pagination for typedSql queries

amitaibu:
## What

Gives `typedSql` first-class pagination, matching what IHP already offers for the other two query styles. Adds **`paginatedTypedSql`** and **`paginatedTypedSqlWithOptions`** that take a `TypedQuery` and return `([model], Pagination)` — the typedSql analogue of the existing `paginatedSqlQuery` / `paginatedSqlQueryWithOptions`.

## Why

IHP has `paginate`/`paginateWithOptions` (QueryBuilder) and `paginatedSqlQuery`/`…WithOptions` (raw `Text` SQL), but nothing for typedSql. Today, paginating a typedSql query forces apps to re-read `page`/`maxItems`, compute LIMIT/OFFSET, count rows, and build the `Pagination` record by hand — duplicating logic IHP already has internally (`pageSize'`, `offset'`, the 200 cap, `Pagination` construction).

## Where it lives, and why (dependency direction)

`TypedQuery` lives in **`ihp-typed-sql`**; `Pagination`/`Options` live in **`ihp`**. The cabal deps run **`ihp-typed-sql` → `ihp`** (one-way), so the new functions go in a new module **`IHP.TypedSql.Pagination`** in `ihp-typed-sql` — the only package that can see both types without creating a cycle.

## Keeping both paginators in lockstep

The page-math helpers (`pageSize'` with its 200-item cap, `page`, `offset'`) were unexported locals in `IHP.Pagination.ControllerFunctions`. They are extracted **verbatim** into a new shared **`IHP.Pagination.Internal`** (in `ihp`), which both `ControllerFunctions` and the new typedSql paginator import — so the two can't drift. No behavior change to existing pagination functions; this is purely additive.

The new paginator mirrors `paginatedSqlQueryWithOptions` exactly: it wraps `tqSnippet` as `SELECT count(subquery.*) FROM (<query>) as subquery` for the count and `SELECT subquery.* FROM (<query>) as subquery LIMIT ? OFFSET ?` for the page, runs both via `sqlQueryHasql` using `tqResultDecoder` for rows and an int8 decoder for the count.

> [!NOTE]
> Because the query is wrapped in a subquery before `LIMIT`/`OFFSET`, any `ORDER BY` must live **inside** the query you pass in (same as `paginatedSqlQuery`). The Haddock documents this, plus the same AutoRefresh/`trackTableRead` caveat.

## Tests

`Test/Test/TypedSqlSpec.hs` gets a `runtimeTest` that mirrors the raw-SQL `ControllerFunctionsSpec`: seed 100 rows, then assert the page slice and `Pagination` fields (`pageSize`, `totalItems`, `currentPage`, `window`) across default options, a `maxItems` request-param override, page+maxItems together, and custom `Options`.

## Packaging

- `IHP.TypedSql.Pagination` + `wai` added to `ihp-typed-sql.cabal`; `IHP.Pagination.Internal` added to `ihp.cabal`.
- `wai` also added to `ihp-typed-sql/default.nix` — IHP's nix builds use the checked-in cabal2nix-style `default.nix` (not `callCabal2nix`), so cabal deps must be mirrored there.
- `## Unreleased` changelog entry in `ihp-typed-sql/changelog.md`.

## Verification

Authoritative nix build of both changed derivations is green:
- `ihp-1.6.0` compiles under `-Werror=unused-imports` (validates `Internal` + the trimmed `ControllerFunctions`).
- `ihp-typed-sql-1.6.0` library + test-suite compile.

The new runtime test's DB-backed assertions `pendingWith` when Postgres is unavailable (the same gating as every other typedSql runtime test) — run `nix flake check --impure`, or the suite with `PGHOST` set, to exercise them against a database.


amitaibu:
Need to add this to the docs

mpscholten (approved):
Great idea, thanks 👍

PR #2738, opened 2026-06-23
Add typed scalar query helpers to ihp-typed-sql

amitaibu:
## Summary

Adds `sqlQueryTypedScalar` and `sqlQueryTypedScalarOrNothing` to `IHP.TypedSql`, the typed counterparts of the now-deprecated `sqlQueryScalar` / `sqlQueryScalarOrNothing` in `IHP.ModelSupport`.

## Why

`sqlQueryScalar` is deprecated in favor of `[typedSql| ... |]` + `sqlQueryTyped`, but `sqlQueryTyped :: TypedQuery result -> IO [result]` only returns a list. For single-value queries like `SELECT count(*)` there was no typed equivalent of the scalar helper, forcing callers to write `head <$> sqlQueryTyped ...`. This fills that gap so the deprecation story is complete.

## What

- `sqlQueryTypedScalar` — runs a typed query expected to return exactly one row of one value, returning `result` directly (reuses `sqlQueryTyped`'s plumbing, swapping `rowList` for `singleRow`).
- `sqlQueryTypedScalarOrNothing` — same, but returns `Nothing` when there is no row (uses `rowMaybe`).
- Both exported from `IHP.TypedSql`, with Haddock examples.
- Added a `runtimeScalarModule` test exercising both helpers (count on empty + populated table, `OrNothing` with and without a matching row).
- Changelog entry under v1.6.0.

No new types and no other files changed.

## Verification

`nix build` of `ihp-typed-sql` (compiles the library + its test suite) passes.


mpscholten:
One semantic concern: `HasqlDecoders.Row result` is still the right low-level decoder type here, even for a scalar SQL result, because PostgreSQL returns result sets as rows and columns (`SELECT count(*)` is one row with one column).

The issue is that `TypedQuery result` currently does not encode whether the row decoder came from a one-column query or from a multi-column query. Since multi-column typed SQL queries are wrapped as `SqlRow ...`, `sqlQueryTypedScalar [typedSql| SELECT name, views FROM ... LIMIT 1 |]` would return a single `SqlRow` instead of failing like deprecated `sqlQueryScalar` does for non-`Only` results.

So I think this should either:

- enforce a one-column invariant, e.g. add column-count/shape metadata to `TypedQuery` and reject `tqColumnCount /= 1` in `sqlQueryTypedScalar*`, or
- be renamed/documented as a single-row helper rather than a scalar helper.

Given the PR says these are typed counterparts of `sqlQueryScalar` / `sqlQueryScalarOrNothing`, enforcing the one-column scalar shape seems like the better fit.

amitaibu:
Enforced the single-column invariant at compile time (87ca4ec) rather than via a runtime `tqColumnCount` check, to keep it consistent with the rest of typed-sql being compile-time.

`IHP.TypedSql` now has a closed type family:

```haskell
type family ScalarResult result :: Constraint where
    ScalarResult (SqlRow fields) = TypeError (...)   -- multi-column -> rejected
    ScalarResult result          = ()
```

and both `sqlQueryTypedScalar` / `sqlQueryTypedScalarOrNothing` carry a `ScalarResult result` constraint. Since multi-column ad-hoc queries are inferred as `SqlRow`, `sqlQueryTypedScalar [typedSql| SELECT name, views FROM ... |]` is now a compile error pointing at the single-column requirement. Added a compile-fail test (`rejects a multi-column query passed to sqlQueryTypedScalar`) and guide/changelog notes.

One caveat vs. a strict column-count check: this rejects multi-column `SqlRow` results (the reported footgun), but a full-row `typedSqlStar` query — which infers to a model record rather than a `SqlRow` — would still be accepted and return one row. There's no structural type-level signal to reject an arbitrary model record at compile time. If you'd rather enforce "exactly one column" including the star case, I'll add a `tqColumnCount` field to `TypedQuery` and do the runtime check as you originally suggested — let me know which you prefer.

mpscholten:
Great solution with the type family 👍

mpscholten (approved):
Thanks 👍

amitaibu:
> Great solution with the type family 👍

It was my friend Claude :smile:

PR #2740, opened 2026-06-24
Generate DefaultParamEncoder [Maybe enum] for typedSql interpolation

amitaibu:
## What

Adds a `DefaultParamEncoder [Maybe <Enum>]` instance to the schema-compiler output for each generated enum type.

## Why

The schema compiler already emits `DefaultParamEncoder` instances for enums (`<Enum>`), `Maybe <Enum>`, and `[<Enum>]`, so `${enumVal}`, `${Just enumVal}`, and `${[enumVal]}` already bind in `typedSql` queries. The only missing case was a **list of nullable** enum values — `${[Just enumVal]}` had no matching instance.

> Note: a generic `instance DefaultParamEncoder a => DefaultParamEncoder (Maybe a)` in core is **not expressible** with this class — `defaultParam :: NullableOrNot Value a` already bakes in nullability, so the inner `Value a` can't be generically recovered and re-wrapped as `Nullable`. That's why both `hasql-implicits` and IHP's `IHP.Hasql.Encoders` enumerate concrete instances, and why enums get theirs from the generator. This PR follows that same per-enum approach.

## What changed

- **`ihp-schema-compiler`**: emit `DefaultParamEncoder [Maybe <Enum>]` (a nullable-element enum array) per enum, alongside the existing scalar / `Maybe` / `[<Enum>]` instances.
- **`ihp-ide` `SchemaCompilerSpec`**: golden tests updated for the new instance (Mood / Province / PropertyType). ✅ 47 examples, 0 failures locally.
- **`ihp-typed-sql` `TypedSqlSpec`**: new runtime test that round-trips through `${...}` in a `WHERE` clause:
  - `${enumVal}` (scalar)
  - `${Just enumVal}` (Maybe, against a nullable column)
  - `${(Nothing :: Maybe Enum)}` (binds as SQL `NULL`)
  - `${[enumVal]}` and `${[Just enumVal]}` via `= ANY(...)`
- **`CHANGELOG.md`**: noted both changes.

## Goal

`${enumVal}`, `${Just enumVal}`, and `${[Just enumVal]}` all compile and bind — removing the need to split enum-filtered joins into fetch-ids-then-`filterWhereIn` or to cast through text in SQL.

## Verification

- `SchemaCompilerSpec` golden tests pass locally (47/47).
- `SchemaCompiler.hs` and `TypedSqlSpec.hs` typecheck in the dev shell.
- The new runtime test requires Postgres; relying on CI to execute it (couldn't run an isolated local cluster in this environment).


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- The schema compiler now also generates a `DefaultParamEncoder [Maybe <Enum>]` instance for each enum type, alongside the existing `<Enum>`, `Maybe <Enum>`, and `[<Enum>]` instances. This lets `${[Just enumVal]}` (a list of nullable enum values) be interpolated into `typedSql` queries, e.g. `WHERE status = ANY(${[Just Active, Just Pending]})`, removing the need to split enum-filtered joins into fetch-ids-then-`filterWhereIn` or to cast through text in SQL.
+- Added an `ihp-typed-sql` runtime test that round-trips an enum, a `Maybe` enum (including `Nothing` as SQL `NULL`), and a `[Maybe Enum]` list through `${...}` interpolation in a `WHERE` clause.
```
amitaibu:
```suggestion
```

amitaibu:
@mpscholten — design question before this goes further. 🙏

**Goal of this PR:** make enum values usable directly in `typedSql` `${...}` params so enum-filtered joins don't need fetch-ids-then-`filterWhereIn` or text casts — specifically `${enumVal}`, `${Just enumVal}`, and `${[Just enumVal]}`.

**What I found while implementing:**

1. The schema compiler **already** generates `DefaultParamEncoder` instances for `<Enum>`, `Maybe <Enum>`, and `[<Enum>]` (added in c287aa3c). This PR adds the one missing sibling, `[Maybe <Enum>]`, to round out the family (matches how `hasql-implicits` enumerates per-type instances).

2. But the actual blocker for `${Just enumVal}` / `${[Just enumVal]}` is **not** an instance gap — it's that `typedSql` is intentionally designed to take **non-nullable** param inputs:
   - `ParamHints.resolveParamHintTypes` calls `stripMaybeType` on the column type, with the comment *"parameter hints should be non-nullable inputs"*.
   - `Quoter.hs` emits `coerce expr :: <scalarType>` for every param.
   
   So for an enum column the param is coerced to the **scalar** enum type (`TypedSqlTestMood` / `[TypedSqlTestMood]`).

**Verified against a real Postgres:**
- ✅ `${enumVal}` and `${[enumVal]}` compile and bind — and these already work on `master` via the existing `enum` / `[enum]` instances.
- ❌ `${Just enumVal}`, `${(Nothing :: Maybe _)}`, and `${[Just enumVal]}` fail: `Maybe _` / `[Maybe _]` aren't `Coercible` to the scalar type.

Supporting the `Just` / `[Just]` forms isn't a one-instance fix — it conflicts with the single-annotation `coerce … :: T` design (one annotation can't accept both bare `x` and `Just x` for the same column). The PR's underlying goal — filter enum joins inline — is already achievable today with the **bare** forms `${enumVal}` and `${[enumVal]}`.

**Decision needed — which direction do you want?**

- **(A)** Ship the bare forms: keep the new `[Maybe <Enum>]` instance (completeness / direct-hasql use), make the runtime test assert `${enumVal}` and `${[enumVal]}`, and document that `typedSql` params are non-nullable (pass bare enum values). Lowest risk; meets the underlying goal.
- **(B)** Extend the macro so nullable columns infer `Maybe T` and params aren't force-coerced to scalar — enables `${Just enumVal}` / `${[Just enumVal]}`, but it's a broad change affecting **all** param types and risks breaking existing bare-value usage.
- **(C)** Drop it — the bare forms already work on `master` and the extra instance is marginal.

I'd lean **(B)**. Happy to implement whichever you prefer.

mpscholten (approved):
Looks good to me 👍

amitaibu:
I'll ask Claude to look into (B) - Extend the macro so nullable columns infer Maybe T and params aren't force-coerced to scalar — enables ${Just enumVal} / ${[Just enumVal]}, but it's a broad change affecting all param types and risks breaking existing bare-value usage.

amitaibu:
Implemented **(B)** — `typedSql` now accepts `Maybe` / `[Maybe]` params, so `${Just enumVal}` and `${[Just enumVal]}` work, without breaking bare-value usage.

**How:** new `IHP.TypedSql.ParamEncoder.TypedSqlParam` class, keyed on the column's *scalar* Haskell type. The quasiquoter annotates each `${...}` with that scalar type (via a `Proxy`, so no `TypeApplications` needed at the splice site) and emits `typedSqlParam (Proxy :: Proxy col) expr` (a hasql `Snippet`) instead of the old `coerce expr :: T`. Four instances dispatch on the value's shape:

| form | accepts |
|---|---|
| `${x}` | `T` (or a newtype coercible to it, e.g. `Id'`) |
| `${Just x}` / `${Nothing}` | `Maybe T` (`Nothing` → SQL `NULL`) |
| `${[x]}` | `[T]` (`IN` / `= ANY`) |
| `${[Just x]}` | `[Maybe T]` |

Because the **value's shape** selects the instance:
- Bare values keep working for every column, nullable or not — no existing test changed.
- Wrong-typed params are still rejected at compile time (no `Coercible`/instance).
- Array-ness comes from the value, so `resolveParamHintTypes` no longer wraps array hints in `[T]`.

**Verified locally against Postgres:** full `ihp-typed-sql` `TypedSqlSpec` → **103 examples, 0 failures** (1 pending = the `initdb`-on-PATH auto-db test), including the enum round-trip test now exercising all four shapes + `Nothing`→NULL. Branch rebased on your changelog suggestion and current with `master`. Letting CI confirm across GHC 9.10/9.12/9.14.

amitaibu:
@mpscholten I have to admit that some of Claude's changes aren't clear to me, so please give them your attention :smile:

mpscholten:
I found two regressions in the new `TypedSqlParam` dispatch that I think should block this as-is:

1. `Id'` container coercions no longer work. The old splice coerced the whole placeholder to the expected type, so `authorIds :: [UUID]` could bind to an `author_id` param inferred as `[Id' "typed_sql_test_authors"]`. With the new exact list/Maybe instances, `typedSqlParam @(Id' "typed_sql_test_authors") authorIds` fails; same for `Maybe UUID` and `[Maybe UUID]`. Bare `UUID` still works because the broad `Coercible a col` instance handles only the scalar shape. This is a compatibility regression for existing `ANY(${ids})` / `IN (${ids})` code using raw PK values.

2. The advertised unannotated `${Nothing}` / `${[]}` forms are ambiguous. The splice now emits `typedSqlParam @col expr`, and `typedSqlParam @Text Nothing` is ambiguous between the broad `Coercible` instance and the `Maybe col` instance. `typedSqlParam @Text []` similarly cannot choose between `[col]` and `[Maybe col]`. The new enum test avoids this by writing `(Nothing :: Maybe TypedSqlTestMood)`, so the gap is not covered.

I reproduced both on PR head `e15930a1` with `direnv exec . ghci ihp-typed-sql/IHP/TypedSql/ParamEncoder.hs`. Example failing checks:

```haskell
typedSqlParam @(Id' "typed_sql_test_authors") (authorIds :: [UUID])
typedSqlParam @(Id' "typed_sql_test_authors") (Just authorId :: Maybe UUID)
typedSqlParam @Text Nothing
typedSqlParam @Text []
```

I think the core issue is that the old expression-level expected type did useful inference/coercion work, while the new class instances only model exact container shapes around `col`. Either the implementation needs to recover element-wise/container coercion and disambiguation, or the API/changelog/tests should be narrowed to require explicit target-shaped values and annotations.

amitaibu:
@mpscholten could you take it from here? Some of the stuff in the PR is going over my head :sweat_smile:

mpscholten:
Fixed the PR and resolved the merge conflicts with current `master`.

What changed:
- Kept the new typed-sql cardinality changes from `master` while preserving this PR’s parameter encoder work.
- Fixed raw primary-key values inside `Maybe`, list, and list-of-`Maybe` typed-sql placeholders by coercing the container contents to the column type.
- Disambiguated bare `${Nothing}`, `${[]}`, and `${[Nothing]}` placeholders so overlapping parameter instances select the intended shape.
- Preserved direct encoding for unhinted array-typed PostgreSQL params such as `TEXT[]`, avoiding `typedSqlParam @[Text] ([...] :: [Text])` treating an array column value as a list of scalar column values.

Verification:
- Local `ghci` load of `ihp-typed-sql/IHP/TypedSql/Quoter.hs` passed.
- Local `ghci` load of `ihp-typed-sql/Test/Test/Main.hs` passed.
- GitHub Actions are now green: `core-size` and `nix-flake-check` both passed on `4959fcbb`.

PR #2742, opened 2026-06-27
Improve typed SQL ergonomics

mpscholten:
## Summary

- add `sqlQueryTypedPipelined` plus `sqlQueryTypedMaybeColumnPipelined` for typed SQL pipeline batches
- add explicit cardinality helpers: `sqlQueryTypedRows`, `sqlQueryTypedOneOrNothing`, and `sqlQueryTypedSingle`
- add `sqlQueryTypedMaybeColumn` for nullable single-column `LIMIT 1` queries that should flatten `Maybe (Maybe a)` to `Maybe a`
- infer JSON build constructors (`json[b]_build_object`, `json[b]_build_array`) as non-null computed JSON expressions
- document the new helper choices in the guide, upgrade notes, and changelogs

## Notes

This keeps JSON nullability inference conservative. It does not globally mark `to_jsonb(...)` as non-null because `to_jsonb(NULL::...)` can be SQL NULL.

## Validation

- `git diff --check`
- `nix build .#checks.aarch64-darwin.ihp-typed-sql -L` (110 examples, 0 failures)

### DIFF
--- a/Guide/typed-sql.markdown
+++ b/Guide/typed-sql.markdown
@@ -56,6 +56,35 @@ maybeName <- sqlQueryTyped [typedSql| SELECT name FROM items ORDER BY name LIMIT
 -- maybeName :: Maybe Text
 ```
 
+If you want the expected shape in the function name, use
+`sqlQueryTypedRows`, `sqlQueryTypedOneOrNothing`, and
+`sqlQueryTypedSingle`:
+
+```haskell
+names <- sqlQueryTypedRows [typedSql| SELECT name FROM items ORDER BY name |]
+-- names :: [Text]
+
+maybeName <- sqlQueryTypedOneOrNothing [typedSql|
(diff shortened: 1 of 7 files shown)


Guide/typed-sql.markdown
```diff
@@ -56,6 +56,35 @@ maybeName <- sqlQueryTyped [typedSql| SELECT name FROM items ORDER BY name LIMIT
 -- maybeName :: Maybe Text
 ```
 
+If you want the expected shape in the function name, use
+`sqlQueryTypedRows`, `sqlQueryTypedOneOrNothing`, and
+`sqlQueryTypedSingle`:
```
amitaibu, line 61: What's the advantage of exposing those? Isn't it better that we already infer the expected shape?

mpscholten: Only useful when `sqlQueryTyped` cannot detect the correct shape

Issue #2750, opened 2026-07-09
ihp-typed-sql auto DB leaks temporary PostgreSQL clusters on macOS


mpscholten:
## Summary

`ihp-typed-sql`'s automatic compile-time database can leave behind temporary PostgreSQL clusters under macOS `$TMPDIR`. On my machine this accumulated tens of GiB of temp data and several live orphaned `postgres` processes.

## Observed impact

On macOS, before cleanup:

- `407` directories matching `ihp-typed-sql--*`
- total size: `25.06 GiB`
- average full temp cluster: about `60-73 MiB`
- `127` dirs were created in the last 24h, totaling `8.94 GiB`
- `16` live `postgres` processes were still running from these temp dirs, parented to `launchd`/`PPID 1`

Example process shape:

```console
/nix/store/...-postgresql-17.10/bin/postgres -D /var/folders/.../T/ihp-typed-sql--.../pgdata
```

Example directory contents:

```console
ihp-typed-sql--.../
  pgdata/
  postgresql.log
  socket/
```

A typical leftover cluster had two `16M` WAL files under `pgdata/pg_wal`, which explains the size per directory. Many logs showed PostgreSQL had shut down cleanly, but the temp root was still left behind; others were still live. I did not see the matching cleanup monitor process running for the orphaned live Postgres daemons.

## Relevant code

This appears to come from `ihp-typed-sql/IHP/TypedSql/CompileTimeDatabase.hs`:

```haskell
root <- createTempDirectory tempDirectory "ihp-typed-sql-"
runChecked "initdb" ["-D", pgData, "--no-locale", "--encoding=UTF8"]
runChecked "pg_ctl" ["-D", pgData, "-l", logFile, "-w", "start"]
```

Cleanup is manual via `cleanupAutoDatabase` and `startCleanupMonitor`, rather than being tied to a scoped temp-directory lifetime. `autoDatabaseState` is also process-local, so repeated GHC/GHCI/HLS/build processes create new clusters.

## Suspected failure mode

The cleanup path is fragile when the compiler/GHCI/iserv process exits unexpectedly, is interrupted, gets killed, or otherwise does not reach `stopAutoDatabase`. The shell cleanup monitor also does not appear to reliably survive/work on macOS in this setup, leaving both:

- stopped-but-not-removed `pgdata` directories
- live orphaned `postgres` processes under old temp roots

## Suggested fixes

Some options that would make this safer:

[...]

PR #2773, opened 2026-08-08
Stop typedSql's auto-database leaking postgres clusters

mpscholten:
## Symptom

On one machine, `IHP_TYPED_SQL_AUTO_DB` had left **266 directories and 7.1 GB** in `$TMPDIR` over roughly a week, 24 of them with a live postmaster.

The consequence is worse than disk. Enough live clusters exhaust SysV shared memory, and from then on *every* `typedSql` compile dies inside `initdb` with:

```
FATAL:  could not create shared memory segment: No space left on device
DETAIL:  Failed system call was shmget(...)
```

which points at the kernel rather than at the accumulated clusters. It took me a while to work out why compiles had started failing.

## Cause

`startCleanupMonitor` is called **last** in `startAutoDatabase`, after the cluster is up and the schema is loaded. That covers:

- a clean exit → monitor handles it (I verified the monitor logic works, including the `$0`/`$1`/`$2` handling, and that it survives SIGHUP)
- a *failed* start → `onException` handles it (verified: a failing `initdb` leaves nothing behind)

What it does **not** cover is a process that dies without unwinding — `Ctrl-C`, a parallel build being cancelled, the OOM killer, a reboot. Those leave the directory forever.

The leftovers are identifiable: `initdb`'s early skeleton (`base`, `global`, `pg_wal`) with **no `PG_VERSION`** — killed mid-`initdb`, between `createTempDirectory` and the monitor starting.

## Fix

**1. Start the monitor before `initdb`**, so a kill at any point after the temp directory exists is still cleaned up. The monitor already tolerates a cluster that was never started (`pg_ctl stop … || true`).

**2. Reap on startup.** The monitor can't help if it's killed too, so each run removes leftover directories whose owning compiler process is gone. The monitor now records that pid in `owner.pid`. Directories without one predate the monitor and are only reaped after a one-hour grace period, so a database being created concurrently is never pulled out from under a parallel build.

### One subtlety worth calling out

Liveness is probed with `ps -p`, not `kill -0`. `kill -0` reports **EPERM** for a process owned by another user, which is indistinguishable from "no such process" by exit code — my first version used it and **deleted a live database** in testing. The regression test pins this by using pid 1, which is both always alive and owned by root.

## Tests

Two new cases in `TypedSqlSpec`, both pure (no postgres needed):

```
TypedSql abandoned auto-database reaping (no postgres)
  reaps a directory whose owning process is gone, keeps a live one [✔]
  leaves a recent directory without an owner file alone [✔]

2 examples, 0 failures
```

## Not addressed here

While investigating I also found that **the interactive dev shell enables `IHP_TYPED_SQL_AUTO_DB=1` (`flake-module.nix:697`) without putting a matching postgres on `PATH`** — only the `nix flake check` derivation does that (`:438-468`). On a nix-darwin machine the quoter then picks up `/run/current-system/sw/bin/initdb`, and `initdb` fails with:

```
FATAL:  could not access file "$libdir/dict_snowball": No such file or directory
```

So compile-time SQL checking doesn't work in the dev shell at all there. `ensureRequiredExecutables` only checks that `initdb` is *on PATH*, never that it matches the postgres it will run against. That felt like a separate change — happy to follow up if you want it in the same PR.


mpscholten:
Closing as a duplicate — I should have searched before writing this. #2750 already describes the problem and #2751 already implements the fix, more thoroughly: same `owner.pid` mechanism (same filename, even), plus `pruneStaleProcessDirectories`, a watchdog that waits for the postmaster to actually exit, `clearStalePostmasterPid`, and resolved absolute tool paths via `PostgreSqlTools`.

Two things from my investigation that might still be useful on #2750 — I've added them there.

PR #2782, opened 2026-08-11
Decode complete table selections as model tuples

### LINKED ISSUE #2781: typed-sql: SELECT a.*, b.* over a LEFT JOIN decodes to one flat SqlRow, not (RecordA, Maybe RecordB)
Tested on v1.6 (`71b7bb685d2a`), `ihp-typed-sql` enabled in the dev shell.

## Setup

```sql
CREATE TABLE authors (
    name TEXT NOT NULL
);
CREATE TABLE items (
    title TEXT NOT NULL,
    author_id UUID DEFAULT NULL
);
```

## What works

Explicit columns over a LEFT JOIN infer nullability exactly as the guide describes:

```haskell
rows <- sqlQueryTyped [typedSql|
    SELECT i.title, a.name FROM items i
    LEFT JOIN authors a ON a.id = i.author_id
|]
-- rows :: [(Text, Maybe Text)]  ✓
```

And a single star in a join context returns the generated record, also as documented:

```haskell
items <- sqlQueryTyped [typedSqlStar|
    SELECT i.* FROM items i LEFT JOIN authors a ON a.id = i.author_id
|]
-- items :: [Item]  ✓
```

## What doesn't

Selecting both tables' stars compiles fine — but decodes to one **flat anonymous `SqlRow`**, with the right side's columns correctly `Maybe`-wrapped and colliding names deduplicated (`id` → `id_1`):

```haskell
rows <- sqlQueryTyped [typedSqlStar|
    SELECT i.*, a.* FROM items i
    LEFT JOIN authors a ON a.id = i.author_id
|]
-- rows :: [SqlRow '[ '("id", Id' "items"), '("title", Text), '("author_id", Maybe UUID)
--                  , '("id_1", Maybe (Id' "authors")), '("name", Maybe Text) ]]
```

There appears to be no way to get `[(Item, Maybe Author)]` — the shape a view built on the generated record types actually wants. In my app I ended up keeping two query-builder queries and pairing in Haskell, because consuming the flat row would mean giving up the generated records the rest of the code is typed against.

## Ask

Would per-table star decoding into record tuples — `t1.*, t2.*` → `(Record1, Maybe Record2)`, in the spirit of postgresql-simple's `:.` — be a welcome addition to `typedSqlStar`? The nullability tracking is already right; it's only the grouping into records that's missing. Happy to test a patch against a real schema.

mpscholten:
Closes #2781.

## What changed

- Decode contiguous, complete, schema-ordered qualified named-column groups into generated model records.
- Represent tables on nullable outer-join sides through the existing `FromRowHasql (Maybe Model)` interface.
- Support INNER, LEFT, RIGHT, FULL, and self-join groupings.
- Keep partial, reordered, unqualified, and expression-mixed selections on the existing `SqlRow` path.
- Preserve the existing `typedSqlStar` escape hatch without recommending stars for model-tuple queries.
- Document explicit named columns as the recommended form and add an unreleased changelog entry.

## Why

PostgreSQL metadata already preserved correct per-column nullability, but typed SQL flattened complete multi-table selections into one anonymous `SqlRow`. This made joined queries incompatible with the generated model types used by the rest of an IHP application.

Explicit named columns are supported so callers do not need to opt into schema-drift-prone star selections. Normal `typedSql` continues to reject `*`.

## Validation

- Schema compiler tests: 47 examples, 0 failures.
- New explicit-column, outer-join, mixed-selection, and self-join grouping tests pass.
- Modified typed-SQL and schema-compiler modules load successfully in GHCi.
- `git diff --check` passes.

The complete PostgreSQL-backed typed-SQL suite could not run locally because the machine had exhausted its System V shared-memory segment limit; unrelated automatic database tests failed during `initdb`.

### DIFF
--- a/Guide/typed-sql.markdown
+++ b/Guide/typed-sql.markdown
@@ -143,6 +143,23 @@ items <- sqlQueryTyped [typedSqlStar|
 |]
 ```
 
+When the select list consists entirely of qualified table stars, each star is
+decoded into that table's generated record type. Tables on the nullable side
+of an outer join are wrapped in `Maybe`:
+
+```haskell
+rows <- sqlQueryTyped [typedSqlStar|
+    SELECT i.*, a.*
+    FROM items i
+    LEFT JOIN authors a ON a.id = i.author_id
+|]
+
+-- rows :: [(Item, Maybe Author)]
+```
+
+This grouping applies only to complete qualified stars such as `i.*`. Mixed
+select lists such as `i.*, a.name` continue to produce a labeled `SqlRow`.
+
 ## Inserting Rows
 
 ### Why `INSERT … VALUES` without a column list is disallowed by default
--- a/Guide/typed-sql.markdown
+++ b/Guide/typed-sql.markdown
@@ -160,22 +160,6 @@ items <- sqlQueryTyped [typedSqlStar|
 |]
 ```
 
-The same model grouping also applies when the select list consists entirely of
-qualified table stars:
-
-```haskell
-rows <- sqlQueryTyped [typedSqlStar|
-    SELECT i.*, a.*
-    FROM items i
-    LEFT JOIN authors a ON a.id = i.author_id
-|]
-
--- rows :: [(Item, Maybe Author)]
-```
-
-Mixed select lists such as `i.*, a.name` continue to produce a labeled
-`SqlRow`.
-
 ## Inserting Rows
 
 ### Why `INSERT … VALUES` without a column list is disallowed by default
--- a/Guide/typed-sql.markdown
+++ b/Guide/typed-sql.markdown
@@ -119,6 +119,23 @@ items <- sqlQueryTyped [typedSql|
 
 The compile error message will suggest the exact column names to use.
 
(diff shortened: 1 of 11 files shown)


Guide/typed-sql.markdown
```diff
@@ -143,6 +143,23 @@ items <- sqlQueryTyped [typedSqlStar|
 |]
 ```
 
+When the select list consists entirely of qualified table stars, each star is
+decoded into that table's generated record type. Tables on the nullable side
+of an outer join are wrapped in `Maybe`:
+
+```haskell
+rows <- sqlQueryTyped [typedSqlStar|
+    SELECT i.*, a.*
```
amitaibu: So if it wasn't only qualified table stars, but named ones, would it work as expected?
I think that actually table stars are discouraged, right?

mpscholten: Yes named ones work 👍 

Correct, stars are discouraged and for normal `SELECT * FROM` queries you already would get an error. I just expanded this error to these joined select star queries as well.

PR #2797, opened 2026-08-16
Fix unreadable code blocks in the Guide's dark mode

amitaibu:
Fenced code blocks written without a language are rendered as a bare `<pre><code>`, without the `.source-code` wrapper that carries the code block styling. They therefore fall through to Bootstrap's `pre{color:#212529}` — near-black text on the dark mode background, which is effectively invisible.

Seen for example on [Deployment → Configuring the Instance](https://ihp.digitallyinduced.com/Guide/deployment.html#configuring-the-instance), where the file-tree block right below the paragraph is unreadable in dark mode, while the `.nix` block just under it (which has a language, so it gets the `.source-code` wrapper) renders fine.

This styles unwrapped `<pre>` blocks the same way as the wrapped ones — dark box, border, light monospace text — so they are readable and visually consistent in both themes. The `:not(.source-code pre)` guard keeps the existing wrapped blocks untouched (they also occur nested in list items, so a `.doc > pre` child selector would have missed some).

Verified by rendering the built `deployment.html` against the patched `layout.html` in Chrome, in both dark and light mode.

### DIFF
--- a/Guide/layout.html
+++ b/Guide/layout.html
@@ -237,6 +237,24 @@
       }
       .source-code pre::-webkit-scrollbar{height:10px}
       .source-code pre::-webkit-scrollbar-thumb{background:var(--code-border);border-radius:8px}
+
+      /* Fenced blocks without a language are emitted as a bare <pre><code>, without
+         the .source-code wrapper. Without this they fall back to Bootstrap's
+         `pre{color:#212529}`, which is unreadable on the dark mode background. */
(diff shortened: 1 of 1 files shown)


mpscholten:
Thanks 👍

amitaibu:
For posterity, here's how it looks before on https://ihp.digitallyinduced.com/Guide/deployment.html#optional-connecting-cloudwatch

image

PR #2800, opened 2026-08-23
Handle PostgreSQL line comments inside statements

vcombey:
## Bug

A PostgreSQL line comment inside a statement is currently consumed as ordinary whitespace:

```sql
CREATE TABLE users (
    id UUID, -- public identifier
    email TEXT
);
```

This can make the parser join tokens across the comment or stop at the wrong place.

## Fix

Treat `-- ...` as a real SQL comment while preserving standalone comments as schema statements. The PostgreSQL dump guard `\restrict` remains a distinct statement.

## Independence

This branch is based directly on the current `upstream/master`; it has no dependency on another parser PR.

## Validation

- `nix build --impure --no-link .#checks.aarch64-darwin.ghc914-ihp-postgres-parser`

Not PostgreSQL 18-specific.

### DIFF
--- a/ihp-postgres-parser/IHP/Postgres/Parser.hs
+++ b/ihp-postgres-parser/IHP/Postgres/Parser.hs
@@ -30,7 +30,8 @@ import qualified Data.List as List
 import Control.Monad (when)
 import Text.Megaparsec
 import Data.Void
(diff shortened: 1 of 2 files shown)


mpscholten (approved):
Thanks 👍

PR #2809, opened 2026-08-23
Keep opaque SQL out of automatic schema diffs

vcombey:
## Bug

Statements that IHP intentionally keeps opaque, such as privileges or anonymous blocks, should stay in `Schema.sql`, but they must not be treated as structured objects to add or remove automatically during schema diffing.

Otherwise a schema diff can propose noisy or unsafe migrations for SQL the migration generator does not understand.

## Fix

Classify opaque statements as migration noise in both schema inputs and cover the behavior with migration-generator tests.

## Independence

This branch is based directly on the current `upstream/master` and has no dependency on another parser PR. It does not broaden the parser or hide syntax errors.

## Validation

- `nix build --impure --no-link .#checks.aarch64-darwin.ghc914-ihp-ide`

Not PostgreSQL 18-specific.


mpscholten (approved):
Thanks 👍

PR #2823, opened 2026-09-14
Add paramOrDefaultIgnoreInvalid, and use it for pagination params

amitaibu:
`paginate` answers a malformed `page` or `maxItems` with a 500. Both are read from the URL, so the value is whatever a person typed, an old link carried, or a crawler guessed:

```
GET /users?page=abc      500
GET /users?maxItems=abc  500
GET /users?page=%205     500   -- a stray space
```

## The change

Adds `paramOrDefaultIgnoreInvalid`, a variant of `paramOrDefault` that also falls back to the default value when a parameter is present but cannot be parsed:

```haskell
paramOrDefaultIgnoreInvalid :: ParamReader a => RequestBody -> Wai.Request -> a -> ByteString -> a
paramOrDefaultIgnoreInvalid requestBody request !defaultValue name =
    case paramOrError requestBody request name of
        Left _ -> defaultValue
        Right value -> value
```

`paramOrDefault` and `paramOrNothing` are **unchanged** — they still throw `ParamCouldNotBeParsedException` on an invalid value, so a bad value in a form field is not silently ignored. The change to `wai-request-params` is purely additive.

Pagination reads `page` and `maxItems` through the new function. A numeric but out-of-range value is unchanged: still clamped to at least 1, still capped at 200 items.

## The view had the same bug

`Pagination.ViewFunctions` builds its page links from `paramOrNothing @Int "maxItems"`, so a malformed `?maxItems=abc` threw *while rendering* — fixing only the paginator would have moved the 500 rather than removed it. It now reads through `paramOrError` and drops the parameter from the generated links.

## Tests

```
Wai.Request.Params    100 examples, 0 failures
IHP.Controller.Param  105 examples, 0 failures
```

Four examples added per package for the new function (valid, empty, missing, invalid). The existing examples asserting that `paramOrDefault` and `paramOrNothing` throw on invalid input are untouched and still pass.

Pagination math was verified separately against a mocked request — valid page, non-numeric page, stray space, non-numeric `maxItems`, plus the unchanged clamp-to-1 and cap-at-200 behavior. The DB-backed pagination tests pend without a local PostgreSQL, as they do in CI.

https://claude.ai/code/session_014b677omTbfZU4qNku1LEuZ


ihp/IHP/Pagination/Internal.hs
```diff
@@ -18,19 +18,33 @@ module IHP.Pagination.Internal
...
+page = max 1 $ intParamOrDefault 1 "page"
+
+-- | The @Int@ a pagination parameter carries, or the default when it carries
+-- nothing or carries something that is not a number.
+--
+-- 'paramOrDefault' falls back only when the parameter is missing and throws a
+-- 'ParamCouldNotBeParsedException' when it is there but malformed, which for
+-- @page@ and @maxItems@ means a 500. These two are navigational: they arrive
+-- from a URL somebody typed, an old link, or a crawler, so @?page=abc@ is
+-- answered with the first page rather than with an error page.
```
amitaibu:
```suggestion
```

ihp/IHP/Pagination/Internal.hs
```diff
@@ -18,19 +18,27 @@ module IHP.Pagination.Internal
...
-pageSize' options = min (max 1 $ paramOrDefault @Int (maxItems options) "maxItems") 200
+pageSize' options = min (max 1 $ intParamOrDefault (maxItems options) "maxItems") 200
 
 -- Page and page size shouldn't be lower than 1.
 page :: (?request :: Request) => Int
-page = max 1 $ paramOrDefault @Int 1 "page"
+page = max 1 $ intParamOrDefault 1 "page"
+
+-- | The @Int@ a pagination parameter carries, or the default when it carries
+-- nothing or carries something that is not a number.
```
mpscholten: Let's also describe the motivation / the example you had where it was failing and why this is being added here. Otherwise in the future another agent might think it can replace this with just `paramOrDefault` again :D

amitaibu: Good call — and following the thread further, the answer turned out to be that it *should* just be `paramOrDefault`. So rather than document the workaround, I fixed it at the source and removed the helper.

`paramOrDefault` and `paramOrNothing` now fall back on a value that is present but unparseable, the same as they already did for a missing one. `param` and `paramOrError` are unchanged and remain the way to reject bad input, and `fill` is unaffected since it already used `paramOrError` and turned parse failures into validation errors.

Two things worth your attention:

- The Guide already documented `paramOrDefault` as falling back "(or when invalid)", so the code was contradicting its own docs.
- This also fixes the same bug in `Pagination.ViewFunctions` (`paramOrNothing @Int "maxItems"`) and in the DataSync REST API, where `?limit=abc` used to 500.

It is a breaking change though: the two existing tests that asserted the throwing behavior are now inverted. That reverses a deliberate, tested decision, so it is worth a proper look rather than a rubber stamp — happy to split the `paramOrNothing` half out if you would rather land them separately.

mpscholten: This is a breaking change, before `paramOrDefault` was specifically designed such that invalid errors are not silently ignored. I would prefer to keep it that way, as it might break some apps in silent unexpected ways.

But as an alternative we could add something like `paramOrDefaultIgnoreInvalid` (or maybe you can find a better name)

amitaibu: > before paramOrDefault was specifically designed such that invalid errors are not silently ignored

Sure, I can change it, but let me push just a little more - Is there a good use case for that, for getting a 500 error instead of falling back to the default value?

mpscholten: Yes, e.g. for apis. For an api you want to error instead of silently accepting invalid input and then falling back to a default silently

Issue #2628, opened 2026-04-12
autoRefresh does not work


avitkauskas:
On latest master 7a76188 (2026-04-11) - though I am sure it stopped working some IHP commits before that.
I noticed that autoRefresh is not working on my app.

I see an error in the browser console:
<img width="514" height="116" alt="Image" src="https://github.com/user-attachments/assets/ecbeee85-8d05-45fc-90b6-03121f265de2" />

In network tab:
<img width="914" height="47" alt="Image" src="https://github.com/user-attachments/assets/82b48869-5bf7-4899-badd-9055ef938f5d" />

I do not see any details about the message. No errors in IHP devenv console.

Is it broken or were there some changes that I missed and should address in my app?

avitkauskas:
Looks like it was not fixed :(
Though no errors now in browser console - but auto refresh does not happen.

PR #2824, opened 2026-09-15
Socket-dependent tests skip when the sandbox denies them

amitaibu:
Two tests open real TCP sockets, which the Nix build sandbox on macOS forbids. Both fail the `aarch64-darwin` check with `IOException of type PermissionDenied`:

- `ControllerSupportSpec` — the #2628 regression test calls `Warp.testWithApplication`: `Network.Socket.bind: permission denied`
- `PortConfigSpec` — `portConfigFromEnvironment` scans for a free port by connecting to candidates: `Network.Socket.connect: permission denied`

This turns CI red on any PR that touches `ihp/` sources (seen on #2823, whose diff is only pagination files).

## Why master looks green

Runs on master resolve the derivation from the cache and never execute the suites on the runner:

```
✅ checks.aarch64-darwin.default (previously built)
```

A PR changing `ihp/` sources changes the derivation hash, forcing a real rebuild — which runs the tests for the first time and hits the restriction. `tests.yml` runs only on `ARM64`, so this is the sole gating check.

## The change

Catch that specific failure in each test and mark the example pending, matching the `pendingWith` pattern already used for an unavailable PostgreSQL in `FetchPipelinedSpec` and `JobQueueSpec`:

```haskell
case result of
    Right () -> pure ()
    Left exception
        | isPermissionError exception ->
            pendingWith "Binding a local socket is not permitted in this sandbox"
        | otherwise -> Exception.throwIO exception
```

Any other exception still propagates, so a genuine regression of #2628 keeps failing the build.

## Verification

The sandbox skip is confirmed by CI itself — the darwin runner now reports:

```
# PENDING: Binding a local socket is not permitted in this sandbox
```

Locally on Linux, where sockets are permitted, both tests still really run rather than silently skipping:

```
  startWebSocketApp
    allows a real WebSocket connection through Warp [v]
8 examples, 0 failures

IHP.IDE.PortConfig.portConfigFromEnvironment
  falls back to scanning from the default port when PORT is unset [v]
2 examples, 0 failures
```

https://claude.ai/code/session_014b677omTbfZU4qNku1LEuZ


mpscholten (approved):
Thanks 👍

PR #2825, opened 2026-09-19
Make ihp-typed-sql a standalone package (like ihp-hsx, ihp-router)

dukefong8:
Purpose
===

ihp-typed-sql no longer depends on ihp and can be installed standalone. The IHP-specific runners (IHP.TypedSql, IHP.TypedSql.Pagination) moved into the ihp package, which now depends on ihp-typed-sql. Id' and PrimaryKey have a single owner (IHP.TypedSql.Id); DefaultParamEncoder instances are owned by exactly one package (PostGIS geometry stays in ihp); hasql 1.x bounds and plain-cabal support (standalone cabal.project, nix-free spec harness) are kept.

Test
===
 
Test scoreboard (nix dev shell, TCP postgres at 127.0.0.1:5432): - ihp-typed-sql:spec (standalone): 119 examples, 0 failures, 9 pending (AUTO_DB private-cluster tests stay xit; suite runs on TCP) - ihp-datasync:test:ihp-datasync-hspec: 74 examples, 0 failures - ihp-ide:test:tests: 489 examples, 0 failures

Co-authored with AI (Muse Spark via OpenCode).


ihp-typed-sql/IHP/TypedSql/Prelude.hs
```diff
@@ -0,0 +1,21 @@
...
+    ( module Prelude
+    , (|>)
+    ) where
+
+import           Prelude
+
+-- | Pipe operator: @a |> f@ is @f a@.
+infixl 8 |>
+(|>) :: a -> (a -> b) -> b
+a |> f = f a
```
mpscholten: Let's just drop `|>` and use normal function composition in the library. Then we avoid this extra module

dukefong8: Hi, please review change pushed below.

mpscholten:
Review of 29622f8af29ce0cbf418ccf34ccb2757fba125d9: I would request changes on three points.

1. **P2 — GHC compatibility regression** ([Metadata.hs:190](https://github.com/digitallyinduced/ihp/blob/29622f8af29ce0cbf418ccf34ccb2757fba125d9/ihp-typed-sql/IHP/TypedSql/Metadata.hs#L190)). Replacing `IHP.Prelude` leaves the unqualified `foldl'` calls supplied only by `Prelude`, which exports that function starting with base 4.20. The package permits older versions, so GHC 9.8 builds fail. Use the qualified `Data.List` function explicitly.

2. **P2 — IHP runtime tests deleted, not migrated** ([TypedSqlSpec.hs:818–825](https://github.com/digitallyinduced/ihp/blob/29622f8af29ce0cbf418ccf34ccb2757fba125d9/ihp-typed-sql/Test/Test/TypedSqlSpec.hs#L818-L825)). The comment claims request-scoped runner and pagination tests now live in `ihp`, but the PR adds no replacement suite there. The existing integration-test module contains only two simple cases, not equivalent coverage. Please move the removed tests alongside the moved implementation.

3. **P2 — All nine AUTO_DB tests unconditionally disabled** ([TypedSqlSpec.hs:233–235](https://github.com/digitallyinduced/ihp/blob/29622f8af29ce0cbf418ccf34ccb2757fba125d9/ihp-typed-sql/Test/Test/TypedSqlSpec.hs#L233-L235)). `xit` prevents execution even when PostgreSQL tools are available. These tests already use isolated fixtures, explicit database environment settings, and capability checks. Please retain conditional execution rather than permanently disabling coverage.

Validation: source review and partial GHCi checking. Full checking was blocked by the available environment having an older `postgresql-syntax` dependency.

